PR #1, opened 2021-07-02
Two initial checks added

miha-stopar:
Two memory checks added - whether the initial state is ok and whether the value that is read is the same as the value that was written at the same address.

### DIFF
--- a/src/state_circuit/memory.rs
+++ b/src/state_circuit/memory.rs
@@ -11,7 +11,7 @@ use std::marker::PhantomData;
 /*
 Example memory table:
 
-| address | step | value | flag |
+| address | global_counter | value | flag |
 ---------------------------------
 |    0    |   0  |   0   |   1  |
 |    0    |  12  |  12   |   1  |
@@ -24,7 +24,7 @@ Example memory table:
 /*
 Example bus mapping:
 
-| step | memory_flag | memory_address | memory_value |
+| global_counter | memory_flag | memory_address | memory_value |
 ------------------------------------------------------
 |  12  |      1      |        0       |      12      |
 |  17  |      1      |        1       |      32      |
@@ -32,7 +32,7 @@ Example bus mapping:
 |  89  |      0      |        1       |      32      |
 */
 
-/// In the state proof, memory operations are ordered first by address, and then by step.
+/// In the state proof, memory operations are ordered first by address, and then by global_counter.
 /// Memory is initialised at 0 for each new address.
 /// Memory is a word-addressed byte array, i.e. a mapping from a 253-bit word -> u8.
 #[derive(Copy, Clone, Debug)]
@@ -40,7 +40,7 @@ struct MemoryAddress<F: FieldExt>(F);
 
 /// Global counter
 #[derive(Copy, Clone, Debug)]
-struct Step(usize);
+struct GlobalCounter(usize);
 
 /// TODO: In the EVM we can only read memory values in 32 bytes, but can write either
 /// single-byte or 32-byte chunks. In zkEVM:
@@ -52,15 +52,15 @@ struct Value<F: FieldExt>(F);
 #[derive(Clone, Debug)]
 enum ReadWrite<F: FieldExt> {
     // flag == 0
-    Read(Step, Value<F>),
+    Read(GlobalCounter, Value<F>),
     // flag == 1
-    Write(Step, Value<F>),
+    Write(GlobalCounter, Value<F>),
 }
 
 impl<F: FieldExt> ReadWrite<F> {
-    fn step(&self) -> Step {
+    fn global_counter(&self) -> GlobalCounter {
         match self {
-            Self::Read(step, _) | Self::Write(step, _) => *step,
+            Self::Read(global_counter, _) | Self::Write(global_counter, _) => *global_counter,
         }
     }
 
@@ -82,13 +82,13 @@ impl<F: FieldExt> ReadWrite<F> {
 /// All the read/write operations that happen at this address.
 pub(crate) struct MemoryOp<F: FieldExt> {
     address: MemoryAddress<F>,
-    steps: Vec<Option<ReadWrite<F>>>,
+    global_counters: Vec<Option<ReadWrite<F>>>,
 }
 
 /*
 Example bus mapping:
 
-| step | memory_flag | memory_address | memory_value |
+| global_counter | memory_flag | memory_address | memory_value |
 ------------------------------------------------------
 |  12  |      1      |       0        |      12      |
 |  17  |      1      |       1        |      32      |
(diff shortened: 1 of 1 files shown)


src/state_circuit/memory.rs
```diff
@@ -129,33 +133,53 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
         // Constraints for each step. These are activated when q_memory is nonzero.
         meta.create_gate("Memory operation", |meta| {
             let q_memory = meta.query_selector(q_memory);
+            let q_memory_init = meta.query_selector(q_memory_init);
```
therealyingtong, line 136: We want to define a synthetic `q_memory_init` selector within the gate, instead of "hard-coding" the intervals at which we enable the init selector outside the gate. This is because each memory location sees an arbitrary number of operations per proof, and we cannot know ahead of time the offsets at which we check for an init.
```suggestion
            // If address_cur != address_prev, this is an `init`.
            let q_memory_init = {
                let address_prev = meta.query_advice(address, Rotation::prev());
                let address_cur = meta.query_advice(address, Rotation::cur());    
                address_cur - address_prev
            };
```

src/state_circuit/memory.rs
```diff
@@ -129,33 +133,53 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
...
+            let q_same_value = meta.query_selector(q_same_value);
```
therealyingtong, line 137: Similarly, define a synthetic selector for `q_same_value`. (Separately, I recommend renaming `q_same_value` -> `q_read`). The number of reads varies from proof to proof, and we cannot hard-code the offsets at which we check for a read.

(We will have to move the `one` expression into scope here.)

```suggestion
            let q_same_value = one - flag.clone();
```

src/state_circuit/memory.rs
```diff
@@ -200,46 +219,39 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
...
-        // Assign memory_flag
-        region.assign_advice(|| "init memory", self.flag, offset, || Ok(F::one()))?;
-
-        Ok(())
-    }
-
     /// Assign cells for each step in an operation.
     fn step(
         &self,
+        step_offset: usize,
```
therealyingtong, line 244: If we only use the `q_memory` selector, and downgrade `q_memory_init` and `q_same_value` to synthetic selectors inside the gate, we no longer have to switch on `step_offset` here.

src/state_circuit/memory.rs
```diff
@@ -200,46 +219,39 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
...
-
-        // Assign `value`
-        region.assign_advice(|| "init value", self.value, offset, || Ok(F::zero()))?;
-
-        // Assign memory_flag
-        region.assign_advice(|| "init memory", self.flag, offset, || Ok(F::one()))?;
-
-        Ok(())
-    }
-
```
therealyingtong, line 233: Restore this: `q_memory_enable` can handle the `init` case.

src/state_circuit/memory.rs
```diff
@@ -357,6 +373,7 @@ mod tests {
         let op_0 = MemoryOp {
             address: MemoryAddress(pallas::Base::zero()),
             steps: vec![
+                Some(ReadWrite::Write(Step(0), Value(pallas::Base::from_u64(0)))),
```
therealyingtong, line 376: We don't have to explicitly provide the init case in the witness; the `init()` function takes care of it for us.

src/state_circuit/memory.rs
```diff
@@ -368,6 +385,7 @@ mod tests {
         let op_1 = MemoryOp {
             address: MemoryAddress(pallas::Base::one()),
             steps: vec![
+                Some(ReadWrite::Write(Step(0), Value(pallas::Base::from_u64(0)))),
```
therealyingtong, line 388: We don't have to explicitly provide the init case in the witness; the `init()` function takes care of it for us.

therealyingtong (changes requested):

src/state_circuit/memory.rs
```diff
@@ -189,9 +239,8 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
                     // Increase offset by 1 after initialising.
                     offset += 1;
 
-                    for (step_offset, step) in op.steps.iter().enumerate() {
-                        let bus_mapping =
-                            self.step(&mut region, offset + step_offset, address, step)?;
+                    for step in op.steps.iter() {
+                        let bus_mapping = self.step(&mut region, offset, address, step)?;
```
therealyingtong, line 243: We didn't need the extra `+ step_offset`. Thanks for the catch!

src/state_circuit/memory.rs
```diff
@@ -135,27 +156,53 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
...
 
-            // TODO: If flag == 0 (read), and step != 0, value_prev == value_cur
+            let value_prev = meta.query_advice(value, Rotation::prev());
+            // If flag == 0 (read), and step != 0, value_prev == value_cur
+            let q_read = {
+                let one = Expression::Constant(F::one());
+                one - flag.clone()
+            };
 
             // step_prev < step_cur
```
therealyingtong, line 188:
```suggestion
            // If address_prev == address_cur, step_prev < step_cur
```

src/state_circuit/memory.rs
```diff
@@ -135,27 +156,53 @@ impl<F: FieldExt, const NUM_STEPS: usize> Config<F, NUM_STEPS> {
...
             // TODO: Figure out how to enforce this. Suggestion: lookup step_cur, step_prev,
             // and their difference `step_cur - step_prev` in a table.
 
             // TODO: address[i] == address[i + 1] == ... == address[i + NUM_STEPS - 1]
```
therealyingtong, line 192: Monotonicity check.

```suggestion
            // TODO: address_prev <= address_cur
```

src/state_circuit/memory.rs
miha-stopar: Cool, thanks, I will change it. BTW, shall I rename step into global_counter everywhere?

therealyingtong: yes, i think renaming `step` -> `global_counter` would be good.

miha-stopar: Cool, just renamed it. We don't need NUM_STEPS, right?

miha-stopar:
Agreed with @therealyingtong to merge this PR.

Issue #7, opened 2021-07-06
Monotonically increasing gadget


therealyingtong:
Check that a list of witnessed values is monotonically increasing.

han0110:
Close due to #11

PR #11, opened 2021-07-08
Implement gadget monotone

han0110:
In memory circuit, we need to check if `gc` are strict-monotone in the same `address`. And `gc` has some not too large maximum value so we can be sure diff of each two adjacent row will not exceed the maximum (I call `RANGE` here). (#7)

In this gadget, we pass a advice column for check, with a function returning any expression to play as a switch of this gadget. In this way, we can use `is_equal_rot`'s expression to check if `gc` are strict-monotone when `address === address_prev`.

Also the gadget has three const-generic parameters `RANGE`, `INCR` and `STRICT` for flexibility.

- `RANGE` - specifying the maximum the diff could be
- `INCR` - should be monotone increasing or monotone decreasing (`q_enable` always specify `Rotation::cur()` and compare to `Rotation::prev()`)
- `STRICT` - should disallow equality of diff or not.

For `gc` we could use `StrictMonoIncrChip` for convenience.

### DIFF
--- a/src/gadget/monotone.rs
+++ b/src/gadget/monotone.rs
@@ -9,17 +9,15 @@ use std::{marker::PhantomData, u64};
 #[derive(Clone, Debug)]
 pub(crate) struct MonotoneConfig {
     range_table: Column<Fixed>,
+    value: Column<Advice>,
 }
 
-/// MonotoneChip helps to check if an advice column is monotone increasing
+/// MonotoneChip helps to check if an advice column is monotonically increasing
 /// within a range. With strict enabled, it disallows equality of two cell.
 pub(crate) struct MonotoneChip<F, const RANGE: usize, const INCR: bool, const STRICT: bool> {
     config: MonotoneConfig,
     _marker: PhantomData<F>,
 }
-
-pub(crate) type StrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, true>;
-pub(crate) type NonStrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, false>;
 
 impl<F: FieldExt, const RANGE: usize, const INCR: bool, const STRICT: bool>
     MonotoneChip<F, RANGE, INCR, STRICT>
@@ -33,7 +31,7 @@ impl<F: FieldExt, const RANGE: usize, const INCR: bool, const STRICT: bool>
     ) -> MonotoneConfig {
         let range_table = meta.fixed_column();
 
-        let config = MonotoneConfig { range_table };
+        let config = MonotoneConfig { range_table, value };
 
         meta.lookup(|meta| {
             let q_enable = q_enable(meta);
@@ -103,7 +101,7 @@ impl<F: FieldExt, const RANGE: usize, const INCR: bool, const STRICT: bool> Chip
 
 #[cfg(test)]
 mod test {
-    use super::{MonotoneConfig, NonStrictMonoIncrChip, StrictMonoIncrChip};
+    use super::{MonotoneChip, MonotoneConfig};
     use halo2::{
         arithmetic::FieldExt,
         circuit::{layouter::SingleChipLayouter, Layouter},
@@ -115,6 +113,9 @@ mod test {
     };
     use pasta_curves::pallas::Base;
     use std::marker::PhantomData;
+
+    type StrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, true>;
+    type NonStrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, false>;
 
     #[test]
     fn mono_incr() {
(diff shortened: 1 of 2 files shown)


src/gadget/monotone.rs
```diff
@@ -0,0 +1,259 @@
...
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Fixed, VirtualCells},
+    poly::Rotation,
+};
+use pasta_curves::arithmetic::FieldExt;
+use std::{marker::PhantomData, u64};
+
+#[derive(Clone, Debug)]
+pub(crate) struct MonotoneConfig {
+    range_table: Column<Fixed>,
+}
```
therealyingtong, line 12: Include the advice column in the Config.

```suggestion
pub(crate) struct MonotoneConfig {
    range_table: Column<Fixed>,
    value: Column<Advice>,
}
```

src/gadget/monotone.rs
```diff
@@ -0,0 +1,259 @@
...
+
+/// MonotoneChip helps to check if an advice column is monotone increasing
+/// within a range. With strict enabled, it disallows equality of two cell.
+pub(crate) struct MonotoneChip<F, const RANGE: usize, const INCR: bool, const STRICT: bool> {
+    config: MonotoneConfig,
+    _marker: PhantomData<F>,
+}
+
+pub(crate) type StrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, true>;
+pub(crate) type NonStrictMonoIncrChip<F, const RANGE: usize> = MonotoneChip<F, RANGE, true, false>;
```
therealyingtong, line 22: I'm not sure that we need to define these here; we can just leave it up to the caller to instantiate `MonotoneChip` with the const generics they need.

src/gadget/monotone.rs
```diff
@@ -0,0 +1,259 @@
...
+
+impl<F: FieldExt, const RANGE: usize, const INCR: bool, const STRICT: bool>
+    MonotoneChip<F, RANGE, INCR, STRICT>
+{
+    /// configure which column should be check. q_enable here as a fn is
+    /// flexible for synthetic selector instead of a fixed one.
+    pub fn configure(
+        meta: &mut ConstraintSystem<F>,
+        q_enable: impl FnOnce(&mut VirtualCells<'_, F>) -> Expression<F>,
```
therealyingtong, line 31: It may be simpler here to just pass in `q_enable: Expression<F>`, which is still general enough for a synthetic selector.

src/gadget/monotone.rs
```diff
@@ -0,0 +1,259 @@
...
+};
+use pasta_curves::arithmetic::FieldExt;
+use std::{marker::PhantomData, u64};
+
+#[derive(Clone, Debug)]
+pub(crate) struct MonotoneConfig {
+    range_table: Column<Fixed>,
+}
+
+/// MonotoneChip helps to check if an advice column is monotone increasing
```
therealyingtong, line 14:
```suggestion
/// MonotoneChip helps to check if an advice column is monotonically increasing
```

src/gadget/monotone.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    #[test]
+    fn mono_incr() {
+        #[derive(Clone, Debug)]
+        struct TestCircuitConfig {
+            q_enable: Selector,
+            value: Column<Advice>,
+            strict_mono_incr: MonotoneConfig,
+            non_strict_mono_incr: MonotoneConfig,
+        }
```
therealyingtong, line 127: I suggest separating the strict and non-strict tests, so we can identify which is expected to fail in the unit tests.

therealyingtong (changes requested):

src/gadget/monotone.rs
han0110: I tried to pass `Expression<F>` at first actually, but the only way to create a `Expression` and put the column into `Gate.queried_selectors` is to call `VirtualCells`'s function, and the only way we have `VirtualCells` is to borrow `mut ConstraintSystem` (call `create_gate` or `lookup`), but the first parameter of `configure` is also a borrowed `mut ConstraintSystem`, which will be blocked by borrow checker. So finally I yield to pass a function to get the expression without borrowing `mut ConstraintSystem`.

Maybe there is other better way to pass through this, do you have any suggestion I can try?

src/gadget/monotone.rs
han0110: Fixed in third commit

src/gadget/monotone.rs
han0110: Fixed in second commit

src/gadget/monotone.rs
han0110: Fixed in second commit

src/gadget/monotone.rs
therealyingtong: Let's use this pattern for now!

therealyingtong (approved):
LGTM

Issue #8, opened 2021-07-06
`is_zero` gadget


therealyingtong:
Gadget to check if a witnessed value is zero.

han0110:
Close due to #12

PR #12, opened 2021-07-08
Implement gadget is_zero

han0110:
In memory circuit, we need to check if `gc` are strict-monotone in the same memory address. At the same time, we have to escape the init row of different memory key (also for stack key). So we need a gadget to help enable `gc`'s monotone check when `is_zero(address - address_prev)`. (#8)

Another example, we might need to check if a witness `op` is equal to another fixed column and enable custom constraint in evm circuit, and we can use `is_zero(op - f_op)` to help.

In this gadget, we need a column to let prover to witness the inverse of `value_diff` or zero if no inverse, which gives us a bool expression indicating that `is_zero(value_diff)` by `1 - value_diff ⋅ value_diff.invert_or_zero()`

For flexibility, the gadget allows parent circuit to pass `value_diff` by expression, which can handle both `address` and `op` case mentioned above.

### DIFF
--- a/src/gadget/is_zero.rs
+++ b/src/gadget/is_zero.rs
@@ -4,7 +4,7 @@ use halo2::{
     poly::Rotation,
 };
 use pasta_curves::arithmetic::FieldExt;
-use std::{array, marker::PhantomData};
+use std::array;
 
 pub(crate) trait IsZeroInstruction<F: FieldExt> {
     /// Given a `value` to be checked if it is zero:
@@ -30,7 +30,6 @@ pub(crate) struct IsZeroConfig<F> {
 
 pub(crate) struct IsZeroChip<F> {
     config: IsZeroConfig<F>,
-    _marker: PhantomData<F>,
 }
 
 impl<F: FieldExt> IsZeroChip<F> {
@@ -40,7 +39,8 @@ impl<F: FieldExt> IsZeroChip<F> {
         value: impl FnOnce(&mut VirtualCells<'_, F>) -> Expression<F>,
         value_inv: Column<Advice>,
     ) -> IsZeroConfig<F> {
-        let mut is_zero_expression = None;
+        // dummy initialization
+        let mut is_zero_expression = Expression::Constant(F::zero());
 
         // Truth table of iz_zero gate:
         // +----+-------+-----------+-----------------------+---------------------------------+-------------------------------------+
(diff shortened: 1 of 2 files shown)


src/gadget/is_zero.rs
```diff
@@ -0,0 +1,440 @@
...
+    pub fn configure(
+        meta: &mut ConstraintSystem<F>,
+        q_enable: Selector,
+        value_diff: impl FnOnce(&mut VirtualCells<'_, F>) -> Expression<F>,
+        value_diff_inv: Column<Advice>,
+    ) -> IsZeroConfig<F> {
+        let mut is_zero_expression = None;
+
+        meta.create_gate(
+            "value_diff ⋅ (1 - value_diff ⋅ value_diff_inv) = 0",
```
therealyingtong, line 46: Consider adding this constraint `value_diff = 0 => valud_diff_inv = 0`
```suggestion
            "value_diff_inv ⋅ (1 - value_diff ⋅ value_diff_inv) = 0",
```

han0110: Fixed in [ac331b5b](https://github.com/appliedzkp/zkevm-circuits/pull/12/commits/ac331b5b2689df4f22635e15c5caeb9d7d0c731e).

src/gadget/is_zero.rs
```diff
@@ -0,0 +1,450 @@
...
+                    is_zero,
+                };
+
+                meta.create_gate("check is_zero", |meta| {
+                    let q_enable = meta.query_selector(q_enable);
+
+                    // This verifies is_zero is calculated correctly
+                    let check = meta.query_advice(config.check, Rotation::cur());
+
+                    vec![q_enable * (config.is_zero.is_zero_expression.clone() - check)]
```
therealyingtong, line 212:
```suggestion
                    vec![q_enable * (config.is_zero.is_zero_expression - check)]
```

src/gadget/is_zero.rs
```diff
@@ -0,0 +1,450 @@
...
```
therealyingtong, line 356:
```suggestion
                    vec![q_enable * (config.is_zero.is_zero_expression - check)]
```

therealyingtong (approved):
Approved with non-blocking suggestions.

src/gadget/is_zero.rs
han0110: The `config` here will be returned, so if in closure we still need `.clone()` (same as the above suggestion)

src/gadget/is_zero.rs
```diff
@@ -0,0 +1,450 @@
...
+            // value_inv ⋅ (1 - value ⋅ value_inv)
+            let poly2 = value_inv * is_zero_expression.clone().unwrap();
+
+            array::IntoIter::new([poly1, poly2]).map(move |poly| q_enable.clone() * poly)
+        });
+
+        IsZeroConfig::<F> {
+            q_enable,
+            value_inv,
+            is_zero_expression: is_zero_expression.unwrap(),
```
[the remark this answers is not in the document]
han0110: Fix the annoying `Option::unwrap` by dummy initialization to `Expression::Constant(F::zero())` (we need the init cause it's borrowed in closure). Unused `_marker` is removed as well in [`e9dd0912`](https://github.com/appliedzkp/zkevm-circuits/pull/12/commits/e9dd09125dbc1f619ead18eff8b033a8505f81ea)

PR #13, opened 2021-07-14
global counter and address constraints

miha-stopar:
The following constraints have been added:
- global_counter is in some predefined range
- global_counter monotonicity (within the same address)
- address is in some predefined range
- address monotonicity

### DIFF
--- a/src/gadget/is_zero.rs
+++ b/src/gadget/is_zero.rs
@@ -9,9 +9,8 @@ use std::array;
 pub(crate) trait IsZeroInstruction<F: FieldExt> {
     /// Given a `value` to be checked if it is zero:
     ///   - witnesses `inv0(value)`, where `inv0(x)` is 0 when `x` = 0,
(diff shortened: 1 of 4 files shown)


miha-stopar:
1ec3e78 commit fixed the fixed column issue - the fixed column was not fixed for all proofs for this circuit. In the same commit, is_zero and monotone gadgets have been used for some checks in the memory circuit.

Also, changed a bit the is_zero gadget, hope that's ok. The changes are:
 - IsZeroInstruction is_zero method was renamed into assign
 - the enabling of the selector has been removed from the assign method (previously is_zero) as I think this needs to be done outside the gadget - because fixed column needs to be the same for all the proofs (for the same circuit) and cannot depend on the number of values that need to be checked whether they are zero
 - diff_inv column is now hidden inside the gadget

therealyingtong:
Closed in favour of https://github.com/appliedzkp/zkevm-circuits/pull/23.

PR #21, opened 2021-07-22
8 bit word encoding

### LINKED ISSUE #17: Implement the 8-bit word encoding
### What's wrong

We plan to use the BN254 curve that supports a max integer of 2**254, which is less than a typical 256-bit word in EVM.

To work with EVM values in the circuit, we need to encode a 256-bit value into 8 bits chunks. Take `word256 = 1` example, the corresponding 8 bit words is `word8s = [1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]`.

Other details are specified here https://hackmd.io/@liangcc/zkvmbook/https%3A%2F%2Fhackmd.io%2FQUiYS3MnTu29s62yg9EtLQ

### How are we fixing this

- Implement the circuit to check if an array of 8-bit words corresponds to a commitment of a 256-bit word.
- The random number is not finalized yet. Can implement `hash(ALL8BITS)` for the time being.

therealyingtong:
Closes #17.

### DIFF
--- a/src/gadget/evm_word.rs
+++ b/src/gadget/evm_word.rs
@@ -1,0 +1,160 @@
+//! An EVM word (256 bits) is represented as a linear combination of 32 bytes:
+//!         encode(word) = b_0 + r * b_1 + ... + r^{31} * b_{31},
+//! where `word` is a 256-bit word, b_i's are bytes, and `r` is a random factor.
+//! This helper returns an expression of `encode(word)`.
+//!
+//! In the zkevm circuit, this `encode(word)` expression will not be directly
+//! looked up. Instead, it will be folded into the bus mapping lookup.
+
+use crate::gadget::Variable;
+use digest::{FixedOutput, Input};
+use halo2::{
+    circuit::Region,
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Fixed, Selector},
+    poly::Rotation,
+};
+use pasta_curves::arithmetic::FieldExt;
+use sha3::{Digest, Keccak256};
+use std::convert::TryInto;
+
+#[cfg(test)]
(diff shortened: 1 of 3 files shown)


src/gadget/evm_word.rs
```diff
@@ -0,0 +1,227 @@
...
+            None
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct WordConfig<F: FieldExt> {
+    // Randomness used to compress the word encoding.
+    r: F,
+    // Selector to toggle the word encoding gate.
```
therealyingtong, line 55:
```suggestion
    // Selector to toggle the word encoding lookups.
```

src/gadget/evm_word.rs
```diff
@@ -0,0 +1,269 @@
...
+        byte_lookup: Column<Fixed>,
+    ) -> Self {
+        // Expression representing `encode(word)`.
+        let mut encode_word_expr = Expression::Constant(F::zero());
+
+        // Lookup each byte in the witnessed word representation to
+        // range-constrain it to 8 bits.
+        //
+        // TODO: Understand why the `for` loop cannot be moved into
+        // the meta.lookup() call.
```
therealyingtong, line 88:
```suggestion
        // The `for` loop is outside the meta.lookup() call because we are reusing
        // the same table column across all bytes.
```

ChihChengLiang: Do we want to includes these doc string updates in the future PR?

ChihChengLiang (commented):
LGTM. I commented on the doc strings.

PR #23, opened 2021-07-23
State circuit

miha-stopar:
File memory.rs renamed into state.rs and made it flexible enough to cover the stack circuit too. Also, range value tables (global_counter and address) now load the maximum value too - consequently, for example, now you specify 1023 for max address in the stack circuit and not 1024.


src/gadget/is_zero.rs
```diff
@@ -83,18 +83,15 @@ impl<F: FieldExt> IsZeroChip<F> {
...
+    fn assign(
         &self,
-        mut region: &mut Region<'_, F>,
+        region: &mut Region<'_, F>,
         offset: usize,
         value: Option<F>,
     ) -> Result<(), Error> {
         let config = self.config();
 
-        config.q_enable.enable(&mut region, offset)?;
```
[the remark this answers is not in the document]
miha-stopar: OK, aded a comment.

therealyingtong:
[Force-pushed](https://github.com/appliedzkp/zkevm-circuits/compare/f004cbd696e65a3e08a3fc4bb8966ca3202565a2..a62cc37a810aa5b3085ac02cebd8d203e8c05973) to rebase on main.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+            struct StateCircuit<
+                F: FieldExt,
+                const MAX_ROWS: usize,
+                const GLOBAL_COUNTER_MAX: usize,
+                const ADDRESS_MAX: usize,
+                const ADDRESS_INCR: bool,
+                const ADDRESS_DIFF_STRICT: bool,
+            > {
+                ops: Vec<Op<F>>,
+                _marker: PhantomData<F>,
```
therealyingtong, line 558:
```suggestion
```

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct Config<
+    F: FieldExt,
+    const MAX_ROWS: usize,
+    const GLOBAL_COUNTER_MAX: usize,
+    const ADDRESS_MAX: usize,
+    const ADDRESS_INCR: bool,
+    const ADDRESS_DIFF_STRICT: bool,
```
therealyingtong, line 109: `ADDRESS_DIFF_STRICT` seems to always be `false` in the examples.
1. Is this correct -- do we never want to check strict monotonicity?
2. If it is always `false`, we could remove it from the const generics.

src/state_circuit.rs
```diff
@@ -1,3 +1,3 @@
 //! The state circuit implementation.
 
-pub(crate) mod memory;
+pub(crate) mod state;
```
therealyingtong, line 3: Do we expect another `mod storage`?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+                    chip.assign(&mut region, offset, Some(address.0 - address_prev.0))?;
+
+                    // Increase offset by 1 after initialising.
+                    offset += 1;
+
+                    for global_counter in op.global_counters.iter() {
+                        self.q_not_first.enable(&mut region, offset)?;
+                        let bus_mapping =
+                            self.assign_per_counter(&mut region, offset, address, global_counter)?;
+                        chip.assign(&mut region, offset, Some(F::zero()))?;
```
therealyingtong, line 370:
```suggestion
                        // Non-init rows in the same operation have the same address.
                        chip.assign(&mut region, offset, Some(F::zero()))?;
```

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+                    }
+                }
+
+                let last_address = ops.last().unwrap().address;
+                let last_gc = ops.last().unwrap().global_counters.last().unwrap();
+                let mut last_counter = last_gc.as_ref().unwrap().global_counter().0;
+
+                // We pad all remaining memory rows to avoid the check at the first unused row.
+                // Without padding, (address_cur - address_prev) would not be zero at the first unused row
+                // and some checks would be triggered.
```
therealyingtong, line 385: TODO: We eventually want to exclude the padding rows from the bus mapping lookup.

src/state_circuit/state.rs
miha-stopar: Strict monotonicity is needed for global counter when the address stays the same, but there it's a bit tricky to use the monotone gadget and we don't use it currently (https://github.com/appliedzkp/zkevm-circuits/pull/23#discussion_r676472311). Agree, we can remove it, but I am just changing this in #24. Can I do it there?

src/state_circuit.rs
miha-stopar: Not sure right now, I plan to experiment with this in #24. I might start there with all checks for stack and memory in one file and then modularize it (storage would probably come in a later PR). Would be that ok?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+    circuit::{Layouter, Region},
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Fixed, Selector},
+    poly::Rotation,
+};
+use pasta_curves::arithmetic::FieldExt;
+
+use std::marker::PhantomData;
+
+/*
+Example memory table:
```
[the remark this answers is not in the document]
miha-stopar: Yes, another thing I plan to do in #24. I plan to extend the code there to check memory and stack ops and make a proper example table. Would be that ok?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+            //      - values[0] == [0]
+            //      - flags[0] == 1
+            //      - global_counters[0] == 0
+
+            // TODO: global_counter can't be 0 for both - memory and stack init rows
+
+            let address = meta.query_advice(address, Rotation::cur());
+            let one = Expression::Constant(F::one());
+            let max_address = Expression::Constant(F::from_u64(ADDRESS_MAX as u64));
+            // For stack init operation: ADDRESS_INCR is false (thus 0) and the address must be ADDRESS_MAX (that is 1023)
```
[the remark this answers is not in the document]
miha-stopar: ADDRESS_INCR = 0 means decreasing, ADDRESS_INCR = 1 means decreasing. However, I talked with Han about this and the plan is to have stack and memory addresses both increasing. I planned to make these changes in #24 though because there will be some changes related to this anyway.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+
+                let last_address = ops.last().unwrap().address;
+                let last_gc = ops.last().unwrap().global_counters.last().unwrap();
+                let mut last_counter = last_gc.as_ref().unwrap().global_counter().0;
+
+                // We pad all remaining memory rows to avoid the check at the first unused row.
+                // Without padding, (address_cur - address_prev) would not be zero at the first unused row
+                // and some checks would be triggered.
+                for i in offset..MAX_MEMORY_ROWS {
+                    self.q_not_first.enable(&mut region, i)?;
```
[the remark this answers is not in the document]
miha-stopar: Yes, will change this tomorrow.

src/state_circuit/state.rs
miha-stopar: Removed now in this PR.

src/state_circuit/state.rs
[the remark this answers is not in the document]
miha-stopar: Changed.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,878 @@
...
+                q_first * init_address,
+            ]
+        });
+
+        meta.create_gate("State operation", |meta| {
+            let q_not_first = meta.query_selector(q_not_first);
+            // If address_cur != address_prev, this is an `init`. We must constrain:
+            //      - values[0] == [0]
+            //      - flags[0] == 1
+            //      - global_counters[0] == 0
```
gaswhat, line 214: Why does `global_counter` need to be 0 for a new address? Isn't `global_counter` supposed to keep increasing across all addresses?

spartucus: For a new call context, the `global_counter` should be init with 0.

gaswhat: @spartucus This is not a new call context. It just moves to check the next memory/stack address.

therealyingtong: The `init` at each new address is effectively a "dummy write" of `value = 0`. This handles the case where the first operation at an address is a `read`. Since this address has never been written to, the first `read` should return 0. (This "dummy write" isn't a real operation and doesn't affect the `global_counter`.)

In terms of circuit constraints: note that the `q_read` case  involves querying the value on the previous row `value_prev`. Without a dummy init row, the first read would end up querying the last value from the previous address.

(Thanks @han0110 for explaining this to me.)

miha-stopar: Thanks @therealyingtong!

I asked @han0110 a couple of related questions related - global_counter can be set to 0 in the init operation as this is not put in the bus mapping. We don't check global_counter uniqueness in the state circuit - this will be done in the evm circuit. In the evm circuit, we will (1) lookup global_counter one by one and (2) check the degree of bus mapping is equal to the number of operations. If a malicious prover skips a write, (1) will fail. If a malicious prover inserts a non-existing write, (2) fails.

@gaswhat: I think global_counter doesn't need to be increasing across all addresses, it only needs to be ordered increasingly for each address.

gaswhat: Thanks for the explanation @therealyingtong @miha-stopar. I missed the `init` row at each new address previously.

I have a follow up question. Could we skip this `init` row for each new address to save some rows? It seems to me a bit redundant. I think we just need to prove the first row of a new address must be a `write` operation. Should be enough?

barryWhiteHat:
@miha-stopar to make memory byte addressable is all that we need to do is apply a range check on every element we write to memory being 1 byte ? 

If we do this we know that each address maps to a single byte. We would also have to change mstore to access 32 bus mapping element in evm proof.

miha-stopar:
> @miha-stopar to make memory byte addressable is all that we need to do is apply a range check on every element we write to memory being 1 byte ?
> 
> If we do this we know that each address maps to a single byte. We would also have to change mstore to access 32 bus mapping element in evm proof.

@barryWhiteHat yes, I plan to add this constraint as it's in the specs. Was having some problems with switching between memory and stack operations - I still have a couple of things to add, but it works now.

src/state_circuit/state.rs
miha-stopar: @gaswhat I removed stack init rows in #1bdbe5d. This requires some more code and some more gates, but indeed we save quite some rows. I left the memory init rows in there as they seem to be required. I will try to summarize @han0110 explanation below:

Memory init rows are needed to mimic the [EVM interpreter's behaviour](https://github.com/ethereum/go-ethereum/blob/master/core/vm/interpreter.go#L251-L253). For example, when memory is empty and mload(32) is executed (which returns memory[32..64]), EVM expands memory size to 64 and returns 0. Thus, in state circuit we allocate to 0 all addresses that are used to be compatible with EVM. 

We don't have to do such things for stack operations.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,896 @@
...
+------------------------------------------------------
+|  12  |      1      |        0       |      12      |
+|  17  |      1      |        1       |      32      |
+|  24  |      0      |        0       |      12      |
+|  89  |      0      |        1       |      32      |
+*/
+
+/// In the state proof, memory operations are ordered first by address, and then by global_counter.
+/// Memory is initialised at 0 for each new address.
+/// Memory is a word-addressed byte array, i.e. a mapping from a 253-bit word -> u8.
```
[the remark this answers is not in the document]
barryWhiteHat: I thought memory in evm was byte addressable?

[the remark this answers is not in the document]
barryWhiteHat: I think that when you use the compressed form the address will be a253 number its just the random linear combination of the 8 bit words making up the 256 bit address.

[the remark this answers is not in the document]
han0110: If we can assume a circuit memory size hard bound (say `2**24`), it would be easier to use actual value instead of random linear combination for memory index.

For example, when one is going to do like `mload(250)`, we will need to do memory bus mapping lookup in range (`250`, `250 + 1`, `250 + 2`, ..., `281`) from byte to byte. If we use random linear combination, then some of indexes would have the first byte which exceeds 256 and produces carry to the second byte, which makes the random linear combination harder to calculate (hard do that without requiring intermediate witness).

If we assume a hard bound, we just take first 3 bytes and recompose it to `2**24` value to lookup memory, which seems to be easier, but is incompatible with current evm which doesn't has any hard bound.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1241 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct Config<
+    F: FieldExt,
+    const GLOBAL_COUNTER_MAX: usize,
+    const MEMORY_ROWS_MAX: usize,
+    const MEMORY_ADDRESS_MAX: usize,
+    const STACK_ROWS_MAX: usize,
+    const STACK_ADDRESS_MAX: usize,
+> {
+    q_target: Column<Fixed>,
```
gaswhat, line 117: Since we are using `q_target` to represent first row/stack/memory/storage, should we get rid of `q_first` and `q_not_first`?

miha-stopar: Yes, that's the idea. I already tried to remove these columns two but was having some problems. I will get back to this after I add the remaining constraints. However, using only q_target would probably require some more lookups because when you use q_not_first you know it's 0 or 1 and you don't need to normalize the value to 1 in range checks. On the other hand, q_target could be different values and we need to be careful (we need to normalize it) when checking whether this value multiplied by some other value is in some range. I plan to dig a bit into what's more efficient.

miha-stopar:
> > @miha-stopar to make memory byte addressable is all that we need to do is apply a range check on every element we write to memory being 1 byte ?
> > If we do this we know that each address maps to a single byte. We would also have to change mstore to access 32 bus mapping element in evm proof.
> 
> @barryWhiteHat yes, I plan to add this constraint as it's in the specs. Was having some problems with switching between memory and stack operations - I still have a couple of things to add, but it works now.

Now added in https://github.com/appliedzkp/zkevm-circuits/pull/23/commits/9b993a44a467ee01ab8739c41afbe2bdc8703d41.

src/state_circuit/state.rs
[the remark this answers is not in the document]
barryWhiteHat: Yes this makes sense. Hard bound means we can't overflow 2**253 memory elements. 

I was thinking if we used compressed form we would be able to avoid a decompression of the index when it comes from the stack. But you have to decompress to get the ordering right anyway so probably good to use the decompressed form.

src/state_circuit/state.rs
miha-stopar: @gaswhat q_first and and q_not_first have been removed and the number of lookups has been reduced. Hope it helps.

src/state_circuit/state.rs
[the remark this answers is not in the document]
DreamWuGit: 253 bit width range come from Pasta Curves ?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+/// Memory is a word-addressed byte array, i.e. a mapping from a 253-bit word -> u8.
+/// TODO: Confirm if we are using 253- or 256-bit memory addresses.
+#[derive(Copy, Clone, Debug)]
+struct Address<F: FieldExt>(F);
+
+/// Global counter
+#[derive(Copy, Clone, Debug)]
+struct GlobalCounter(usize);
+
+/// TODO: In the EVM we can only read memory values in 32 bytes, but can write either
```
DreamWuGit, line 70: I think this point need to confirm, reading memory values is not fixed 32 except MLOAD op code,  others read values base on actual size parameter.
Refer to https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go

src/state_circuit/state.rs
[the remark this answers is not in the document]
barryWhiteHat: It comes from the bn254 curve we are using. We replace Pasta curve with bn254 and halo2  polynomial commitment with kate polynomial commitment.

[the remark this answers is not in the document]
DreamWuGit: you mean code "pasta_curves::arithmetic::FieldExt;" has been with bn254 or happens on other branch ?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+
+    use pasta_curves::{arithmetic::FieldExt, pallas};
+
+    macro_rules! test_state_circuit {
+        ($k:expr, $global_counter_max:expr, $memory_rows_max:expr, $memory_address_max:expr, $stack_rows_max:expr, $stack_address_max:expr, $memory_ops:expr, $stack_ops:expr, $result:expr) => {{
+            #[derive(Default)]
+            struct StateCircuit<
+                F: FieldExt,
+                const GLOBAL_COUNTER_MAX: usize,
+                const MEMORY_ROWS_MAX: usize,
```
DreamWuGit, line 1045: does this define slot capability rows ?

miha-stopar: It is to have a fixed number of memory rows, a fixed number of stack rows (STACK_ROW_MAX), and a fixed number of storage rows (yet to come). This enables us to use a fixed column for q_target (it is always the same), otherwise we would need to use advice column and more checks would be needed.

src/state_circuit/state.rs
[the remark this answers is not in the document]
barryWhiteHat: I think it happened on another branch but that is our plan.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+            // q_memory is 4 when target is 2, we use 1/4 to normalize the value
+            let inv = F::from_u64(4_u64).invert().unwrap_or(F::zero());
+            let i = Expression::Constant(inv);
+
+            vec![(q_memory * i * value, memory_value_table)]
+        });
+
+        // padding is in the allowed range:
+        meta.lookup(|meta| {
+            let padding = meta.query_advice(padding, Rotation::cur());
```
DreamWuGit, line 577: also need to check q_target is in the allowed range ?

miha-stopar: No need as the fixed column is the same for all proofs (the verifying keys contain the commitment of fixed columns). So, for the proof it always holds:
- the first MEMORY_ROWS_MAX rows have q_target = 2 (except for the first one which has q_target = 1)
- the first STACK_ROWS_MAX rows have q_target = 3 (except for the first one which has q_target = 1)
- the first STORAGE_ROWS_MAX rows have q_target = 4 (except for the first one which has q_target = 1) (this is yet to be implemented)

DreamWuGit: it should be act as constant variables & polynominal of plonk

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+            let padding_value_table = meta.query_fixed(padding_value_table, Rotation::cur());
+
+            vec![(padding, padding_value_table)]
```
gaswhat, line 580: Why not just use `padding * (1 - padding)` so that we can save a lookup table?

miha-stopar: Tried both versions, but using lookup seems to be much faster.

src/state_circuit/state.rs
miha-stopar: Not sure what you mean - do you think something needs to be changed?

DreamWuGit: I mean fixed column 's meaning from backend plonk proof system side, that is OK to me now!

miha-stopar: Ok, cool!

src/state_circuit/state.rs
gaswhat: I see. That's interesting. If that's true, does it apply to Read/Write flag column as well since we also check whether it's 0/1?

src/state_circuit/state.rs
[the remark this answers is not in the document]
therealyingtong: The BN254 / KZG fork of halo2 is on @kilic's branch: https://github.com/kilic/halo2/tree/kzg

src/state_circuit/state.rs
therealyingtong: I think `q_target` should be an advice column, since we'll have different numbers of different targets in each proof instance.

therealyingtong (changes requested):

src/state_circuit/state.rs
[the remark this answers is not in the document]
DreamWuGit: @therealyingtong  I see & Thx

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+        if ops_num > max_rows {
+            panic!("too many operations for the specified fixed value of operations");
+        }
+
+        let mut bus_mappings: Vec<BusMapping<F>> = Vec::new();
+
+        let mut offset = start_offset;
+        for (index, op) in ops.iter().enumerate() {
+            let address = op.address;
+            let mut address_prev = ops.first().unwrap().address;
```
DreamWuGit, line 720: address_prev line  can move above outside "for" loop to reduce redundant assignment times according to ops.size()

src/state_circuit/state.rs
miha-stopar: @gaswhat Exactly, it applies to read/write flag too. Will do some more benchmarks on this though.

src/state_circuit/state.rs
miha-stopar: True, will change it. Tnx

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+            if index > 0 {
+                address_prev = ops[index - 1].address;
+            }
+
+            if target == 2 {
+                // memory ops have init row
+                if index == 0 {
+                    self.init(region, offset, address, 1)?;
+                } else {
```
DreamWuGit, line 729: index>0 seems no need to execute init operation ?

src/state_circuit/state.rs
miha-stopar: There can be different numbers of different targets - there is a max number of memory, max number of stack, and max number of storage operations, but any number of operations up to the max value can be used. I think one of the main reasons I used a fixed column is because this way we can easily avoid the first row problem (which doesn't have Rot:prev). Instead, we could have one fixed column which would set a special value only for the first row, and q_target as an additional advice column - do you think this makes sense? It would complicate the constraints a bit though - for example, the first row could then be either memory, stack, or storage.

CPerezz:
This will need to be rebased into `main` now that https://github.com/appliedzkp/zkevm-circuits/pull/41 was merged.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+                * (four - q_target);
+
+            let value = meta.query_advice(value, Rotation::cur());
+            let memory_value_table = meta.query_fixed(memory_value_table, Rotation::cur());
+
+            // q_memory is 4 when target is 2, we use 1/4 to normalize the value
+            let inv = F::from_u64(4_u64).invert().unwrap_or(F::zero());
+            let i = Expression::Constant(inv);
+
+            vec![(q_memory * i * value, memory_value_table)]
```
DreamWuGit, line 572: how about values in stack ops  ? seems they share the memory value column, if so,  don't need to check whether it is  q_memory ?

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+                )?;
+                offset += 1;
+            }
+
+            for (internal_ind, global_counter) in op.global_counters.iter().enumerate() {
+                if target == 2 {
+                    let bus_mapping =
+                        self.assign_per_counter(region, offset, address, global_counter, target)?;
+                    bus_mappings.push(bus_mapping);
+                    address_diff_is_zero_chip.assign(region, offset, Some(F::zero()))?;
```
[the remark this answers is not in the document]
miha-stopar: Removed, it is redundant as the default value is zero.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+                        if index == 0 {
+                            let bus_mapping = self.assign_per_counter(
+                                region,
+                                offset,
+                                address,
+                                global_counter,
+                                1,
+                            )?;
+                            bus_mappings.push(bus_mapping);
+                            // set some non-zero diff for the first stack op
```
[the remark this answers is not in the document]
miha-stopar: You are right, no need at all. It was used in some of the previous versions when the constraints were written differently, but now it was redundant, it is removed now.

src/state_circuit/state.rs
miha-stopar: We are checking only memory values (need to be < 256), stack values are being checked to be in some range.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+            // q_stack_not_first is 6 when target is 3, we use 1/6 to normalize the value
+            let inv = F::from_u64(6_u64).invert().unwrap_or(F::zero());
+            let i = Expression::Constant(inv);
+            let q_stack_not_first = q_target.clone()
+                * (q_target.clone() - one.clone())
+                * (q_target.clone() - two)
+                * (four - q_target)
+                * i;
+
+            let q_not_first = q_memory_not_first + q_stack_not_first;
```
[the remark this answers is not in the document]
miha-stopar: Yes, this was very much needed. Provided now. Thanks!

src/state_circuit/state.rs
miha-stopar: Added a benchmark to measure the performance of checking the value being binary using a gate or a lookup. I was wrong, the gate approach is actually faster. I changed the padding constraint accordingly. Thanks for suggestion, @therealyingtong!

src/state_circuit/state.rs
miha-stopar: Just checked it and saw why I put it there - when ops is empty, there is no ops.first(). Could be written differently, but it's not that important I guess.

src/state_circuit/state.rs
```diff
@@ -0,0 +1,1609 @@
...
+                * (q_target_next.clone() - one.clone())
+                * (q_target_next.clone() - two)
+                * (four - q_target_next);
+
+            let flag = meta.query_advice(flag, Rotation::cur());
+            let q_read = one - flag;
+
+            vec![
+                q_stack_first * q_read, // first stack op has to be write (flag = 1)
+            ]
```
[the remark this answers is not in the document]
miha-stopar: Changed.

therealyingtong (approved):
LGTM! some non-blocking nits and questions.

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -0,0 +1,1569 @@
...
+        });
+
+        meta.create_gate("First stack row operation", |meta| {
+            let q_stack_first = q_stack_first(meta);
+            let one = Expression::Constant(F::one());
+            let flag = meta.query_advice(flag, Rotation::cur());
+
+            vec![
+                q_stack_first * (one - flag), /* first stack op has to be
+                                               * write (flag = 1) */
```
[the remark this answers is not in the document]
miha-stopar: There are two tests that cover these two cases: memory_values doesn't have stack ops and stack_first_write doesn't have memory ops.

miha-stopar:
> LGTM! some non-blocking nits and questions.

Thanks @therealyingtong! I addressed most of the comments and left one question about inlining the closures.

PR #26, opened 2021-08-10
EVM Circuit and trait OpGadget

### LINKED ISSUE #9: [EVM] `ADD` example
Implement the `ADD` opcode in the EVM circuit.

This requires a bus mapping as input.

han0110:
This PR aims to provides:

1. A flexible design of evm circuit
2. A trait for op gadget for us to implmenent without caring about the actual circuit layout
3. An example `AddGadget` for reference (Closes #9)

Todo:

- [x] Trait `OpGadget`
- [x] Check `linear_combinations` configured by `OpGadget`
- [x] Check `lookups` configured by `OpGadget`
  - [x] Word 8bit lookups
  - [x] Bus mapping lookups for stack and memory
- [x] `AddGadget` and `PushGadget` as examples.
- [x] ~~Feed `ExecutionStep` by crate `bus-mapping` into circuit~~ -> Next PR

### DIFF
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -738,7 +738,7 @@ impl<F: FieldExt> EvmCircuit<F> {
                 }
 
                 // Range32
-                for idx in 0..31 {
+                for idx in 0..32 {
                     region.assign_fixed(
                         || "Range32: tag",
                         self.fixed_table[0],
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -3,7 +3,9 @@
 use halo2::{
     arithmetic::FieldExt,
     circuit::{self, Layouter, Region},
-    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Selector},
+    plonk::{
+        Advice, Column, ConstraintSystem, Error, Expression, Fixed, Selector,
+    },
     poly::Rotation,
 };
 use std::convert::TryInto;
@@ -12,7 +14,7 @@ mod op_execution;
 use op_execution::{OpExecutionGadget, OpExecutionState};
 
 #[derive(Clone, Debug)]
-enum CallField {
+pub(crate) enum CallField {
     // Transaction id.
     TxId,
     // Global counter at the begin of call.
@@ -48,7 +50,7 @@ enum CallField {
 }
 
 #[derive(Clone, Debug)]
-enum CallStateField {
+pub(crate) enum CallStateField {
     // Program counter.
     ProgramCounter,
     // Stack pointer.
@@ -68,7 +70,7 @@ enum CallStateField {
 }
 
 #[derive(Clone, Debug)]
-enum BusMappingLookup<F> {
+pub(crate) enum BusMappingLookup<F> {
     // Read-Write
     OpExecutionState {
         field: CallStateField,
@@ -77,8 +79,9 @@ enum BusMappingLookup<F> {
     },
     Stack {
         // One can only access its own stack, so id is no needed to be
-        // specified. Stack index is always deterministic respect to
-        // stack pointer, so an index offset is enough.
+        // specified.
+        // Stack index is always deterministic respect to stack pointer, so an
+        // index offset is enough.
         index_offset: i32,
         value: Expression<F>,
         is_write: bool,
@@ -106,15 +109,31 @@ enum BusMappingLookup<F> {
         field: CallField,
         value: Expression<F>,
     },
-    /* TODO: Block,
-     * TODO: Bytecode,
-     * TODO: Tx,
-     * TODO: TxCalldata, */
+    Bytecode {
+        hash: Expression<F>,
+        index: Expression<F>,
+        value: Expression<F>,
+    },
+    // TODO: Block,
+    // TODO: Bytecode,
+    // TODO: Tx,
+    // TODO: TxCalldata,
+}
+
+#[derive(Clone, Copy, Debug, PartialOrd, PartialEq, Eq, Ord)]
+pub(crate) enum FixedLookup {
+    // Noop provides [0, 0, 0, 0] row
+    Noop,
+    // meaningful tags start with 1
+    Range256,
+    BitwiseAnd,
+    BitwiseOr,
+    BitwiseXor,
 }
 
 #[derive(Clone, Debug)]
-enum Lookup<F> {
-    RangeLookup(/* usize, Expression<F> */),
+pub(crate) enum Lookup<F> {
+    FixedLookup(FixedLookup, [Expression<F>; 3]),
     BusMappingLookup(BusMappingLookup<F>),
 }
 
@@ -269,6 +288,7 @@ pub(crate) struct ExecutionStep {
 #[derive(Clone)]
 struct EvmCircuit<F> {
     selector: Selector,
+    fixed_table: [Column<Fixed>; 4],
     op_execution_gadget: OpExecutionGadget<F>,
 }
 
@@ -282,6 +302,18 @@ impl<F: FieldExt> EvmCircuit<F> {
         let columns = (0..Self::WIDTH)
             .map(|_| meta.advice_column())
             .collect::<Vec<_>>();
+
+        // fixed_table contains pre-built tables identified by tag including:
+        // - different size range tables
+        // - bitwise table
+        // - comparator table
+        // - ...
+        let fixed_table = [
+            meta.fixed_column(), // tag
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+        ];
 
         let mut selector_exp = Expression::Constant(F::zero());
         let mut cells = Vec::with_capacity(Self::WIDTH * Self::HEIGHT);
@@ -322,28 +354,155 @@ impl<F: FieldExt> EvmCircuit<F> {
         let op_execution_state_next =
             OpExecutionState::new(op_execution_state_next_cells);
 
+        // independent_lookups collect lookups by independent selectors, which
+        // means we can sum some of them together to save lookups.
+        let mut independent_lookups =
+            Vec::<(Expression<F>, Vec<Lookup<F>>)>::new();
+
         let op_execution_gadget = OpExecutionGadget::configure(
             meta,
+            r,
             selector_exp,
             op_execution_state_curr,
             op_execution_state_next,
             free_cells,
-            r,
+            &mut independent_lookups,
         );
 
         // TODO: call_initialization_gadget
 
-        // TODO: handle lookups
+        Self::configure_lookups(meta, &fixed_table, independent_lookups);
 
         EvmCircuit {
             selector,
+            fixed_table,
             op_execution_gadget,
         }
     }
 
+    fn configure_lookups(
+        meta: &mut ConstraintSystem<F>,
+        fixed_table: &[Column<Fixed>; 4],
+        independent_lookups: Vec<(Expression<F>, Vec<Lookup<F>>)>,
+    ) {
+        // TODO: call_lookups
+        // TODO: bytecode_lookups
+        // TODO: rw_lookups
+
+        let mut fixed_lookups = Vec::<[Expression<F>; 4]>::new();
+
+        for (lookup_selector, lookups) in independent_lookups {
+            let mut fixed_lookup_count = 0;
+
+            for lookup in lookups {
+                match lookup {
+                    Lookup::FixedLookup(tag, exps) => {
+                        let tag_exp =
+                            Expression::Constant(F::from_u64(tag as u64));
+
+                        if fixed_lookups.len() == fixed_lookup_count {
+                            fixed_lookups.push(
+                                [tag_exp]
+                                    .iter()
+                                    .chain(exps.iter())
+                                    .map(|exp| {
+                                        lookup_selector.clone() * exp.clone()
+                                    })
+                                    .collect::<Vec<_>>()
+                                    .try_into()
+                                    .unwrap(),
+                            );
+                        } else {
+                            for (acc, exp) in fixed_lookups[fixed_lookup_count]
+                                .iter_mut()
+                                .zip([tag_exp].iter().chain(exps.iter()))
+                            {
+                                *acc = acc.clone()
+                                    + lookup_selector.clone() * exp.clone();
+                            }
+                        }
+                        fixed_lookup_count += 1;
+                    }
+                    Lookup::BusMappingLookup(_) => {
+                        // TODO:
+                        unimplemented!()
+                    }
+                }
+            }
+        }
+
+        for fixed_lookup in fixed_lookups.iter() {
+            meta.lookup(|meta| {
+                fixed_lookup
+                    .iter()
+                    .zip(fixed_table.iter())
+                    .map(|(exp, fixed)| {
+                        (exp.clone(), meta.query_fixed(*fixed, Rotation::cur()))
+                    })
+                    .collect::<Vec<_>>()
+            });
+        }
+    }
+
+    fn load_fixed_tables(
+        &self,
+        layouter: &mut impl Layouter<F>,
+    ) -> Result<(), Error> {
+        layouter.assign_region(
+            || "fixed table",
+            |mut region| {
+                let mut offest = 0;
+
+                // Noop
+                for (idx, column) in self.fixed_table.iter().enumerate() {
+                    region.assign_fixed(
+                        || format!("Noop: {}", idx),
+                        *column,
+                        offest,
+                        || Ok(F::zero()),
+                    )?;
+                }
+                offest += 1;
+
+                // Range256
+                for idx in 0..256 {
+                    region.assign_fixed(
+                        || "Range256: tag",
+                        self.fixed_table[0],
+                        offest,
+                        || Ok(F::from_u64(FixedLookup::Range256 as u64)),
+                    )?;
+                    region.assign_fixed(
+                        || "Range256: value",
+                        self.fixed_table[1],
+                        offest,
+                        || Ok(F::from_u64(idx as u64)),
+                    )?;
+                    for (idx, column) in
+                        self.fixed_table[2..].iter().enumerate()
+                    {
+                        region.assign_fixed(
+                            || format!("Range256: padding {}", idx),
+                            *column,
+                            offest,
+                            || Ok(F::zero()),
+                        )?;
+                    }
+                    offest += 1;
+                }
+
+                // TODO: BitwiseAnd
+                // TODO: BitwiseOr
+                // TODO: BitwiseXor
+
+                Ok(())
+            },
+        )
+    }
+
     fn assign(
         &self,
-        mut layouter: impl Layouter<F>,
+        layouter: &mut impl Layouter<F>,
         execution_steps: &[ExecutionStep],
     ) -> Result<(), Error> {
         layouter.assign_region(
@@ -428,9 +587,12 @@ mod test {
         fn synthesize(
             &self,
             config: Self::Config,
-            layouter: impl Layouter<F>,
+            mut layouter: impl Layouter<F>,
         ) -> Result<(), Error> {
-            config.evm_circuit.assign(layouter, &self.execution_steps)
+            config.evm_circuit.load_fixed_tables(&mut layouter)?;
+            config
+                .evm_circuit
+                .assign(&mut layouter, &self.execution_steps)
         }
     }
 }
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -1,6 +1,6 @@
 //! The EVM circuit implementation.
 
-use bus_mapping::operation::Target;
+use bus_mapping::{evm::OpcodeId, operation::Target};
 use halo2::{
     arithmetic::FieldExt,
     circuit::{self, Layouter, Region},
@@ -302,7 +302,7 @@ enum Case {
 
 // TODO: use ExecutionStep from bus_mapping
 pub(crate) struct ExecutionStep {
-    opcode: u8,
+    opcode: OpcodeId,
     case: Case,
     values: Vec<[u8; 32]>,
 }
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -16,7 +16,7 @@ use std::{convert::TryInto, iter};
 mod op_execution;
 use op_execution::{OpExecutionGadget, OpExecutionState};
 mod param;
-use param::{CIRCUIT_HEIGHT, CIRCUIT_WIDTH, NUM_CELL_OP_EXECTION_STATE};
+use param::{CIRCUIT_HEIGHT, CIRCUIT_WIDTH, NUM_CELL_OP_EXECUTION_STATE};
 
 #[derive(Clone, Debug)]
 pub(crate) enum CallField {
@@ -46,7 +46,7 @@ pub(crate) enum CallField {
     // Value in wei of call.
     Value,
     // If call succeeds or not in the future.
-    IsPersistant,
+    IsPersistent,
     // If call is within a static call.
     IsStatic,
     // If call is `CREATE*`. We lookup op from calldata when is create,
@@ -173,8 +173,8 @@ struct Constraint<F> {
 pub(crate) struct Cell<F> {
     // expression for constraint
     expression: Expression<F>,
-    // relative position to selector for synthesis
     column: Column<Advice>,
+    // relative position to selector for synthesis
     rotation: usize,
 }
 
@@ -460,7 +460,7 @@ impl<F: FieldExt> EvmCircuit<F> {
             vec![0.expr()]
         });
 
-        let num_cells_next_asseccible = NUM_CELL_OP_EXECTION_STATE;
+        let num_cells_next_asseccible = NUM_CELL_OP_EXECUTION_STATE;
 
         let cells_next =
             &mut cells_curr[..num_cells_next_asseccible].to_vec()[..];
@@ -477,11 +477,11 @@ impl<F: FieldExt> EvmCircuit<F> {
         });
 
         let op_execution_state_curr =
-            OpExecutionState::new(&cells_curr[..NUM_CELL_OP_EXECTION_STATE]);
+            OpExecutionState::new(&cells_curr[..NUM_CELL_OP_EXECUTION_STATE]);
         let op_execution_state_next =
-            OpExecutionState::new(&cells_next[..NUM_CELL_OP_EXECTION_STATE]);
+            OpExecutionState::new(&cells_next[..NUM_CELL_OP_EXECUTION_STATE]);
         let op_execution_free_cells =
-            cells_curr[NUM_CELL_OP_EXECTION_STATE..].to_vec();
+            cells_curr[NUM_CELL_OP_EXECUTION_STATE..].to_vec();
 
         let mut qs_op_execution = 0.expr();
         meta.create_gate(
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -10,6 +10,7 @@ use halo2::{
     },
     poly::Rotation,
 };
+use num::BigUint;
 use std::{convert::TryInto, iter};
 
 mod op_execution;
@@ -167,6 +168,7 @@ struct Constraint<F> {
     polys: Vec<Expression<F>>,
     lookups: Vec<Lookup<F>>,
 }
+
 #[derive(Clone, Debug)]
 pub(crate) struct Cell<F> {
     // expression for constraint
@@ -208,7 +210,7 @@ impl<F: FieldExt> Expr<F> for Cell<F> {
 struct Word<F> {
     // random linear combination expression of cells
     expression: Expression<F>,
-    // inner cells for syntesis
+    // inner cells for synthesis
     cells: [Cell<F>; 32],
 }
 
@@ -309,7 +311,7 @@ enum Case {
 pub(crate) struct ExecutionStep {
     opcode: OpcodeId,
     case: Case,
-    values: Vec<[u8; 32]>,
+    values: Vec<BigUint>,
 }
 
 // TODO: use Operation from bus_mapping
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -145,10 +145,9 @@ pub(crate) enum Lookup<F> {
 struct Constraint<F> {
     name: &'static str,
     selector: Expression<F>,
-    linear_combinations: Vec<Expression<F>>,
+    polys: Vec<Expression<F>>,
     lookups: Vec<Lookup<F>>,
 }
-
 #[derive(Clone, Debug)]
 pub(crate) struct Cell<F> {
     // expression for constraint
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -158,7 +158,7 @@ pub(crate) struct Cell<F> {
 }
 
 impl<F: FieldExt> Cell<F> {
-    fn exp(&self) -> Expression<F> {
+    fn expr(&self) -> Expression<F> {
         self.expression.clone()
     }
 
@@ -199,13 +199,13 @@ impl<F: FieldExt> Word<F> {
                 .iter()
                 .rev()
                 .fold(Expression::Constant(F::zero()), |acc, byte| {
-                    acc * r.clone() + byte.exp()
+                    acc * r.clone() + byte.expr()
                 }),
             cells: cells.to_owned().try_into().unwrap(),
         }
     }
 
-    fn exp(&self) -> Expression<F> {
+    fn expr(&self) -> Expression<F> {
         self.expression.clone()
     }
 
@@ -439,7 +439,7 @@ impl<F: FieldExt> EvmCircuit<F> {
             "Query synthetic selector for OpExecutionGadget",
             |meta| {
                 qs_op_execution = meta.query_selector(q_step)
-                    * op_execution_state_curr.is_executing.exp();
+                    * op_execution_state_curr.is_executing.expr();
 
                 vec![Expression::Constant(F::zero())]
             },
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -155,7 +155,7 @@ pub(crate) struct Cell<F> {
     expression: Expression<F>,
     // relative position to selector for synthesis
     column: Column<Advice>,
-    offset: usize,
+    rotation: usize,
 }
 
 impl<F: FieldExt> Cell<F> {
@@ -172,12 +172,12 @@ impl<F: FieldExt> Cell<F> {
         region.assign_advice(
             || {
                 format!(
-                    "Cell {{ column: {:?}, offset: {}}}",
-                    self.column, self.offset
+                    "Cell column: {:?} and rotation: {}",
+                    self.column, self.rotation
                 )
             },
             self.column,
-            offset + self.offset,
+            offset + self.rotation,
             || value.ok_or(Error::SynthesisError),
         )
     }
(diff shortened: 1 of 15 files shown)


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,598 @@
...
+pub(crate) struct Cell<F> {
+    // expression for constraint
+    expression: Expression<F>,
+    // relative position to selector for synthesis
+    column: Column<Advice>,
+    offset: usize,
+}
+
+impl<F: FieldExt> Cell<F> {
+    fn exp(&self) -> Expression<F> {
```
gaswhat, line 158:
```suggestion
    fn expr(&self) -> Expression<F> {
```
Avoid the confusion with `exp` for exponential

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,447 @@
...
+            };
+            {$name:ident = $gadget:ident; $($tail:tt)+} => {
+                constrcut_op_gadget!{$name = $gadget};
+                op_selector_idx += 1;
+                constrcut_op_gadget!{$($tail)+};
+            };
+        }
+
+        constrcut_op_gadget! {
+            add_gadget = AddGadget;
```
DreamWuGit, line 201: add future gadget here also after implemented  ?

han0110: correct

zkevm-circuits/src/evm_circuit.rs
han0110: Thanks, `expr` sounds much better! Fixed in recent commit.

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,447 @@
...
+    r: F,
+    op_execution_state_curr: OpExecutionState<F>,
+    op_execution_state_next: OpExecutionState<F>,
+    op_selectors: Vec<Cell<F>>,
+    free_cells: Vec<Cell<F>>,
+    resumption: Resumption<F>,
+    // TODO: use OpcodeId from bus_mapping
+    op_selector_idx_map: HashMap<u8, usize>,
+    default_values_map: HashMap<(usize, Case), Vec<(usize, F)>>,
+    add_gadget: AddGadget<F>,
```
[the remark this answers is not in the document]
han0110: yes for now

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,443 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct OpExecutionState<F> {
+    is_executing: Cell<F>,
+    global_counter: Cell<F>,
+    call_id: Cell<F>,
+    program_counter: Cell<F>,
+    stack_pointer: Cell<F>,
+    gas_counter: Cell<F>,
```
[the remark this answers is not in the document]
han0110: `gas_counter` counts accumulated gas usage

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,443 @@
...
+
+mod arithmetic;
+use arithmetic::AddGadget;
+
+#[derive(Debug)]
+struct CaseConfig {
+    case: Case,
+    num_word: usize,
+    num_cell: usize,
+    will_resume: bool,
```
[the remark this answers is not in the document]
han0110: On the contrary, `will_resume` tells that if such case the call should terminate and resume to caller or next tx if it's a top level call. Maybe the name `will_resume` is bad

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,699 @@
...
+            vec![Expression::Constant(F::zero())]
+        });
+
+        let num_cells_next_asseccible = NUM_CELL_OP_EXECTION_STATE;
+
+        let cells_next =
+            &mut cells_curr[..num_cells_next_asseccible].to_vec()[..];
+        meta.create_gate("Query cells for next step", |meta| {
+            for cell in cells_next.iter_mut() {
+                cell.expression = meta.query_advice(
```
gaswhat, line 421: `cell.rotation` is not updated in the `cells_next`

gaswhat: I think it might be easier to just create new `Cell` for `cells_next`.

zkevm-circuits/src/evm_circuit/op_execution.rs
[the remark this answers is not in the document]
DreamWuGit: so better to use a vec to holds all op codes as list ?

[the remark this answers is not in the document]
han0110: Tried but found trait object doesn't support generic, any concrete example to make it a `Vec`?

zkevm-circuits/src/evm_circuit.rs
han0110: Thanks for catching this! Fixed in [`4c9dc9d`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/4c9dc9d33ac28d9f7eaac2af389d52aa02ad146e)

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,930 @@
...
+    // Last callee's unique identifier.
+    CalleeID,
+    // Offset of returndata.
+    ReturndataOffset,
+    // Size of returndata.
+    ReturndataSize,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum BusMappingLookup<F> {
```
CPerezz, line 76: Seems that this could be externalized to the bus mapping as most of the stuff could be parsed there and you would not need to define it here.
WDYT?

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,930 @@
...
+                })
+                .collect()
+        })
+    }
+}
+
+// TODO: Should we maintain these state in circuit or we prepare all of them
+// when parsing debug trace?
+#[derive(Debug)]
+pub(crate) struct CoreStateInstance {
```
CPerezz, line 249: I would prepare that during the debug step. So that reduces the complexity from this crate. And we have much more data avaliable in the trace than we have here when the trace has already been parsed.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,930 @@
...
+            call_id,
+            program_counter: 0,
+            stack_pointer: 1024,
+            gas_counter: 0,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+enum Case {
```
CPerezz, line 272: Can we get this info during the parsing of the trace? 
If so, I would externalize that too to `bus-mapping`.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,930 @@
...
+    values: Vec<[u8; 32]>,
+}
+
+// TODO: use Operation from bus_mapping
+pub(crate) struct Operation<F> {
+    gc: usize,
+    target: Target,
+    is_write: bool,
+    values: [F; 4],
+}
```
CPerezz, line 316: Should we use in this PR the real types? Or you want to wait until we parse all the info from geth? As long as it doesn't affect the Opcode, I don't really care. But we should file issues for that.

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,532 @@
...
+            program_counter: cells[3].clone(),
+            stack_pointer: cells[4].clone(),
+            gas_counter: cells[5].clone(),
+            opcode: cells[6].clone(),
+        }
+    }
+}
+
+trait OpGadget<F: FieldExt> {
+    const RESPONSIBLE_OPCODES: &'static [u8];
```
CPerezz, line 169: What is this exactly? Should we use types from `bus-mapping`?

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,532 @@
...
+
+    const CASE_CONFIGS: &'static [CaseConfig];
```
CPerezz, line 171: Should this be provided by an `ExecutionStep` or similar? Also, I don't really see how is possible to make this a const if it changes for every opcode. And between opcode instances.

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,532 @@
...
+    resumption: Resumption<F>,
+    // TODO: use OpcodeId from bus_mapping
+    qs_op_idx_map: HashMap<u8, usize>,
+    preset_map: HashMap<(usize, Case), Preset<F>>,
+    add_gadget: AddGadget<F>,
+    push_gadget: PushGadget<F>,
+}
+
+impl<F: FieldExt> OpExecutionGadget<F> {
+    // TODO: refactor input type
```
CPerezz, line 214: Can we file an issue for that?

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,532 @@
...
+        let mut qs_op_idx = 0;
+
+        let mut constraints = vec![Constraint {
+            name: "op selectors",
+            selector: Expression::Constant(F::one()),
+            polys: bool_switches_constraints(qs_ops),
+            lookups: vec![],
+        }];
+
+        macro_rules! construct_op_gadget {
```
CPerezz, line 243: We should document macros as much as possible.

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
```diff
@@ -0,0 +1,564 @@
...
+    swap: Cell<F>,
+    a: Word<F>,
+    b: Word<F>,
+    c: Word<F>,
+    carry: [Cell<F>; 32],
+}
+
+#[derive(Clone, Debug)]
+pub struct AddGadget<F> {
+    success: AddSuccessAllocation<F>,
```
CPerezz, line 25: To me this should also contain the data of one or multiple `ExecutionStep`s which contain in this case an `ADD` instruction.

zkevm-circuits/src/evm_circuit/op_execution/push.rs
```diff
@@ -0,0 +1,307 @@
...
+            num_cell: 0,
+            will_resume: true,
+        },
+        CaseConfig {
+            case: Case::OutOfGas,
+            num_word: 0,
+            num_cell: 0,
+            will_resume: true,
+        },
+    ];
```
CPerezz, line 54: This kinda reforces my previous argument that it doesn't seem to fit really well to have CaseConfigs as a const associated to the gadget since it can change depending on the instance.

CPerezz (commented):
Overall it looks good if what we plan is to have a Layout example.

But I have several questions:
1. Seems that we still need to integrate `bus-mapping` in here. Do we plan do do it in a follow-up PR?
2. The `Opcode` trait should (IMO) contain one or multiple `ExecutionStep`s so that we can basically have all of the `CaseConfig` data an all the other stuff extracted from there.
3. `Case` and `CaseConfig` as well as `CallContext` should probably be migrated to `bus-mappign` and obtained during the parsing process.
4. We should document more. Even if are not polished docs. But at least to help improve the general idea that the development path follows.
5. `CaseConfig` as said in the comments should probably not be an associated const and instead be the result of a fn or similar. The reason is that multiple instances of the same Opode trait object can result in different `CaseConfig`.

zkevm-circuits/src/evm_circuit.rs
han0110: This is only for circuit constraint configuration, it's fixed in circuit. The things needed to be externalized to `bus-mapping` are `ExecutionStep` and `Operation` you mentioned below, they are actually parsed from `bus-mapping`

zkevm-circuits/src/evm_circuit.rs
han0110: Yeah make sense, this is only for making the current work testable.

zkevm-circuits/src/evm_circuit.rs
han0110: Yep, we should move this to `bus-mapping`

zkevm-circuits/src/evm_circuit.rs
han0110: I tried but found I have to change a lot on `bus-mapping` side (export interface, add methods, etc...), which could take me some more days to finish this. So I plan to use a simple struct for this PR, and file an issue for what `evm-circuit` would need from `bus-mapping`.

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: This is the opcodes the gadget would handle. We should use the `OpcodeId` in `bus-mapping` to replace a all ambiguous `u8` which stands for opcode. Fixed in [`b50e4e9`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/b50e4e9e544fa5468e29ee6b2b44d487aa6fabf6)

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: But this is only for circuit constraint configuration usage, it tells how many cases this `OpGadget` could encounter, and `OpExecutionGadget` will allocate enough cells for it.

When prover is assigning witness, the `OpGadget` will assign correct values depends on the case provided by `ExecutionStep`.

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: Sure, there are also many other places using many inputs like this, we should probably group some of them in a more readable way.

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: Added in [`29b25cd`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/29b25cd033b82b7a4e2187418a5ae6d5715253ae)

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,930 @@
...
+                            *column,
+                            offset,
+                            || Ok(F::zero()),
+                        )?;
+                    }
+                    offset += 1;
+                }
+
+                // Range32
+                for idx in 0..31 {
```
han0110, line 741: Thanks to @therealyingtong to point out the range error here.

```suggestion
                for idx in 0..32 {
```

zkevm-circuits/src/evm_circuit.rs
CPerezz: Perfect. Should we file an issue for that?

zkevm-circuits/src/evm_circuit.rs
CPerezz: Should we add an issue for that?

zkevm-circuits/src/evm_circuit.rs
CPerezz: Perfect! Let's file an issue bofore merging then! :smile:

zkevm-circuits/src/evm_circuit/op_execution.rs
CPerezz: >But this is only for circuit constraint configuration usage, it tells how many cases this OpGadget could encounter, and 
> OpExecutionGadget will allocate enough cells for it.

This sounds  more to me like an enum than rather an slice. But if you think that is the most practical way, I'm fine for now! :)

zkevm-circuits/src/evm_circuit/op_execution.rs
CPerezz: Sounds nice. Let's do so before merging!

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
CPerezz: Or should have implemented a way on which: `Provided a slice of `ExecutionStep`s related to the addition, witness all of the data and build all the lookups/constraints.

han0110: But the lookups and constraints should not depend on proving time witness, it's fixed when keygen, right?

zkevm-circuits/src/evm_circuit.rs
han0110: Mentioned in #61

zkevm-circuits/src/evm_circuit.rs
han0110: Mentioned in #61

zkevm-circuits/src/evm_circuit.rs
han0110: Mentioned in #61

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: Ah, I think I got your point, you mean we should separate each case as a unique `OpGadget`?

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
```diff
@@ -0,0 +1,566 @@
...
+            offset,
+            Some(execution_step.values[1]),
+        )?;
+        self.success.c.assign(
+            region,
+            offset,
+            Some(execution_step.values[2]),
+        )?;
+        self.success
+            .carry
```
gaswhat, line 289: Should we calculate the `carry` in the `assign_success` instead of getting it from caller?

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
```diff
@@ -0,0 +1,577 @@
...
+                    add_constraints,
+                ]
+                .concat(),
+                lookups: bus_mapping_lookups,
+            }
+        };
+
+        let stack_underflow = {
+            let (zero, minus_one) = (
+                Expression::Constant(F::from_u64(1024)),
```
DreamWuGit, line 197: stack pointer should range from 0...1023,  check 1024 is redundant ?

CPerezz: I think you got a point there! @han0110 It seems correct to me. We should have 1023 as max.

zkevm-circuits/src/evm_circuit.rs
CPerezz: My bad. Thanks for the explanation!

zkevm-circuits/src/evm_circuit/op_execution.rs
CPerezz: Exactly. But you'll know better what fits there.

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
han0110: In evm circuit, we have `stack_pointer` initialized at 1024, which represents it's empty. For handling `ADD`, we need to pop 2 values from stack, which means we have 2 kinds of stack underflow condition: when stack is empty or stack only contains 1 item. We need to check both conditions, right?

DreamWuGit: seems right that stack_pointer initialized at 1024, which represents point to  empty state, pointing to 1023 means one item in the stack,  do we have code to ensure initialized  stack_pointer?

CPerezz: Well I thought that the stack has 1024 elements. Therefore, we check range [0-1023] which is 1024 elements inside since 0 counts.

We also work on the same way in the Stack repr of the crate. The stack pointer points at the beginning of the execution to 1023 and goes down to 0 (being 0 the last slot to allocate in the stack).

han0110: `CallInit` I am currently working on initializes `OpExecutionState`.

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
```diff
@@ -0,0 +1,577 @@
...
+        }
+    }
+
+    fn constraints(
+        &self,
+        state_curr: &OpExecutionState<F>,
+        state_next: &OpExecutionState<F>,
+    ) -> Vec<Constraint<F>> {
+        let (add, sub) = (
+            Expression::Constant(F::from_u64(1)),
```
DreamWuGit, line 89: use OpcodeId::ADD , OpcodeId::Sub here to replace hard code?

han0110: Refactored in [`f0bd8af`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/f0bd8afe2607abfb428c5fb171730625448ddd25). Thanks!

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+
+impl<F: FieldExt> Expr<F> for FixedLookup {
+    fn expr(&self) -> Expression<F> {
+        Expression::Constant(F::from_u64(*self as u64))
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum Lookup<F> {
+    FixedLookup(FixedLookup, [Expression<F>; 3]),
```
DreamWuGit, line 159: what does three expressions mean ? one is not enough ?

han0110: The `FixedLookup` is the global 3-column fixed table with an extra tag column. It's used for range lookup, bitwise lookup, or anything linear-combination unfriendly.

Naively 3-columns table suits most common used table for 2 input and 1 output. We also put range table here to save columns, we might found the 3-columns table is not enough anymore, then we separate them out to another 1-column table to not waste cells.

CPerezz (approved):
LGTM. 

We will need to address all of the issues that have been raised here BTW. Not sure if before or after the first opcode impls. But the more we add the bigger the refactor will become.

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
CPerezz: The values you place on the lookups should no??

I think that you got a point in the sense iof that this just states the logic on which the allocation is done, not the actual witness-related actions.

miha-stopar (commented):
The ideas used and the implementation both seem very nice to me. I was thinking about possible optimizations but nothing came to my mind that hasn't already been considered. I have one question though - would we need one additional constraint to make sure that execution_step.opcode and the opcode selector (talking about 80 cells) are in sync? Wouldn't be now possible to set execution_step.opcode to ADD and set the opcode selector to PUSH (the one of 80s that corresponds to PUSH) which would make some attacks possible?

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+                            ..
+                        } = &op_execution_state_curr;
+
+                        if rw_lookups.len() == rw_lookup_count {
+                            rw_lookups
+                                .push(vec![0.expr(); 7].try_into().unwrap());
+                        }
+
+                        let mut exprs = vec![
+                            global_counter.expr() + rw_lookup_count.expr(),
```
DreamWuGit, line 568: not sure why need to plus rw_lookup_count.expr() ?  if so , even the same buss mapping lookup will make different exprs .

zkevm-circuits/src/evm_circuit/op_execution.rs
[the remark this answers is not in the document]
han0110: Will try trait with `F` as associated type in the future work to see if it works better

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
han0110: Yes, but the witness will be fed in fn `assign_*`, the gadget only needs to memorize which cell to assign it, right?

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
han0110: Make sense, I used to think **auxiliary witness** should be generated in gadget instead of being fed from caller, but it also hard to see which are auxiliary or not before bus mapping has a stable structure. IMO it's ok for now cause it should be trivial to adjust if we need to generate the `carry` here.

han0110:
> I have one question though - would we need one additional constraint to make sure that execution_step.opcode and the opcode selector (talking about 80 cells) are in sync? Wouldn't be now possible to set execution_step.opcode to ADD and set the opcode selector to PUSH (the one of 80s that corresponds to PUSH) which would make some attacks possible?

@miha-stopar According to your question, currently in `AddGadget` and `PushGadget` we check if `opcode == 1 or opcode == 3` and `0 <= opcode - PUSH1 <= 31` respectively to keep them in sync. If we set opcode to `ADD`, the constraint in `PushGadget` should not pass, right?

zkevm-circuits/src/evm_circuit.rs
han0110: We might have several `rw_lookup` in single step, but since global counter serves as the order of access, we should ensure its uniqueness property. Otherwise we have no way to check if a malicious prover skips or inserts some write in bus mapping to cheat.

> even the same bus mapping lookups will make different exprs.

Why does it have same lookups in single step?

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
gaswhat: sounds good.

miha-stopar:
@han0110 yes, that's ok. What I meant are these assignments: https://github.com/appliedzkp/zkevm-circuits/blob/f0bd8afe2607abfb428c5fb171730625448ddd25/zkevm-circuits/src/evm_circuit/op_execution.rs#L484. Not sure if there is a check that these assignments correspond to the actual opcode?

zkevm-circuits/src/evm_circuit.rs
DreamWuGit: I tested same lookups in single step  condition, so be aware of it.  avoiding same lookups in single step seems make sense.

miha-stopar:
@han0110 I can see it now - opcode number cell is used only as kind of a hint and the 80 cells determine the branching. Sorry, all good!

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+            .try_into()
+            .unwrap();
+
+        // TODO: consider encode first 3 information into one field element
+        // rw_table contains read-write data in bus mapping for lookup
+        let rw_table = [
+            meta.advice_column(), // global_counter
+            meta.advice_column(), // target
+            meta.advice_column(), // is_write
+            meta.advice_column(), // val1
```
gaswhat, line 348: Add comment for actual meaning.

```suggestion
            meta.advice_column(), // val1, call_id
```

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+            meta.advice_column(), // val2
```
gaswhat, line 349:
```suggestion
            meta.advice_column(), // val2, address
```

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+            meta.advice_column(), // val3
```
gaswhat, line 350:
```suggestion
            meta.advice_column(), // val3, byte value
```

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,916 @@
...
+            meta.advice_column(), // val4
```
gaswhat, line 351:
```suggestion
            meta.advice_column(), // val4, reserved
```

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,524 @@
...
+        32 * self.num_word
+            + self.num_cell
+            + if self.will_halt {
+                NUM_CELL_RESUMPTION
+            } else {
+                0
+            }
+    }
+
+    // allocate indexes of cells for words, cells and unused. It assumes input
```
gaswhat, line 58:
```suggestion
    // allocate indices of cells for words, cells and unused. It assumes input
```

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,524 @@
...
+    // free_cells are in order of rotation and then column index.
+    fn allocation_idxs<F: FieldExt>(
```
gaswhat, line 60: Just a better function name suggestion. 

```suggestion
    fn allocate_indices<F: FieldExt>(
```

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,524 @@
...
+struct CaseAllocation<F> {
+    selector: Cell<F>,
+    words: Vec<Word<F>>,
+    cells: Vec<Cell<F>>,
+    resumption: Option<Resumption<F>>,
+}
+
+#[derive(Clone, Debug)]
+struct Resumption<F> {
+    caller_id: Cell<F>,
```
gaswhat, line 126: Use `call_id` to be consistent with `call_id` in the `OpExecutionState`?

```suggestion
    call_id: Cell<F>,
```

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
```diff
@@ -0,0 +1,553 @@
...
+                swap.expr() * no_swap.clone(),
+                swap.expr() * (opcode.expr() - OpcodeId::SUB.expr()),
+                no_swap.clone() * (opcode.expr() - OpcodeId::ADD.expr()),
+            ];
+
+            // add constraints
+            let add_constraints = {
+                let mut constraints = Vec::with_capacity(32);
+                // 256 * carry_out + c
+                let lhs = carry[0].expr() * 256.expr() + c.cells[0].expr();
```
gaswhat, line 126: Do we need to prove the `carry` belongs to [0, 1]? I wonder if we could find a value other than 1 such that x * 256 = 256 under Fp region.

gaswhat (approved):
LGTM. Just a few minor comments

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
han0110: Thanks for catching this! Fixed in [`628bafa`](https://github.com/appliedzkp/zkevm-circuits/commit/628bafa3c9713ccb9e6f353b998722383091db02).

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: This is intended for caller's id instead of current call's id, call should resume to caller's execution after halting if it's a internal call. Will add more document on this when I really use it in the cases which halt.

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: Thanks, `allocate_indices` sounds better. Updated in [`ce345d5`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/ce345d5b26172555e678910b2e36c09389f022f9).

zkevm-circuits/src/evm_circuit/op_execution.rs
gaswhat: I see. Thanks for the clarification.

zkevm-circuits/src/evm_circuit.rs
han0110: These `val*` mean different things depends on the target, so I left it arbitrary instead of concrete meaning. For stack or memroy it would be call_id, but for storage it would be account address. Will try to document it in a better way to clarify.

therealyingtong (commented):
I've not looked through this whole PR and am just submitting a few nits -- please don't block on my review.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1 +1,918 @@
...
+            vec![0.expr()]
+        });
+
+        // TODO: Enable this when switch to kzg branch and remove the failure
+        //       VerifyFailure::ConstraintPoison
+        // meta.create_gate("Query byte lookup as a synthetic selector", |meta| {
+        //     let expr = meta.query_advice(qs_byte_lookup, Rotation::cur());
+
+        //     vec![expr.clone() * (Expression::Constant(F::one()) - expr)]
+        // });
```
ChihChengLiang, line 446: Nitpick: let's move this to a new issue. Avoid commit commented code.

```suggestion
```

zkevm-circuits/src/evm_circuit/op_execution/push.rs
```diff
@@ -0,0 +1,337 @@
...
+                //             index: selector.expr()
+                //                 * (program_counter.expr()
+                //                     + one.clone()
+                //                     + Expression::Constant(F::from_u64(
+                //                         idx as u64,
+                //                     ))),
+                //             value: cell.expr(),
+                //         })
+                //     })
+                //     .collect::<Vec<_>>(),
```
ChihChengLiang, line 182: Nitpick: move this to a new issue too. Avoid commit commented code.
```suggestion
```

ChihChengLiang (approved):
LGTM. Added some nitpicks

zkevm-circuits/src/evm_circuit/op_execution/push.rs
han0110: Fixed in [`26019ee`](https://github.com/appliedzkp/zkevm-circuits/pull/26/commits/26019ee70b8a7ee231d6faf4275722ef92105076)

zkevm-circuits/src/evm_circuit.rs
han0110: Fixed in [`8063d88`](https://github.com/appliedzkp/zkevm-circuits/commit/8063d888c9db7e47b9f3e8088570048f0f8201a6).

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -0,0 +1,524 @@
...
+                    &mut preset_map,
+                    &mut constraints,
+                );
+                qs_op_idx += 1;
+            };
+        }
+
+        construct_op_gadget!(add_gadget);
+        construct_op_gadget!(push_gadget);
+        let _ = qs_op_idx;
```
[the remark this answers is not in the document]
han0110: This is intended for escaping the clippy warning since the macro increases `qs_op_idx` even it's the last one.

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -0,0 +1,12 @@
+// FIXME: decide a optimal circuit dimension in the future
+// Circuit dimension
+// NOTE: CIRCUIT_WIDTH should be divide 32 evenly for whole row byte lookup
+pub const CIRCUIT_WIDTH: usize = 32;
+pub const CIRCUIT_HEIGHT: usize = 10;
+
+// Number of cells used for each purpose
+// TODO: pub const NUM_CELL_CALL_INITIALIZATION_STATE: usize = ;
+pub const NUM_CELL_OP_EXECTION_STATE: usize = 7;
```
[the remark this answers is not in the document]
han0110: Fixed with all above suggestions in [`702039d`](https://github.com/appliedzkp/zkevm-circuits/commit/702039d8aa118fbe8a6381e326163d171aef7387).

zkevm-circuits/src/evm_circuit/op_execution.rs
han0110: Mentioned in #74

zkevm-circuits/src/evm_circuit/op_execution/arithmetic.rs
CPerezz: You're absolutely right. Sorry.

miha-stopar (approved):

PR #27, opened 2021-08-13
[WIP] Implement SUB opcode

gaswhat:
Similar to ADD opcode. Will rebase to main after #15 and #26 is merged.

bus mapping is also not implemented since it's not merged to main yet.


gaswhat:
close this PR and will use scroll-dev account for PR in the future

PR #37, opened 2021-08-20
State circuit: storage operations

miha-stopar:
This PR adds storage operations to the state circuit. There are two things missing though:
- checking account address monotonicity (a problem because of big values for range proof)
- initialization and finalization with Merkle tree


zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -431,6 +474,125 @@ impl<
...
+                q_storage_not_first.clone()
+                    * q_read
+                    * (value_cur - value_previous.clone()), // when reading, the value is the same as at the previous op
+                // Note that this last constraint needs to hold only when address and storage key don't change,
+                // but we don't need to check this as the first operation at new address and
+                // new storage key always has to be write - that means q_read is 1 only when
+                // the address and storage key doesn't change.
+                is_not_padding
+                    * q_storage_not_first
+                    * flag
```
therealyingtong, line 562: Discussed with @miha-stopar that we should also check this when reading.
```suggestion
```

therealyingtong (approved):
LGTM, modulo the constraint change.

PR #44, opened 2021-08-26
fix clippy errors

spartucus:


miha-stopar:
Thanks for the PR! However, not sure if it makes sense to open PRs for Clippy changes. I think it's a good policy if Clippy changes are pushed together with other changes (by each push). I didn't follow that policy until now, but I will try in the future. However, PRs are welcome and we will try to maintain an up-to-date list of issues. Thanks anyway! :+1:

PR #45, opened 2021-08-26
Evm circuit push

miha-stopar:
PUSH opcode first version.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/op_execution/push.rs
+++ b/zkevm-circuits/src/evm_circuit/op_execution/push.rs
@@ -85,20 +85,21 @@ impl<F: FieldExt> OpGadget<F> for PushGadget<F> {
         )];
 
         let success = {
-            let one = Expression::Constant(F::one());
+            let (one, two) = (
+                Expression::Constant(F::one()),
+                Expression::Constant(F::from_u64(2)),
+            );
 
             // interpreter state transition constraints
             let state_transition_constraints = vec![
                 state_next.global_counter.expr()
-                    - (state_curr.global_counter.expr()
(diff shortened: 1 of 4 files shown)


zkevm-circuits/src/evm_circuit/op_execution/push.rs
```diff
@@ -0,0 +1,306 @@
...
+impl<F: FieldExt> PushGadget<F> {
+    fn assign_success(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        op_execution_state: &mut OpExecutionStateInstance,
+        execution_step: &ExecutionStep,
+    ) -> Result<(), Error> {
+        op_execution_state.global_counter += 1;
+        op_execution_state.program_counter +=
```
DreamWuGit, line 247: push1 operation increases program_counter + 2 which is different than other op code

zkevm-circuits/src/evm_circuit/op_execution/push.rs
```diff
@@ -0,0 +1,306 @@
...
+            (execution_step.opcode - 95) as usize;
+        op_execution_state.stack_pointer += 1;
```
DreamWuGit, line 249: should be stack_pointer -= 1

zkevm-circuits/src/evm_circuit/op_execution/push.rs
han0110: Fixed. Thanks!

zkevm-circuits/src/evm_circuit/op_execution/push.rs
han0110: Fixed. Thanks!

PR #54, opened 2021-08-30
keccak256 layout

ChihChengLiang:
### Problem

- Add a skeleton for the keccak chip
- Add some simple tests for later test


rahulbs98:
Adding Base13 to Base9 conversion table

ChihChengLiang:
Close in favor of the split up approach in #103 #104 and #105

PR #63, opened 2021-09-06
Implement POP opcode

scroll-dev:
Implemented the POP opcode

### DIFF
--- a/zkevm-circuits/src/evm_circuit/op_execution/pop.rs
+++ b/zkevm-circuits/src/evm_circuit/op_execution/pop.rs
@@ -131,14 +131,13 @@ impl<F: FieldExt> OpGadget<F> for PopGadget<F> {
         };
 
         let out_of_gas = {
-            let (one, two, three) = (
+            let (one, two) = (
                 Expression::Constant(F::from_u64(1)),
                 Expression::Constant(F::from_u64(2)),
-                Expression::Constant(F::from_u64(3)),
             );
             let (case_selector, gas_available) = &self.out_of_gas;
             let gas_overdemand = op_execution_state_curr.gas_counter.expr()
-                + three.clone()
+                + two.clone()
                 - gas_available.expr();
             Constraint {
                 name: "PopGadget out of gas",
@@ -147,8 +146,7 @@ impl<F: FieldExt> OpGadget<F> for PopGadget<F> {
                     common_polys,
                     vec![
                         (gas_overdemand.clone() - one)
-                            * (gas_overdemand.clone() - two)
-                            * (gas_overdemand - three),
+                            * (gas_overdemand - two)
                     ],
                 ]
                 .concat(),
--- a/zkevm-circuits/src/evm_circuit/op_execution/pop.rs
+++ b/zkevm-circuits/src/evm_circuit/op_execution/pop.rs
@@ -4,6 +4,7 @@ use super::{
     CaseAllocation, CaseConfig, OpExecutionState, CoreStateInstance,
     OpGadget
 };
+use bus_mapping::evm::OpcodeId;
 use halo2::plonk::Error;
 use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Expression};
 use std::convert::TryInto;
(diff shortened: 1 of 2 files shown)


zkevm-circuits/src/evm_circuit/op_execution/pop.rs
```diff
@@ -0,0 +1,305 @@
...
+                        (stack_pointer.clone() - zero)
+                            * (stack_pointer - minus_one),
+                    ],
+                ]
+                .concat(),
+                lookups: vec![],
+            }
+        };
+
+        let out_of_gas = {
```
Brechtpd, line 136: I think this does the out-of-gas check as if POP costs 3 gas instead of 2 gas?

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
```diff
@@ -0,0 +1,305 @@
...
+                polys: [
+                    common_polys.clone(),
+                    op_execution_state_transition_constraints,
+                ]
+                .concat(),
+                lookups: vec![]
+            }
+        };
+
+        let stack_underflow = {
```
Brechtpd, line 115: I think this does the stack underflow check as if POP pops 2 values from the stack instead of 1? (only when the pointer is 1024 should there be underflow).

DreamWuGit: thanks , indeed only need to check one , but stack pointer range from 0..1023 , so should be check only 1023 I think, let me know your point

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
DreamWuGit: it is 2, you can refer to gas table in https://github.com/djrtwo/evm-opcode-gas-costs/blob/master/opcode-gas-costs_EIP-150_revision-1e18248_2017-04-12.csv

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
```diff
@@ -0,0 +1,305 @@
...
+            ),
+        }
+    }
+
+    fn constraints(
+        &self,
+        op_execution_state_curr: &OpExecutionState<F>,
+        op_execution_state_next: &OpExecutionState<F>,
+    ) -> Vec<Constraint<F>> {
+        let pop = Expression::Constant(F::from_u64(80));
```
ChihChengLiang, line 75: nitpick: What do you think using `OpcodeId::POP` here too?

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
```diff
@@ -0,0 +1,305 @@
...
+                        + one.clone()),
+                op_execution_state_next.stack_pointer.expr()
+                    - (op_execution_state_curr.stack_pointer.expr()
+                        + one.clone()),
+                op_execution_state_next.program_counter.expr()
+                    - (op_execution_state_curr.program_counter.expr()
+                        + one.clone()),
+                op_execution_state_next.gas_counter.expr()
+                    - (op_execution_state_curr.gas_counter.expr()
+                        + Expression::Constant(F::from_u64(2))),
```
ChihChengLiang, line 94: nitpick: What do you think we add a constant for `GasQuickStep = 2`

ChihChengLiang (commented):
Added some nitpicks.

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
DreamWuGit: good to use ,thanks

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
DreamWuGit: nice, I think should place on common module such as  gas_counter.rs , which holds all op codes' gas counter there. it is a good modularization design ? what do u think ?

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
DreamWuGit: changed to only 1024 for now, also discussing this point on pr #26

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
Brechtpd: Sorry for being so brief, what I meant was that this seems to still add 3 to the gas counter instead of 2 to check against out-of-gas, so I think this could give a out-of-gas error 1 gas too soon. I believe the code should be like this:
```
let out_of_gas = {
    let (one, two) = (
        Expression::Constant(F::from_u64(1)),
        Expression::Constant(F::from_u64(2)),
    );
    let (case_selector, gas_available) = &self.out_of_gas;
    let gas_overdemand = state_curr.gas_counter.expr() + two.clone()
        - gas_available.expr();
    Constraint {
        name: "Pop out of gas",
        selector: case_selector.expr(),
        polys: [
            common_polys,
            vec![
                (gas_overdemand.clone() - one)
                    * (gas_overdemand.clone() - two),
            ],
        ]
        .concat(),
        lookups: vec![],
    }
};
```

zkevm-circuits/src/evm_circuit/op_execution/pop.rs
DreamWuGit: update to latest GasCost.

PR #65, opened 2021-09-06
Rahulbs98/issue39

rahulbs98:
Adding base conversion tables for keccak


rahulbs98:
Adding conversion tables

PR #67, opened 2021-09-07
Remove warning and fix some typos

3for:
Remove the `broken_intra_doc_links ` warning and fix some typos.

### DIFF
--- a/bus-mapping/src/evm/opcodes/ids.rs
+++ b/bus-mapping/src/evm/opcodes/ids.rs
@@ -477,6 +477,7 @@ impl FromStr for OpcodeId {
             "STATICCALL" => OpcodeId::STATICCALL,
             "SELFDESTRUCT" => OpcodeId::SELFDESTRUCT,
             "CHAINID" => OpcodeId::CHAINID,
+            "BASEFEE" => OpcodeId::BASEFEE,
             _ => return Err(Error::OpcodeParsing),
         })
     }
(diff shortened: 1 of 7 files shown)


bus-mapping/src/evm/opcodes/ids.rs
```diff
@@ -477,6 +477,7 @@ impl FromStr for OpcodeId {
             "STATICCALL" => OpcodeId::STATICCALL,
             "SELFDESTRUCT" => OpcodeId::SELFDESTRUCT,
             "CHAINID" => OpcodeId::CHAINID,
+            "BASEFEE" => OpcodeId::BASEFEE,
```
ChihChengLiang, line 480: @han0110 correct me if I'm wrong. I remember we didn't add `BASEFEE` because we decided not to implement it too early. The more specific reason is that `BASEFEE` makes sense in the L1 validity proof use case but not the zkrollup use case.

ChihChengLiang (approved):
Asked a question for basefee. The rest looks good to me.

bus-mapping/src/evm/opcodes/ids.rs
han0110: Yeah, you are right. And how to support EIP1559 in circuit is not decided yet. (we can implement 2 version for each use case, or single version but turn off EIP1559 in zkrollup use case)

For here I think it's missed in `FromStr`, but it's still included in previous `OpcodeId`

ChihChengLiang: Per offline discussion, adding the missing `FromStr` method can prevent Rust's panic in parsing.

PR #75, opened 2021-09-17
Implement XOR opcode

scroll-dev:
Implement XOR opcode


zkevm-circuits/src/evm_circuit/op_execution/arithmetic/xor.rs
```diff
@@ -0,0 +1,348 @@
...
+                selector: selector.expr(),
+                polys: state_transition_constraints,
+                lookups: [xor_lookups, bus_mapping_lookups].concat(),
+            }
+        };
+
+        let stack_underflow = {
+            let OpExecutionState { stack_pointer, .. } = &state_curr;
+            Constraint {
+                name: "AddGadget stack underflow",
```
DreamWuGit, line 144: change to XorGadget

zkevm-circuits/src/evm_circuit/op_execution/arithmetic/or.rs
```diff
@@ -0,0 +1,343 @@
...
+    selector: Cell<F>,
+    a: Word<F>,
+    b: Word<F>,
+    c: Word<F>,
+}
+
+#[derive(Clone, Debug)]
+pub struct OrGadget<F> {
+    success: OrSuccessAllocation<F>,
+    stack_underflow: Cell<F>, // case selector
```
spartucus, line 23: Does `or` case have stack_underflow situation?

zkevm-circuits/src/evm_circuit/op_execution/arithmetic/bitwise.rs
```diff
@@ -0,0 +1,270 @@
...
+                    &self,
+                    state_curr: &OpExecutionState<F>,
+                    state_next: &OpExecutionState<F>,
+                    name: &'static str,
+                ) -> Constraint<F> {
+                    let mut cb = ConstraintBuilder::default();
+
+                    // for idx in 0..32 {
+                    //     cb.add_fixed_lookup(
+                    //         FixedLookup::[<Bitwise $op>],
```
Brechtpd, line 79: I don't know if this will matter, but if the only difference between these bitwise opcodes is the lookup table, perhaps it's a bit more efficient to just change the `tag` of `FixedLookup` to an expression and set the value to the correct table depending on the opcode being handled?

icemelon: `FixedLookup` also includes range lookup tables. So we cannot use the opcode as the tag unless we separate them from the lookup tables. But I think it's a bit overkill.

Brechtpd: The only difference is the `tag` value for all fixed lookups no? Which is unique per table so I think it should work without splitting them up. The tag would be easy and cheap to calculate as `opcode.as_u8() - OpcodeId::AND.as_u8() + BitwiseAnd as u8`.

If this is overkill or not I don't know, I have no idea what kinds of overheads additional opcode implementations bring, but this seems like a minor change to be able to reuse all constraints between those opcodes.

icemelon:
Blocked on #159 as we need the improvement for lookup performance in the MockProver

barryWhiteHat:
Did you try adding the --release tag as @ChihChengLiang suggested ?

icemelon:
> Did you try adding the --release tag as @ChihChengLiang suggested ?

I couldn't try out the latest fix for lookup table in the MockProver. The lookup check perf improvement [PR](https://github.com/zcash/halo2/pull/389) is merged after the [lookup column](https://github.com/zcash/halo2/pull/354) change. So if we want to include the perf improvement from Halo2, we need to fix the lookup column issue first.

barryWhiteHat:
Right there is another fix that CC found. If he ads the release time to a PR the CLI completes very quickly. @ChihChengLiang am i undrestanding correctly?

ChihChengLiang:
yes, I get 72 sec running "or test" after adding `--release` flag to `cargo test` .

ChihChengLiang:
@icemelon IIUC, this PR is blocked on #159 because we want the MockProver improvement in https://github.com/zcash/halo2/pull/389. However, as I pointed out (and just rechecked again on my device now) in https://github.com/zcash/halo2/issues/398, the https://github.com/zcash/halo2/pull/389 did not improve the performance.

Do you think it is possible to continue without the MockProver improvement?

icemelon:
@ChihChengLiang Yes, I think we can first try it without the MockProver improvement. I have fixed the failed test. Could you approve the CI to run again?

icemelon:
@miha-stopar @ChihChengLiang The CI has passed. Could you take a look again?

ChihChengLiang:
@miha-stopar @barryWhiteHat I'm on vacation but note that the test time takes 1hr33min to complete.

miha-stopar (approved):
LGTM

miha-stopar:
> @miha-stopar @ChihChengLiang The CI has passed. Could you take a look again?

@icemelon Had a look at specs and implementation and both look great to me. Not sure - should I have a look into execution time too or is this ok now (although it's still very long)? CC is on vacations this week so I will try to replace him on this PR.

icemelon:
@miha-stopar I think it is ok for now after CC introduced the `--release` in the test.

miha-stopar:
> @miha-stopar I think it is ok for now after CC introduced the `--release` in the test.

I see, thanks!

han0110 (approved):
LGTM!

PR #87, opened 2021-09-22
Implement MLOAD/MSTORE opcodes

Brechtpd:
Spec: https://github.com/appliedzkp/zkevm-specs/pull/61

Currently the value ranges in the memory size/cost calculations can get quite big because of no early range limitation (except only using the 5 LSBs of the address). Perhaps an earlier check that limits the size a little bit more earlier could save a couple of lookups. But let me know what you think of the current implementation. The testing is also quite limited for now because it's currently not easy to verify all the values against expected ones.

Does not currently implement `MSTORE8`. My idea is to implement this opcode in the same `MemoryGadget` because very few changes are needed. The only challenge is that ideally we can just do the 32 memory lookups to the same single memory write in the busmapping, but for that I need to be able to explicitly assign the `gc` in the busmapping lookup. Because of this complication I think it makes sense to do that in a future PR if this is deemed a good idea. @han0110  
EDIT: Implementation of this can be found here: https://github.com/Brechtpd/zkevm-circuits/pull/1

### DIFF
--- a/bus-mapping/src/evm/gas.rs
+++ b/bus-mapping/src/evm/gas.rs
@@ -16,10 +16,12 @@ impl GasCost {
     pub const SLOW: Self = Self(10);
     /// Constant cost for ext step
     pub const EXT: Self = Self(20);
+    /// Constant cost for every additional word when expanding memory
+    pub const MEMORY: Self = Self(3);
     /// Constant cost for a cold SLOAD
     pub const COLD_SLOAD_COST: Self = Self(2100);
     /// Constant cost for a cold account access
-    pub const COLD_ACCOUNT_ACCESS_COST: Self = Self(100);
+    pub const COLD_ACCOUNT_ACCESS_COST: Self = Self(2600);
     /// Constant cost for a warm storage read
     pub const WARM_STORAGE_READ_COST: Self = Self(100);
 }
--- a/bus-mapping/src/evm/opcodes/mload.rs
+++ b/bus-mapping/src/evm/opcodes/mload.rs
@@ -38,12 +38,21 @@ impl Opcode for Mload {
             StackOp::new(RW::READ, stack_position, stack_value_read),
         );
 
-        //
-        // First mem read -> 32 MemoryOp generated.
-        //
+        // Read the memory
         let mut mem_read_addr: MemoryAddress = stack_value_read.try_into()?;
         let mem_read_value = next_steps[0].memory().read_word(mem_read_addr)?;
 
+        //
+        // First stack write
+        //
+        ctx.push_op(
+            exec_step,
+            StackOp::new(RW::WRITE, stack_position, mem_read_value),
+        );
+
+        //
+        // First mem read -> 32 MemoryOp generated.
+        //
         mem_read_value.inner().iter().for_each(|value_byte| {
             ctx.push_op(
                 exec_step,
@@ -53,14 +62,6 @@ impl Opcode for Mload {
             // Update mem_read_addr to next byte's one
             mem_read_addr += MemoryAddress::from(1);
         });
-
-        //
-        // First stack write
-        //
-        ctx.push_op(
-            exec_step,
-            StackOp::new(RW::WRITE, stack_position, mem_read_value),
-        );
 
         Ok(())
     }
@@ -132,6 +133,16 @@ mod mload_tests {
             ),
         );
 
+        // Add the last Stack write
+        ctx.push_op(
+            &mut step_1,
+            StackOp::new(
+                RW::WRITE,
+                StackAddress::from(1023),
+                EvmWord::from(0x80u8),
+            ),
+        );
+
         // Add the 32 MemoryOp generated from the Memory read at addr 0x40<->0x80 for each byte.
         EvmWord::from(0x80u8)
             .inner()
@@ -144,16 +155,6 @@ mod mload_tests {
                     MemoryOp::new(RW::READ, idx.into(), *byte),
                 );
             });
-
-        // Add the last Stack write
-        ctx.push_op(
-            &mut step_1,
-            StackOp::new(
-                RW::WRITE,
-                StackAddress::from(1023),
-                EvmWord::from(0x80u8),
-            ),
-        );
 
         // Generate Step1 corresponding to PUSH1 40
         let step_2 = ExecutionStep {
--- a/zkevm-circuits/src/evm_circuit/op_execution/byte.rs
+++ b/zkevm-circuits/src/evm_circuit/op_execution/byte.rs
@@ -18,7 +18,8 @@ static STATE_TRANSITION: StateTransition = StateTransition {
     gc_delta: Some(3), // 2 stack pops + 1 stack push
     pc_delta: Some(1),
     sp_delta: Some(1),
-    gas_delta: Some(GasCost::FASTEST.as_usize()),
+    gas_delta: Some(GasCost::FASTEST.as_u64()),
+    next_memory_size: None,
 };
 const NUM_POPPED: usize = 2;
 
@@ -161,7 +162,7 @@ mod test {
         ($execution_steps:expr, $operations:expr, $result:expr) => {{
             let circuit =
                 TestCircuit::<Base>::new($execution_steps, $operations);
-            let prover = MockProver::<Base>::run(10, &circuit, vec![]).unwrap();
+            let prover = MockProver::<Base>::run(11, &circuit, vec![]).unwrap();
             assert_eq!(prover.verify(), $result);
         }};
     }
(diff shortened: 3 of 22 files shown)


bus-mapping/src/evm/gas.rs
```diff
@@ -14,6 +14,8 @@ impl GasCost {
     pub const SLOW: Self = Self(10);
     /// Constant cost for ext step
     pub const EXT: Self = Self(20);
+    /// Constant cost for every additional word when expanding memory
+    pub const MEMORY: Self = Self(3);
```
DreamWuGit, line 18: why not use existing FASTEST field which represents "3" ?

zkevm-circuits/src/evm_circuit/op_execution/context.rs
```diff
@@ -0,0 +1,284 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Context<F> {
+    pub expressions: Vec<Expression<F>>,
+    pub(super) lookups: Vec<Lookup<F>>,
+    pub call_id: Expression<F>,
+    pub stack_offset: i32,
+}
+
+impl<F: FieldExt> Context<F> {
```
DreamWuGit, line 27: does it stand for callContext ?

zkevm-circuits/src/evm_circuit/op_execution/math.rs
```diff
@@ -0,0 +1,215 @@
+use super::super::Cell;
```
DreamWuGit, line 1: math.rs file name seem too generic

bus-mapping/src/evm/gas.rs
Brechtpd: Following the yellow paper here where these are separate constants. Even though they currently have the same value, changing one does not mean the other one needs to be changed as well, which could happen at some point.

zkevm-circuits/src/evm_circuit/op_execution/context.rs
Brechtpd: Sorry for the bad name. The idea is to have something like `ConstraintSystem` or `ProtoBoard` which we can more easily use in the op code gadgets to generate the necessary constraints. This should also help for code reuse/duplication and readability. Currently named it `ConstraintBuilder` which is hopefully a better name. 

Easy example: instead of doing this in all the gadgets to pop a value of the stack

```
Lookup::BusMappingLookup(BusMappingLookup::Stack {
    index_offset: 0,
    value: value.expr(),
    is_write: false,
})
```
we can now just write

```
cb.stack_pop(value.expr());
```

Easier to read, no way to forget to include the constraint that is being built and if something changes to the busmapping API there's only a single place where the code needs to be updated. So just a simple extra layer of abstraction which I think is useful to do for a couple of reasons.

zkevm-circuits/src/evm_circuit/op_execution/math.rs
Brechtpd: Yes agreed, currently changed it to `math_gadgets` which is hopefully at least a bit better. :)

Brechtpd:
Yes, will be picking this up again in the next few days! The main reason I finished other opcodes earlier than this one is because this did the code reuse stuff _and_ needed to do quite complex opcodes. So it seemed better to first introduce the code reuse stuff together with a simple opcode so I (and reviewers) could focus on one thing at a time.

There's also still the open question of where to store the memory size: busmapping vs state. This PR currently stores it into the state.

zkevm-circuits/src/evm_circuit/op_execution/utils/memory_gadgets.rs
```diff
@@ -0,0 +1,231 @@
...
+/// This gas cost is the difference between the next and current memory costs:
+/// `memory_cost = Gmem * memory_size + floor(memory_size * memory_size / 512)`
+#[derive(Clone, Debug)]
+pub(crate) struct MemoryExpansionGadget<F> {
+    address_memory_size: MemorySizeGadget<F>,
+    next_memory_size: MaxGadget<F, MAX_MEMORY_SIZE_IN_BYTES>,
+    curr_quad_memory_cost:
+        ConstantDivisionGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 }>,
+    next_quad_memory_cost:
+        ConstantDivisionGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 }>,
```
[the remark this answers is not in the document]
han0110: I see, used for the OOG case makes sense to me. So the `u128` is then needed.

> I currently made the max quad cost size configurable so that we can indeed change this to 8 for the success case, but allow it to be set bigger to support all possible input values for the out-of-gas case. Does that make sense to you?

Yeah, make sense!

> Yes, for sure. My assumption was that this would be done externally of the op gadgets because practically all op codes require this.

Yes, or naively we can also have a `gas_gadget.rs` to help that.

zkevm-circuits/src/evm_circuit/op_execution/memory.rs
```diff
@@ -0,0 +1,604 @@
...
+
+impl<F: FieldExt> MemorySuccessCase<F> {
+    pub(crate) const CASE_CONFIG: &'static CaseConfig = &CaseConfig {
+        case: Case::Success,
+        num_word: 2  // value + address
+            + MemoryExpansionGadget::<F, MAX_GAS_SIZE_IN_BYTES>::NUM_WORDS
+            + IsEqualGadget::<F>::NUM_WORDS,
+        num_cell: 1  // call_id
+            + MemoryExpansionGadget::<F, MAX_GAS_SIZE_IN_BYTES>::NUM_CELLS
+            + IsEqualGadget::<F>::NUM_CELLS,
```
han0110, line 54:
```suggestion
        num_cell: MemoryExpansionGadget::<F, MAX_GAS_SIZE_IN_BYTES>::NUM_CELLS
            + IsEqualGadget::<F>::NUM_CELLS,
```

Brechtpd: Ah thanks, good catch.

bus-mapping/src/evm/gas.rs
spartucus: It seems like that this `gas.rs` file is not used, or I just missed something?

Brechtpd: I also don't know why that is currently the case. I think there was once a TODO to remove one of them, but I don't know why it is currently like this so I just update both.

bus-mapping/src/evm/opcodes/mload.rs
```diff
@@ -38,12 +38,21 @@ impl Opcode for Mload {
...
-        // First mem read -> 32 MemoryOp generated.
-        //
+        // Read the memory
         let mut mem_read_addr: MemoryAddress = stack_value_read.try_into()?;
         let mem_read_value = next_steps[0].memory().read_word(mem_read_addr)?;
 
+        //
+        // First stack write
+        //
+        ctx.push_op(
```
spartucus, line 48: Why change this order, are these two implementation(32 MemoryOp read first or stack write first) different?

zkevm-circuits/src/evm_circuit/op_execution/byte.rs
```diff
@@ -161,7 +162,7 @@ mod test {
         ($execution_steps:expr, $operations:expr, $result:expr) => {{
             let circuit =
                 TestCircuit::<Base>::new($execution_steps, $operations);
-            let prover = MockProver::<Base>::run(10, &circuit, vec![]).unwrap();
+            let prover = MockProver::<Base>::run(11, &circuit, vec![]).unwrap();
```
spartucus, line 165: What's the reason that `1024` is not enough?

miha-stopar (approved):
Nice! LGTM!

zkevm-circuits/src/evm_circuit/op_execution/byte.rs
miha-stopar: It seems there were too many rows for the number of unity roots that is computed out of param 10.

bus-mapping/src/evm/opcodes/mload.rs
miha-stopar: @Brechtpd?

I will merge it once this is resolved.

Brechtpd: To make the circuit implementation easier to do both MLOAD/MSTORE implementation. `Stack - Memory - Stack` only really makes sense for MLOAD, for MSTORE `Stack - Stack - Memory` is how you would order them normally. Having the order be the same for both (and MSTORE8 in a future PR) makes the circuit easier because otherwise we would have to have extra expressions to set `gc` correctly depending on the opcode.

So a single order is easier, and ``Stack - Stack - Memory` seems nicer to me because the different data locations are grouped together, but that's of course not really important, the only important thing is that they are consistent so I just picked that one.

spartucus: Nice, thanks for explanation:)

miha-stopar: Thanks @Brechtpd!

PR #100, opened 2021-09-30
Implement DUPX opcode

scroll-dev:


ChihChengLiang:
Referencing the DUPX spec here https://github.com/appliedzkp/zkevm-specs/pull/42

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -213,6 +215,22 @@ pub(crate) struct OpExecutionGadget<F> {
     push_gadget: PushGadget<F>,
     lt_gadget: LtGadget<F>,
     pop_gadget: PopGadget<F>,
+    dup1_gadget: DupGadget<F, 0x80>,
+    dup2_gadget: DupGadget<F, 0x81>,
+    dup3_gadget: DupGadget<F, 0x82>,
+    dup4_gadget: DupGadget<F, 0x83>,
```
DreamWuGit, line 221: have used const generic to wrap all dupx code into dup.rs gadget, but at the call side, we also have to distinguish them one by one after have tried vec! & encapsulations.  if you have good idea about this, welcome to point out . thanks

zkevm-circuits/src/evm_circuit/op_execution.rs
```diff
@@ -558,6 +608,103 @@ impl<F: FieldExt> OpExecutionGadget<F> {
                     core_state,
                     execution_step,
                 )?,
+                (_, OpcodeId::DUP1) => self.dup1_gadget.assign(
```
miha-stopar, line 611: I think you could have here only one dup gadget the way we have it for push. You would need to change the match statement a bit to handle dup and push cases. But then a problem appears in dup.rs:

vec![Lookup::BusMappingLookup(BusMappingLookup::Stack {
                    index_offset: (DUPX - usize::from(OpcodeId::DUP1.as_u8()))
                        as i32,
                    value: word.expr(),
                    is_write: false,
                })],

We don't have DUPX here anymore, but talked with @han0110 a bit about this and he suggested you could change index_offset to be an expression and then use opcode.expr() instead of DUPX. Would be that ok for you?

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
```diff
@@ -0,0 +1,462 @@
...
+                }
+                // selectors needs to be 0 or 1
+                dup_constraints.push(
+                    selectors[idx].expr() * (1.expr() - selectors[idx].expr()),
+                );
+            }
+
+            let selectors_sum =
+                selectors.iter().fold(0.expr(), |sum, s| sum + s.expr());
+            dup_constraints.push(selectors_sum - num_duplicated.clone());
```
han0110, line 146: What's the purpose of `selectors` here?

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
```diff
@@ -0,0 +1,462 @@
...
+                lookups: vec![],
+            }
+        };
+
+        let stack_pointer = state_curr.stack_pointer.expr();
+        let diff = 1024.expr() - stack_pointer - num_duplicated;
+
+        // diff's maxium is 1023 when stack_pointer = 0 and position = 1,
+        // the minimum is 0 when stack_pointer + position = 1024 , i.e. stack_point = 1020, DUPX = 4
+        let stack_underflow = {
```
han0110, line 186: Why is maximum `1023`? The underflow case should be `0 <= num_duplicated + stack_pointer - 1025 < 16`, which fits in a `Range16` lookup.

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
```diff
@@ -0,0 +1,462 @@
...
+            Case::Success => {
+                self.assign_success(region, offset, core_state, execution_step)
+            }
+            Case::StackOverflow => {
+                unimplemented!()
+            }
+            Case::OutOfGas => {
+                unimplemented!()
+            }
+            _ => unreachable!(),
```
han0110, line 247: Missing `Case::StackUnderflow`

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
han0110: Doesn't `num_duplicated` already identify that?

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
DreamWuGit: added

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
DreamWuGit: looks like somewhat redundant indeed.  thanks for catching ,will try deletion to see the result

DreamWuGit: what do you think if push gadget really needs selectors ?

zkevm-circuits/src/evm_circuit/op_execution.rs
DreamWuGit: change index_offset to expression make sense here and applied

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
han0110: That serve as boolean selectors for independent lookups, which we can't only use `num_duplicated` to achieve.

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
```diff
@@ -0,0 +1,428 @@
...
+    ];
+
+    fn construct(case_allocations: Vec<CaseAllocation<F>>) -> Self {
+        let [mut success, stack_overflow, stack_underflow, out_of_gas]: [CaseAllocation<F>; 4] =
+            case_allocations.try_into().unwrap();
+        Self {
+            success: DupSuccessAllocation {
+                case_selector: success.selector.clone(),
+                word: success.words.pop().unwrap(),
+                //selectors: success.cells.try_into().unwrap(),
```
miha-stopar, line 82: You can remove this now.

zkevm-circuits/src/evm_circuit/op_execution/dup.rs
```diff
@@ -0,0 +1,428 @@
...
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        core_state: &mut CoreStateInstance,
+        execution_step: &ExecutionStep,
+    ) -> Result<(), Error> {
+        core_state.global_counter += 2; // read + write operation
+        core_state.program_counter += 1;
+        core_state.stack_pointer -= 1;
+        core_state.gas_counter += 3;
```
ChihChengLiang, line 243: Nitpick: these numbers appear here and between line 107~120. We could define constants for them like `byte.rs`.

```
const GC_DELTA: usize = 2;
const PC_DELTA: usize = 1;
const SP_DELTA: usize = 1;
const GAS: GasCost = GasCost::FASTEST;
```

ChihChengLiang:
Great work, thank you @DreamWuGit

PR #102, opened 2021-09-30
First stack op constraint removed

miha-stopar:
The constraint for the first stack operation when (stack) address changes to be write is now removed as this is enforced by EVM circuit.

### DIFF
--- a/zkevm-circuits/src/state_circuit/state.rs
+++ b/zkevm-circuits/src/state_circuit/state.rs
@@ -26,10 +26,10 @@ Example state table:
 |    2     |    1    |       2        |  12   |   0  |    0    |             |            |
 |    2     |         |                |       |      |    1    |             |            |   // padding
 |    2     |         |                |       |      |    1    |             |            |   // padding
-|    1     |    0    |       3        |  4    |   1  |    0    |             |            |   // first stack op at the new address has to be write
+|    1     |    0    |       3        |  4    |   1  |    0    |             |            |
 |    3     |    0    |       17       |  32   |   1  |    0    |             |            |
 |    3     |    0    |       89       |  32   |   0  |    0    |             |            |
-|    3     |    1    |       48       |  32   |   1  |    0    |             |            |   // first stack op at the new address has to be write
+|    3     |    1    |       48       |  32   |   1  |    0    |             |            |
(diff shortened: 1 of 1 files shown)


zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -1486,8 +1472,7 @@ mod tests {
     #[test]
     fn first_write() {
         let stack_op_0 = StackOp::new(
```
han0110, line 1474: Why not remove `stack_op_0` entirely? Or it's for dummy usage?

han0110 (commented):
LGTM!

zkevm-circuits/src/state_circuit/state.rs
miha-stopar: Thanks! Yes, not sure if it makes much sense to leave it there, but the idea was to have one test which has no memory ops, but some stack and storage ops. There is also, for example, a test "storage" which has no memory and stack ops. These tests (first_write and storage) actually test other things, but I thought it's good to cover cases where there are empty ops of some type(s). Do you think it's better to remove it?

han0110: Ah okay, make sense to me now. But it reminds me that we didn't have the failure case that the memory is read when memory address changes (so as a first writes). It would be good to include that in this test, right? (we could add it further PR for sure)

han0110: Ah okay, make sense to me now. But it reminds me that we didn't have the failure case that the memory is read when memory address changes (so as a first writes). It would be good to include that in this test, right? (we could add it further PR for sure)

han0110 (approved):

zkevm-circuits/src/state_circuit/state.rs
miha-stopar: Yes, I was thinking about that some time ago. There is no such test because "init" memory rows are inserted automatically and I would need to pass the information to the circuit about not automatically inserting "init" rows to enable such test. Do you think we should do this?

han0110: Ah I see, maybe just keep it as is now.

PR #106, opened 2021-09-30
Implement BYTE opcode

Brechtpd:
Also contains a couple of helper classes and utilities to more easily reuse frequently used code which should also improve readability. Currently these are only used for the BYTE opcode implementation because I'm not sure if people will like this, so let me know what you think.


zkevm-circuits/src/evm_circuit/op_execution/byte.rs
```diff
@@ -0,0 +1,350 @@
...
+        cb.stack_push(self.result.expr());
+
+        // State transitions
+        #[allow(clippy::needless_update)]
+        utils::StateTransitions {
+            gc_delta: Some(GC_DELTA.expr()),
+            sp_delta: Some(SP_DELTA.expr()),
+            pc_delta: Some(PC_DELTA.expr()),
+            gas_delta: Some(GAS.expr()),
+            ..Default::default()
```
ChihChengLiang, line 116: nitpick: My clippy seems happy with the suggested update.
```suggestion
        utils::StateTransitions {
            gc_delta: Some(GC_DELTA.expr()),
            sp_delta: Some(SP_DELTA.expr()),
            pc_delta: Some(PC_DELTA.expr()),
            gas_delta: Some(GAS.expr()),
```

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
```diff
@@ -0,0 +1,85 @@
+use super::super::CaseAllocation;
+use super::super::Cell;
+use crate::util::Expr;
+use halo2::plonk::Error;
+use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Expression};
+
+use super::constraint_builder::ConstraintBuilder;
+
+// a == 0
```
ChihChengLiang, line 9: I was confused at first for not being aware of its returning value. Suggest adding the doc to emphasize the return value part.
Notes to other reviewers: This gadget is the equivalent of circomlib's `IsZero` templates https://github.com/iden3/circomlib/blob/master/circuits/comparators.circom#L23
```suggestion
// a == 0
// The gadget returns 1 if the input value `a` is 0, and it returns 0 otherwise. 
```

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
```diff
@@ -0,0 +1,85 @@
...
+        offset: usize,
+        value: F,
+    ) -> Result<bool, Error> {
+        let inverse = value.invert().unwrap_or(F::zero());
+        self.inverse.assign(region, offset, Some(inverse))?;
+        Ok(value.is_zero())
+    }
+}
+
+// a == b
```
ChihChengLiang, line 51: Suggesting doc to emphasize the returning value.
Notes to other reviewers: This gadget is the equivalent of circomlib's IsEqual templates https://github.com/iden3/circomlib/blob/master/circuits/comparators.circom#L36-L45
```suggestion
// a == b
// The gadget returns 1 if `a==b` and returns 0 if `a!=b`. 
```

Brechtpd:
> Really nice work! I like the utilities too! I would only perhaps slightly modify the tests to have the variables which are then passed into ExecutionSteps and Operations (because now changing the value to be tested requires changes in multiple places).

I kind of neglected the tests because I hoped to be able to use the tracer test tool for that, but yeah doesn't seem that will be very soon so should have put in a bit more time there regardless. I have updated the test code.

> Would it possible to set some max_degree on ConstraintBuilder to panic when the function makes the degree too high?

I have added a check on the set length, because I think that was kind of the main idea of the check. But I wonder how useful that check really is. The expressions that are passed in could be a higher degree already so perhaps better to add a check on the degree of the final expression that is built? Although even that may not really work as expected always because a high degree expression could be passed in as a single element set, which I guess would be a valid use of the function as it could not be done more efficiently with a lookup directly? What do you think?

han0110:
`<= max_degree` makes sense to me :)

> I'm assuming we should also guard the degree of expressions used in lookups in the same way.

For sure, cause the lookup currently is composed in a strange way, which might have a high degree of input. I think we probably want a simpler lookup in the future... For example a selector with some advice columns, and when one wants to enable the lookup, it constrain the input in these columns, and also the selector to be enabled. It seems a more reasonable way to construct lookup (we make use of it on much more rows), and then we don't need to worry about lookup's degree anymore.

Regarding to lookup's degree, it also has a [`required_degree`](https://github.com/zcash/halo2/blob/4283713ec7/src/plonk/lookup.rs#L65-L70) depends on input and table. So if to check lookup's degree, it would be like `assert!(expr.degree() + 3 <= self.max_degree)` if we assume table has max degree 1 (we might gain extra 1 degree if we disable the zk feature).

BTW, IIUC all of gate or arguments' required degree would be [subtracted by one](https://github.com/zcash/halo2/blob/4283713ec7/src/poly/domain.rs#L40) due to [`divide_by_vanishing_poly`](https://github.com/zcash/halo2/blob/4283713ec76051eaf21a06d0279fa7d3497cafb6/src/poly/domain.rs#L383), so our `DEFAULT_MAX_DEGREE` could be `2**n + 1` like `9`.

Brechtpd:
Alright, thanks for the info and help! I think I have added these checks correctly now.

zkevm-circuits/src/evm_circuit/op_execution/byte.rs
```diff
@@ -0,0 +1,292 @@
...
+            check_byte_gadget(
+                test_value.clone(),
+                idx.into(),
+                if idx < 32 {
+                    (idx + 1).into()
+                } else {
+                    0u64.into()
+                },
+            );
+        }*/
```
ChihChengLiang, line 290: Instead of leaving dead codes here, can we create an issue and put this code snippet there?

Brechtpd: Marked the exhaustive test as ignored so it doesn't run by default, but all ignored tests can be run when doing `cargo test -- --ignored`. Does that sound good to you? No more dead code and the test code is still nicely available next to the other code, and all future slow tests can do the same so there's just one command that can be run to all test these as well.

PR #107, opened 2021-10-01
Extract common error constraints into functions

scroll-dev:


Brechtpd:
I did something similar in #106 to reuse more code, please have a look and let me know what you prefer.

PR #108, opened 2021-10-01
Implement opcode swap1~swap16

scroll-dev:


zkevm-circuits/src/evm_circuit/op_execution/swap.rs
```diff
@@ -0,0 +1,483 @@
+use super::super::{
```
DreamWuGit, line 1: refer to spec https://github.com/appliedzkp/zkevm-specs/pull/48

zkevm-circuits/src/evm_circuit/op_execution/swap.rs
```diff
@@ -0,0 +1,483 @@
...
+struct SwapSuccessAllocation<F> {
+    case_selector: Cell<F>,
+    words: [Word<F>; 2],
+}
+
+#[derive(Clone, Debug)]
+pub struct SwapGadget<F> {
+    success: SwapSuccessAllocation<F>,
+    stack_underflow: Cell<F>,
+    // case selector
```
ChihChengLiang, line 21: nitpick: Do you intend to comment like the suggested?
```suggestion
    stack_underflow: Cell<F>,  // case selector
```

DreamWuGit: OK with it & updated

zkevm-circuits/src/evm_circuit/op_execution/swap.rs
```diff
@@ -0,0 +1,482 @@
...
+        )];
+
+        let num_swaped = opcode.expr() - OpcodeId::SWAP1.expr() + 1.expr();
+        let success = {
+            let SwapSuccessAllocation {
+                case_selector,
+                words,
+            } = &self.success;
+
+            let state_transition_constraints = vec![
```
han0110, line 103: We should also constraint `stack_pointer` to be equal to next step. Otherwise the prover could choose any `stack_pointer` it wants after `SWAP*`.

zkevm-circuits/src/evm_circuit/op_execution/swap.rs
```diff
@@ -0,0 +1,482 @@
...
+                name: "SwapGadget success",
+                selector: case_selector.expr(),
+                polys: state_transition_constraints,
+                lookups: bus_mapping_lookups,
+            }
+        };
+
+        // stack_underflow condition: 0 <= num_swaped +  stack_pointer - 1025 < 16
+        let stack_pointer = state_curr.stack_pointer.expr();
+        let diff = num_swaped + stack_pointer - 1025.expr();
```
han0110, line 155: The stack underflow case for `SWAP*` is slightly different than `DUP*`, we need `stack.len() >= num_swaped + 1`, for example the `SWAP1` swaps `stack[-1]` and `stack[-2]`.

So the range lookup should be `0 <= num_swaped +  stack_pointer - 1024 < 17`, notice that the max value is `16` dut to `SWAP16`. So we might probably need a `Range17` tag for this.

PR #111, opened 2021-10-01
Labeler and release draft

ChihChengLiang:
### Problem

We want to cut releases for the repo so that people working on the testing have some versions to track.

It would be great to inform in the release draft that what features or changes we added in that release

### Solution

1. We use the auto labeler to label PRs that are related to "zkevm-circuit" (core-API), "opcodes", "busmapping", or "keccak".
2. We use the release drafter to automatically generate reports that sort PRs under each label.

Example release would look like this https://github.com/thehubbleproject/hubble-contracts/releases/tag/v0.10.0


CPerezz (approved):
Looks awesome!! Thanks for that CC!!

LGTM

ChihChengLiang:
Thanks for the review @CPerezz!

PR #112, opened 2021-10-01
remove beta lint

ChihChengLiang:
### Problem

The CI has a beta lint and a stable lint to format the rust code.
The original intention is to enforce the stable lint for the codebase and stay informed of changes in beta.
But the two types of lint could create warning messages and are confusing.

### Solution

Keep the stable lint and remove the beta lint


CPerezz (commented):
As already expressed in the call. I'm happy removing these and just keeping `stable`.

LGTM. Let's see what others say! :)

CPerezz (approved):
LGTM. Ready to merge!

PR #113, opened 2021-10-05
add missing src

ChihChengLiang:
### Problem

Due to the wrong path regex, the labeler can't correctly label `zkevm-circuits` and `opcode`

### Solution

Fix the path regex


CPerezz (approved):
LGTM!

Good catch @ChihChengLiang :smile:

PR #120, opened 2021-10-15
Add EQ support in LtGadget

Brechtpd:
Good/bad idea? I would think this is more efficient than having a separate gadget because this only adds a few expressions to `LtGadget`. Also not 100% sure this works correctly, so would like to know from the original implementors if there may be some unexpected problems with doing EQ like this.

Tried to keep the diff minimal to show only the necessary changes for EQ support, so did not do anything else (like use `ConstraintBuilder`).

Also fixed a small issue with the tests where the trace was incorrectly built.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
+++ b/zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
@@ -1,379 +1,205 @@
 use super::super::{
-    BusMappingLookup, Case, Cell, Constraint, CoreStateInstance, ExecutionStep,
-    FixedLookup, Lookup, Word,
+    Case, Cell, Constraint, CoreStateInstance, ExecutionStep, Word,
 };
+use super::utils::common_cases::{OutOfGasCase, StackUnderflowCase};
+use super::utils::constraint_builder::ConstraintBuilder;
+use super::utils::math_gadgets::{ComparisonGadget, IsEqualGadget};
+use super::utils::{self, from_bytes, select, StateTransitions};
 use super::{CaseAllocation, CaseConfig, OpExecutionState, OpGadget};
+use crate::impl_op_gadget;
 use crate::util::{Expr, ToWord};
 use bus_mapping::evm::{GasCost, OpcodeId};
 use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Error};
-use std::{array, convert::TryInto};
+use std::convert::TryInto;
+
+const GC_DELTA: usize = 3;
+const PC_DELTA: usize = 1;
+const SP_DELTA: usize = 1;
+const GAS: GasCost = GasCost::FASTEST;
+const NUM_POPPED: usize = 2;
+
+impl_op_gadget!(
+    [LT, GT, EQ]
+    ComparatorGadget {
+        ComparatorSuccessCase(),
+        StackUnderflowCase(NUM_POPPED),
+        OutOfGasCase(GAS.as_usize()),
+    }
+);
 
 #[derive(Clone, Debug)]
-struct LtSuccessAllocation<F> {
-    selector: Cell<F>,
-    // A 0/1 value. The gadget always checks if `a < b`. When the swap value is 0,
-    // the gadget performs opcode LT and treats top 2 stack values as `a, b`.
-    // Otherwise, the gadget performs opcode GT, and the top 2 stack values are
-    // interpreted as `b, a`.
-    swap: Cell<F>,
+struct ComparatorSuccessCase<F> {
+    case_selector: Cell<F>,
     a: Word<F>,
     b: Word<F>,
-    // A 0/1 value of opcode output.
     result: Word<F>,
-    // Satisfy `a + c = b` and allow overflow at 256 bit
-    c: [Cell<F>; 32],
-    // The carry of the addition check `a + c = b` in 128 bit chunks.
-    carry: Cell<F>,
-    // The sum of all limbs in `c`.
-    sumc: Cell<F>,
-    // The inverse of `sumc`.
-    sumc_inv: Cell<F>,
+    comparison_lo: ComparisonGadget<F, 16>,
+    comparison_hi: ComparisonGadget<F, 16>,
+    is_eq_op: IsEqualGadget<F>,
+    swap: IsEqualGadget<F>,
 }
 
-#[derive(Clone, Debug)]
-pub struct LtGadget<F> {
-    success: LtSuccessAllocation<F>,
-    stack_underflow: Cell<F>,
-    out_of_gas: (Cell<F>, Cell<F>),
-}
+impl<F: FieldExt> ComparatorSuccessCase<F> {
+    pub(crate) const CASE_CONFIG: &'static CaseConfig = &CaseConfig {
+        case: Case::Success,
+        num_word: 3,
+        num_cell: ComparisonGadget::<F, 16>::NUM_CELLS * 2
+            + IsEqualGadget::<F>::NUM_CELLS * 2,
+        will_halt: false,
+    };
 
-impl<F: FieldExt> OpGadget<F> for LtGadget<F> {
-    const RESPONSIBLE_OPCODES: &'static [OpcodeId] =
-        &[OpcodeId::LT, OpcodeId::GT];
-
-    const CASE_CONFIGS: &'static [CaseConfig] = &[
-        CaseConfig {
-            case: Case::Success,
-            num_word: 3,
-            num_cell: 36, // 32 c + swap + carry + sumc + sumc_inv
-            will_halt: false,
-        },
-        CaseConfig {
-            case: Case::StackUnderflow,
-            num_word: 0,
-            num_cell: 0,
-            will_halt: true,
-        },
-        CaseConfig {
-            case: Case::OutOfGas,
-            num_word: 0,
-            num_cell: 0,
-            will_halt: true,
-        },
-    ];
-
-    fn construct(case_allocations: Vec<CaseAllocation<F>>) -> Self {
-        let [mut success, stack_underflow, out_of_gas]: [CaseAllocation<F>; 3] =
-            case_allocations.try_into().unwrap();
+    pub(crate) fn construct(alloc: &mut CaseAllocation<F>) -> Self {
         Self {
-            success: LtSuccessAllocation {
-                selector: success.selector,
-                swap: success.cells.pop().unwrap(),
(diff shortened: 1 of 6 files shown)


zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
```diff
@@ -1,379 +1,196 @@
...
+        // - `eq` when EQ
+        let result = select::expr(is_eq_op, eq, lt);
+
+        // Pop a and b from the stack, push the result on the stack.
+        // When swap is enabled we swap stack places between a and b.
+        // We can push result here directly because
+        // it only uses the LSB of a word.
+        cb.stack_pop(select::expr(swap.clone(), self.b.expr(), self.a.expr()));
+        cb.stack_pop(select::expr(swap, self.a.expr(), self.b.expr()));
+        cb.stack_push(result);
```
icemelon, line 117: So this can be correct only when the value is less than 256, right?

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
```diff
@@ -83,3 +85,146 @@ impl<F: FieldExt> IsEqualGadget<F> {
...
+
+    pub(crate) fn constraints(
+        &self,
+        cb: &mut ConstraintBuilder<F>,
+        lhs: Expression<F>,
+        rhs: Expression<F>,
+    ) -> Expression<F> {
+        let diff = utils::from_bytes::expr(self.diff.to_vec());
+        // The equation we require to hold: `lhs == rhs + diff - (lt * range)`.
+        cb.require_equal(lhs, rhs + diff - (self.lt.expr() * self.range));
```
icemelon, line 127: Though they are the same, I feel it looks more clear changing it this way.

```suggestion
        cb.require_equal(lhs - rhs, diff - (self.lt.expr() * self.range));
```

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
```diff
@@ -83,3 +85,146 @@ impl<F: FieldExt> IsEqualGadget<F> {
...
+        let lt = lhs < rhs;
+        self.lt.assign(
+            region,
+            offset,
+            Some(F::from_u64(if lt { 1 } else { 0 })),
+        )?;
+
+        // Set the bytes of diff
+        let diff = (lhs - rhs) + (if lt { self.range } else { F::zero() });
+        let diff_bytes = diff.to_bytes();
```
icemelon, line 156: One question here is that I remember previously @han0110 and @miha-stopar thinks the intermediate results should still be provided in the witness instead of computing it in the assignment. Should we still follow this principle?

zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
Brechtpd: Yep, see Han's comment here: https://github.com/appliedzkp/zkevm-circuits/pull/120#discussion_r738913187

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
Brechtpd: Agree!

zkevm-circuits/src/evm_circuit/op_execution/utils/math_gadgets.rs
Brechtpd: Hmm I don't believe that would make things better. All these intermediate results are implementation dependent, while the actual input witness for an opcode would best be as general as possible to allow the circuits to use that data as needed. Let's say we change the implementation to work on 2 bytes instead of 1 after we add a `range65536` lookup table to optimize things, I would expect this to not impact the witness format generated by the busmapping.

han0110: The reason we don't feed the `geth` trace into circuit directly is because we need to pre-process to get some metadata (like counter, result in the future, etc...), then we thought maybe we could just generate everything in pre-processing stage including "intermediate witness".

However, after more opcodes are implemented, such "intermediate witness" seems hard to generated from crate `bus-mappping` since their type and amount could be very different, and it becomes even trivial for those opcodes with complicated logics. So I feel that @Brechtpd's idea makes more sense to me now.

Maybe we should just let `bus-mappping` provide the `Operation` which contains all necessary kinds of read/write (which is just current API), and let circuit generate "intermediate witness" itself for flexibility. What do you think @CPerezz, @ed255?

zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
```diff
@@ -1,379 +1,205 @@
...
-                constraint
-            })
-            .collect()
-    }
+        // `a[16..32] <= b[16..32]`
+        let (lt_hi, eq_hi) = self.comparison_hi.constraints(
+            &mut cb,
+            from_bytes::expr(self.a.cells[16..32].to_vec()),
+            from_bytes::expr(self.b.cells[16..32].to_vec()),
+        );
```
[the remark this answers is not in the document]
miha-stopar: > That's an interesting way of looking at it! I think this would definitely work if we didn't have to do the comparison in two parts. But because of the two separate parts I don't think there's a 1:1 relation to the `swap` value for both parts (the first part can be smaller while the second part can be bigger, so the equations may need to be made valid in different ways). I do think there's probably a way to use the `swap` value directly in just the `hi` part because some 1:1 mapping could probably be found there... But I would guess (and very much a guess!) this probably wouldn't make the efficiency of the circuit better (in the best case save a single cell with maybe some more complex expressions in return because `swap` isn't just `1` when GT) and the current method allows the reuse of a general gadget that can be used everywhere which is nice. Am I still missing something or would you agree?

I like the implementation and abstraction, but I am wondering whether some constraint is missing to somehow connect swap and lt? For example, if you change

`check(OpcodeId::LT, hi_lo.clone(), lo_hi.clone(), 0x00u64.into());
`
to
`check(OpcodeId::GT, hi_lo.clone(), lo_hi.clone(), 0x00u64.into());
`

and if you change assign function (which prover can do) from
`let swap = step.opcode == OpcodeId::GT;
`
to
`let swap = false;`

the test will pass.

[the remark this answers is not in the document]
Brechtpd: That's because currently the compressed value of both `hi_lo` and `lo_hi` is the same because `r` is set to `1`: https://github.com/appliedzkp/zkevm-circuits/blob/dfff7bd1d46098e7c3c1d51929cb904d3ac0b662/zkevm-circuits/src/evm_circuit.rs#L981-L982

So the the compression simply does a sum of all byte values, which for `hi_lo` and `lo_hi` is the same. If you use values where this isn't the case that will fail as expected. Also note that the `swap` value that was changed in your example only impacts the swap of the stack values, not the constraint that checks what the opcode actually is (the witness for that is calculated in the `IsEqualGadget`). So definitely a bit confusing because of these things together why that worked :).

I think all tests currently depend on `r == 1` for generating the compressed values for the test values because `r` is not directly made available as far as I know.

zkevm-circuits/src/evm_circuit/op_execution/utils.rs
```diff
@@ -10,6 +10,11 @@ pub(crate) mod common_cases;
 pub(crate) mod constraint_builder;
 pub(crate) mod math_gadgets;
 
+/// The maximum number of bytes that a field element
+/// can be broken down into without causing the value it
+/// represents to overflow a single field element.
+pub const MAX_BYTES_FIELD: usize = 31;
```
icemelon, line 16: It'll be better if we move this const to `src/evm_circuit/param.rs`.

Brechtpd: Done! There's some other constants I introduced in my other PRs that I will also move there after this PR is merged.

zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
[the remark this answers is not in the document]
miha-stopar: Thanks! Didn't know about `r = 1`. I changed it to 2, but then all tests fail (unchanged) - do I need to change something else to enable different `r`?

Yes, I am aware of `swap` affecting only the stack order - that's actually where my concern came from, because it seemed to me that we can for example provide proof for a < b, but then we use the same proof for a > b by switching a and b in the proof, because the information about being LT or GT is on one hand in the swap and on the other hand in opcode/result, but somehow it seems to me that this information is not checked to be the same. Probably I am just missing something. I am sorry for taking your time, I am probably just a bit too paranoid about forgetting some constraints.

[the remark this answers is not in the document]
Brechtpd: > Thanks! Didn't know about r = 1. I changed it to 2, but then all tests fail (unchanged) - do I need to change something else to enable different r?

Modified the `compress` helper function used in the comparator tests to use a hardcoded `r` value. If you change it to match the one set on the circuit it should work. Currently hard coded so will need to be improved in the future so we can get this value from withing `assign` somehow. A lot of tests currently explicitly sum the bytes manually for the compression so definitely some general test improvements necessary. I left the test cases intact in my other refactor PR because those PRs already change enough, but something that needs to be done in a future PR.

> Yes, I am aware of `swap` affecting only the stack order - that's actually where my concern came from, because it seemed to me that we can for example provide proof for a < b, but then we use the same proof for a > b by switching a and b in the proof, because the information about being LT or GT is on one hand in the swap and on the other hand in opcode/result, but somehow it seems to me that this information is not checked to be the same. Probably I am just missing something. I am sorry for taking your time, I am probably just a bit too paranoid about forgetting some constraints.

The direct connection is there for the constraints right? In the witness generation there were indeed two independent variables, but that's just for setting the witnesses so it doesn't matter. But I changed the witness assignment to more closely follow how the constraints work, I hope that makes things more clear how everything is connected with the constraints.

icemelon (approved):
LGTM

zkevm-circuits/src/evm_circuit/op_execution/comparator.rs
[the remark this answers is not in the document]
miha-stopar: Thanks for your patience, Brecht! Sorry for troubling you!

miha-stopar (approved):

PR #123, opened 2021-10-20
add running sum and block count check for rho

### LINKED ISSUE #103: [Keccak] Implement Rho and Pi circuit
Pi is trivial, should be resolved in 1 or 2 rows.

Rho needs these checks

- Running sum by slices
- block count checks

Some initial work is in #54

ChihChengLiang:
### Problem

Fix #103, now only rho gate left

### Solution

Fix the assign_region for lane config, chunk config, special chunks config, and block_count check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 Cargo.lock
+*.png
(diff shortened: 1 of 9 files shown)


ChihChengLiang:
close in favor of #156

PR #131, opened 2021-10-22
Fix the test error for geth-utils on MacOS

scroll-dev:
See the issue in #130

### DIFF
--- a/geth-utils/build.rs
+++ b/geth-utils/build.rs
@@ -8,7 +8,11 @@ fn main() {
     gobuild::Build::new().file("./lib/lib.go").compile(lib_name);
 
     // Files the lib depends on that should recompile the lib
-    let dep_files = vec!["./gethutil/trace.go"];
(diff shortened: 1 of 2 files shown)


scroll-dev:
cc @DreamWuGit @Brechtpd

geth-utils/build.rs
```diff
@@ -8,7 +8,11 @@ fn main() {
     gobuild::Build::new().file("./lib/lib.go").compile(lib_name);
 
     // Files the lib depends on that should recompile the lib
-    let dep_files = vec!["./gethutil/trace.go"];
+    let dep_files = vec![
+        "./gethutil/asm.go",
+        "./gethutil/trace.go",
+        "./gethutil/util.go",
```
Brechtpd, line 14: The lib currently doesn't depend on `asm.go` and `util.go`, but you're right that it's probably a good idea to add these here anyway because that could change at any time.

PR #147, opened 2021-10-26
Refactor all opcodes to use the constraint builder + other utils

Brechtpd:
Updated all opcodes implementations to use all the utilities we currently have available that did not already do so. Unfortunately this makes for a large diff but better earlier than later I think. The comparator gadget was already updated in #120 so did not update it again here.

All implementations should be identical to the previous version. If you find a change it's very likely I made a mistake in the conversion. I did do a small change in `push.rs` lines 75..91 because I think there was an issue there (did not check the MSB selector) but maybe I misunderstood something.

### DIFF
--- a/bus-mapping/src/evm.rs
+++ b/bus-mapping/src/evm.rs
@@ -268,7 +268,7 @@ pub struct GasCost(u64);
 impl GasCost {
     /// Constant cost for free step
     pub const ZERO: Self = Self(0);
-    /// Constant cost for jumpdest step, only it takes one gas
+    /// Constant cost for jumpdest step, only takes one gas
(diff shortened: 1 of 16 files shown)


miha-stopar (approved):
LGTM!

miha-stopar:
> LGTM!

Just saw there is a conflict with math_gadgets.rs, can you resolve it @Brechtpd? Will merge right after that.

Brechtpd:
> > LGTM!
> 
> Just saw there is a conflict with math_gadgets.rs, can you resolve it @Brechtpd? Will merge right after that.

Done! @miha-stopar

PR #157, opened 2021-11-04
udpate toolchain to enable the array_map feature

ChihChengLiang:
### Problem

#156 is blocked by using a feature array_map, which is considered unstable in the current toolchain

### Solution

We upgrade the toolchain and fix related issues


han0110 (approved):
LGTM!

PR #159, opened 2021-11-08
Attempt to upgrade halo2 version

ChihChengLiang:
### Problem

We would like to know how much change it affects the current codebase if we are to use the latest halo2 version.

### Current status

- It's not compiled yet.

### DIFF
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -659,6 +659,7 @@ impl<F: FieldExt> EvmCircuit<F> {
                     .map(|(expr, column)| {
                         (
                             expr.clone(),
+                            // I think we have some problem here
                             meta.query_advice(*column, Rotation::cur()),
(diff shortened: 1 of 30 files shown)


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -665,6 +659,7 @@ impl<F: FieldExt> EvmCircuit<F> {
                     .map(|(expr, column)| {
                         (
                             expr.clone(),
+                            // I think we have some problem here
                             meta.query_advice(*column, Rotation::cur()),
```
ChihChengLiang, line 663: In this line, we are doing a lookup to an `advice` column. The new API seems not to allow us to do that anymore. Would like to hear feedback from @han0110 and @therealyingtong.

spartucus:
Definitely necessary:)

ChihChengLiang:
close in favor of the work in #161

PR #161, opened 2021-11-11
[WIP]Kzg Prover

NoCtrlZ:
# Abstract
I am replacing with Kate commitment for prover container.
This halo2 using `lookup2` temporary method as experiment defined as following.
https://github.com/appliedzkp/halo2/pull/6/files

## Done
- Update halo2 version and compile all
- Success all test except `evm_word` test
- Setup docker
- Make prover work with Kate polynomial commitment

## Problem
- `evm_word` test failed to pass the test
- following state circuit lookup2 doesn't work
```rust
// Memory address is in the allowed range.
        meta.lookup2(|meta| {
            let q_memory =
                q_memory_first_norm(meta) + q_memory_not_first_norm(meta);
            let memory_address_table_zero =
                meta.query_fixed(memory_address_table_zero, Rotation::cur());

            vec![(q_memory * address_cur, memory_address_table_zero)]
        });
```
## Assumption
- `lookup2` has problem
- The way to use lookup table was changed

## Working on
Rewrite `monotone.rs` test with latest `lookup` because `monotone.rs` circuit only has `Column<Fixed>` so we can rewrite it.


han0110:
It seem we have mixed the selector in the expression, which could not be optmized by the new feature `compress_selectors`. It's fixed after changing the [`q_step`](https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit.rs#L342) to `complex_selector`.

NoCtrlZ:
> It seem we have mixed the selector in the expression, which could not be optmized by the new feature compress_selectors. It's fixed after changing the q_step to complex_selector.

Thank you for the comment!
I am going to work on that.

NoCtrlZ:
I would like to integrate this PR to following PR.
#192

PR #170, opened 2021-11-14
Bytecode unroller

Brechtpd:
Spec: https://github.com/appliedzkp/zkevm-specs/pull/74

Disabled the `bytecode_invalid_index` for now because with the halo update it now gives an internal error on a circuit that should fail: `panicked at 'internal error: entered unreachable code', /home/brecht/.cargo/git/checkouts/halo2-b8251ca10f182e2b/b78c39c/src/dev/util.rs:48:42`. With the previous halo this test worked fine so not sure what's going on yet.

The constraint builder could be modified a bit more to make the circuit writing for these standalone circuits even a bit nicer, but Han's PR already does a couple of good changes so will await that one instead before doing so.

### DIFF
--- a/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
+++ b/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
@@ -1,0 +1,867 @@
+use crate::{
+    evm_circuit::util::{
+        and, constraint_builder::BaseConstraintBuilder, not, or, select, RandomLinearCombination,
+    },
+    gadget::{
+        evm_word::encode,
+        is_zero::{IsZeroChip, IsZeroConfig, IsZeroInstruction},
+    },
+    util::Expr,
+};
+use bus_mapping::evm::OpcodeId;
+use halo2::{
+    circuit::{Layouter, Region},
+    plonk::{Advice, Column, ConstraintSystem, Error, Fixed, Selector, VirtualCells},
+    poly::Rotation,
+};
+use keccak256::plain::Keccak;
+use pairing::arithmetic::FieldExt;
+use std::{convert::TryInto, vec};
+
+use super::param::{KECCAK_WIDTH, PUSH_TABLE_WIDTH};
+
+/// Public data for the bytecode
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct BytecodeRow<F: FieldExt> {
+    hash: F,
+    index: F,
+    is_code: F,
+    byte: F,
+}
+
+/// Unrolled bytecode
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct UnrolledBytecode<F: FieldExt> {
+    bytes: Vec<u8>,
+    rows: Vec<BytecodeRow<F>>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Config<F> {
+    r: F,
+    minimum_rows: usize,
+    q_enable: Selector,
+    q_first: Column<Fixed>,
+    q_last: Selector,
+    hash: Column<Advice>,
+    index: Column<Advice>,
+    is_code: Column<Advice>,
+    byte: Column<Advice>,
+    push_rindex: Column<Advice>,
+    hash_rlc: Column<Advice>,
+    hash_length: Column<Advice>,
+    byte_push_size: Column<Advice>,
+    is_final: Column<Advice>,
+    padding: Column<Advice>,
+    push_rindex_inv: Column<Advice>,
+    push_rindex_is_zero: IsZeroConfig<F>,
+    push_table: [Column<Fixed>; PUSH_TABLE_WIDTH],
+    keccak_table: [Column<Advice>; KECCAK_WIDTH],
+}
+
+impl<F: FieldExt> Config<F> {
+    pub(crate) fn configure(meta: &mut ConstraintSystem<F>, r: F) -> Self {
+        let q_enable = meta.complex_selector();
+        let q_first = meta.fixed_column();
+        let q_last = meta.selector();
+        let hash = meta.advice_column();
+        let index = meta.advice_column();
+        let is_code = meta.advice_column();
+        let byte = meta.advice_column();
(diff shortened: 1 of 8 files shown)


icemelon:
I would suggest renaming the folder `bytecode_unroller_circuit` to `bytecode_circuit`.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -0,0 +1,865 @@
...
+use keccak256::plain::Keccak;
+use pasta_curves::arithmetic::FieldExt;
+use std::marker::PhantomData;
+
+use super::param::{
+    BLOCK_SIZE, KECCAK_INPUT_WIDTH, KECCAK_OUTPUT_WIDTH, WORD_SIZE,
+};
+
+#[derive(Clone, Debug)]
+pub struct BytecodeUnrollerConfig<F> {
```
DreamWuGit, line 25: I think placing  bytecode circuit to bussmapping repo more reasonable, like  rw_table , evm _circuit just take use of it, maybe good modularization, what do you think?

Brechtpd: I followed the example of the state circuit which is also put into a subfolder in zkevm-circuits. For the actual circuit part I think you would expect the repo named `zkevm-circuits` to, well, contain the circuits for zkevm. :) But I think you may be right that the busmapping repo may be the place to generate some of the input data for the circuit.

miha-stopar (approved):
Nice!

Sorry for not reviewing this PR earlier! It looks really good to me and can be merged after rebase from my point of view. At first I thought padding could be set to 0 in is_final or before, but then I saw you have a constraint for padding to be 0 when is_final, so all good!

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -0,0 +1,880 @@
...
+        // else it's push data.
+        let push_rindex_is_zero = IsZeroChip::configure(
+            meta,
+            |meta| {
+                // Conditions:
+                // - Not on the first row
+                meta.query_selector(q_enable)
+                    * not::expr(meta.query_fixed(q_first, Rotation::cur()))
+            },
+            |meta| meta.query_advice(push_rindex, Rotation::prev()),
```
DreamWuGit, line 96: one question: why not  current rotation `meta.query_advice(push_rindex, Rotation::cur())` when checking `push_rindex`?

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -0,0 +1,880 @@
...
+        };
+
+        meta.create_gate("continue", |meta| {
+            let mut cb = ConstraintBuilder::default();
+            // `index` needs to increase by 1
+            cb.require_equal(
+                meta.query_advice(index, Rotation::cur()),
+                meta.query_advice(index, Rotation::prev()) + 1u64.expr(),
+            );
+            // `is_code := push_rindex_prev == 0`
```
DreamWuGit, line 117: how about `is_code := push_rindex == 0` ?

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
Brechtpd: Certainly possible to do it in different ways, but I think always necessary to check 2 rows for some column because inherently we do not have enough data on a single row (need some history to handle the PUSH data). With how it currently works, the `push_rindex == 0` check is done against the previous row so that `push_rindex` of the current row can store the rindex for the byte in the current row (`push_rindex := is_code ? byte_push_size : push_rindex_prev - 1`). It's also possible I guess to check the value in the current row, and write the value in the next row, which is kind of the same thing but I think complicates things somewhat (because we then need special handling of the last byte which cannot use a value for the next row).

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
Brechtpd: Same explanation as above I think.

ChihChengLiang:
@Brechtpd Can we do a rebase for this?

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
DreamWuGit: sound good to me,  thanks !

DreamWuGit (approved):
LGTM !

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -0,0 +1,880 @@
...
+
+        // keccak table
+        layouter.assign_region(
+            || "keccak table",
+            |mut region| {
+                for (offset, bytecode) in
+                    bytecodes.iter().map(|v| v.bytes.clone()).enumerate()
+                {
+                    let hash: F = keccak(&bytecode[..], self.r);
+                    let rlc: F = linear_combine(bytecode.clone(), self.r);
```
DreamWuGit, line 490: one more:  keccak table validity(hash correctness) will be checked by elsewhere like RW_table, Tx_table, right?

Brechtpd: That's right, will be checked in the keccak circuit.

Brechtpd:
> @Brechtpd Can we do a rebase for this?

Done. The latest `ConstraintBuilder` contains things specifically for the opcode implementations, and I used the old `ConstraintBuilder` for just some basic constraint management. I think that still makes a lot of sense, but to do that I had to split up the `ConstraintBuilder` in a more general and basic reusable part for all circuits, and the more opcode specific part that also makes use of the basic part. I don't think it's very pretty but I haven't come up with something better yet.

ChihChengLiang:
@Brechtpd Merged, thank you for the epic work!

PR #193, opened 2021-11-23
bus-mapping: Add StateDB key-value database

ed255:
This is a small PR to introduce the StateDB key-value database that will be used in the `CircuitInputBuilder`.   It's not yet integrated because in order to use it in the `CircuitInputBuilder` we first need to fill this database with results from geth, but we don't have the methods yet, and we also need to keep track of all state operations to know what to query from geth.

Resolve https://github.com/appliedzkp/zkevm-circuits/issues/184


CPerezz (approved):
LGTM! 

Maybe now is a good time to go for #177

PR #196, opened 2021-11-24
Make gadgets to handle an execution state instead of an opcode per step

### LINKED ISSUE #68: Verify gas is indeed sufficient in success case
Currently in EVM circuit we didn't verify gas is indeed sufficient in success case. Also currently it tracks `gas_counter` as accumulated gas usage, which is not friendly to verify since we need to read out the `gas_available` of this call which we don't actually track step by step every time.

We should track `gas_left` instead and verify it's indeed sufficient in every success case naively by decode `next_state.gas_left` into 8 bytes to verify it's not underflow.

Also after `gas_left` is capped in `u64`, then everything calculated to be gas cost like `quad_memory_cost` could also be capped in 8 bytes in success case.

### LINKED ISSUE #71: [EVM] Allow access to compressed words at witness generation
Sometimes just loading the compressed value itself in the circuit is sufficient to handle certain opcodes, and so there is no need to do the compression/decompression in the circuit itself. However there currently doesn't seem a way to access the `r` value in the `assign` function for opcodes to compress the input byte values and compress the word manually at witness generation time. Making `r` available in this function somehow would solve this issue.

Another way this could be done could be to calculate the compressed value on a higher level and pass it into the op gadget, but I think that probably makes less sense.

### LINKED ISSUE #73: Implement `BusMappingLookup::Bytecode`
Currently in evm circuit the opcode is not verified with contract's bytecode if it's a contract interaction.

Some opcodes like `PUSH*` that would fetch the contract's bytecode should also be verified.

### LINKED ISSUE #74: Refactor `EvmCircuit` and internal gadgets with `Config` struct
Currently all gadget holds the constraint system related configuration directly, which makes function inputs too many and hard to manage. Having other struct `Config` to hold the configuration might improve it.

### LINKED ISSUE #88: Testing time too long
In the XOR circuit implementation (#75), I added the XOR lookup table with 2^16 rows. In order to accommodate so many rows, I need to increase the table size in Halo2 to 2^17. But after that, the testing time becomes very long. It took me more than 10 minutes to finish the test. What's worse is that after having XOR circuit, all other opcode tests need to bump up their table size. It's very likely that the Github test will timeout. We need to find a way to address this issue.

han0110:
Currently each `OpGadget` is required to be able to verify all kind of execution result of its responsible opcode, including the successful case and error cases. However, it brings much complexity and duplicated work for every `OpGadget` implementation.

This PR aims to:

- Define all possible execution result of a step to handle them without too much duplication
- Remove any unnecessary constraints to speed up proving time
- Partially Integrate crate `bus_mapping` with a temporary embedded crate which has similar structure as current implemented ones.
- Add missing `tx_table` and `bytecode_table` lookup and assigments

At the same time, it also:

- Implement `constant_gas_cost` for `OpcodeId`.
- Use `ConstraintBulider` as gadget construction helper directly, and add name to each constraint for debugging more friendly
- Simplify overall `EvmCircuit` configuration (resolves #74)
- Introduce `SameContextGadget` to check correct opcode and sufficient `gas_left` (resolves #68)
- Add opcode lookup part of `PushGadget` and remove the first selector because it's always `1` (resolves #73)
- Fix `AddGadget`'s constraint (when `ADD` we do pop `a`, `b`, then push `c`; when `SUB` we do pop `c`, `b`, then push `a`, then the addition check is always constraining `a + b == c`)
- Introduce `WordAdditionGadget` for `AddGadget` and further balance add/sub usage
- Use instance column for `randomness` and allow it be accessible at all time when configure gadget and assign witness (resolves #71)
- Mitigate the long proving time due to large table by allowing loading partial fixed table instead a complete one (resolves #88)

Also error handling like `ErrorStackOverflow`, `ErrorStackUnderflow` and `OutOfGasConstant` are discarded for next PR to handle together in the same gadget. And testing data are currently handcrafted, but will be auto-generated by the crate `bus_mapping` in the future.


zkevm-circuits/src/evm_circuit/execution/add.rs
```diff
@@ -0,0 +1,223 @@
...
+            OpcodeId::SUB.expr(),
+            OpcodeId::ADD.expr(),
+        );
+
+        // ADD: Pop a and b from the stack, push c on the stack
+        // SUB: Pop c and b from the stack, push a on the stack
+        let WordAdditionGadget { a, b, c, .. } = &word_addition;
+        cb.stack_pop(select::expr(is_sub.expr().0, c.expr(), a.expr()));
+        cb.stack_pop(b.expr());
+        cb.stack_push(select::expr(is_sub.expr().0, a.expr(), c.expr()));
```
han0110, line 56: The `is_sub` swaps `a` and `c` instead of `b` and `c` ([reference](https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/op_execution/arithmetic/add.rs#L108-L110) to previous impl)

zkevm-circuits/src/evm_circuit/execution/and.rs
```diff
@@ -0,0 +1,328 @@
...
+
+        cb.stack_pop(a.expr());
+        cb.stack_pop(b.expr());
+        cb.stack_push(c.expr());
+
+        // Because opcode AND, OR, and XOR are continuous, so we can make the
+        // FixedTableTag of them also continuous, and use the opcode delta from
+        // OpcodeId::AND as the delta to FixedTableTag::BitwiseAnd.
+        let tag = FixedTableTag::BitwiseAnd.expr()
+            + (opcode.expr() - OpcodeId::AND.as_u64().expr());
```
han0110, line 50: Use the `opcode` delta from `OpcodeId::AND` as the delta to `FixedTableTag::BitwiseAnd` just as suggested by @Brechtpd [here](https://github.com/appliedzkp/zkevm-circuits/pull/75#discussion_r748296868).

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -0,0 +1,284 @@
...
+            let byte = cb.query_cell();
+            if idx == 0 {
+                cb.opcode_lookup_at(index, byte.expr())
+            } else {
+                cb.condition(selectors[idx - 1].expr(), |cb| {
+                    cb.opcode_lookup_at(index, byte.expr())
+                });
+            }
+            byte
+        });
```
han0110, line 53: opcode lookup is added here.

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -0,0 +1,284 @@
...
+};
+use array_init::array_init;
+use bus_mapping::{eth_types::ToLittleEndian, evm::OpcodeId};
+use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Error};
+
+#[derive(Clone, Debug)]
+pub(crate) struct PushGadget<F> {
+    same_context: SameContextGadget<F>,
+    value: Word<F>,
+    selectors: [Cell<F>; 31],
```
han0110, line 26: The first `selector` is removed since it's constraint to be always `1` previously.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -0,0 +1,102 @@
...
+    ) -> Self {
+        cb.opcode_lookup(opcode.expr());
+        cb.add_lookup(Lookup::Fixed {
+            tag: FixedTableTag::ResponsibleOpcode.expr(),
+            values: [
+                cb.execution_result().as_u64().expr(),
+                opcode.expr(),
+                0.expr(),
+            ],
+        });
```
han0110, line 46: `opcode` is checked to be handled by current `execution_result` using lookup.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        if let Some(dynamic_gas_cost) = dynamic_gas_cost {
+            gas_cost = gas_cost + dynamic_gas_cost;
+        }
+
+        // Check gas_left is sufficient
+        let sufficient_gas_left = RangeCheckGadget::construct(
+            cb,
+            cb.curr.state.gas_left.expr() - gas_cost.clone(),
+        );
```
han0110, line 65: `gas_left` is checked sufficient respect to `gas_cost` by `RangeCheckGadget` with 8 bytes.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -0,0 +1,519 @@
...
+        table::{FixedTableTag, Lookup, RwTableTag, TxContextFieldTag},
+        util::{Cell, Word},
+    },
+    util::Expr,
+};
+use halo2::{arithmetic::FieldExt, plonk::Expression};
+use std::convert::TryInto;
+
+// Max degree allowed in all expressions passing through the ConstraintBuilder.
+const MAX_DEGREE: usize = 2usize.pow(3) + 1 + 32;
```
han0110, line 13: `MAX_DEGREE` is allowed more with `32` since the randomness now is from a single instance column, tho it's a constant polynomial, which should not increase polys' degree in theorem, but halo2 counts it anyway.

But if we allocate 32 instance columns, I'm afraid it would affect a lot on the benchmark of an actual prover, so here I tried to just allow it to be so large.

han0110 (commented):
Add comments on major changes.

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,552 @@
...
+            Self::ErrorOutOfGasPureMemory,
+            Self::ErrorOutOfGasSHA3,
+            Self::ErrorOutOfGasCALLDATACOPY,
+            Self::ErrorOutOfGasCODECOPY,
+            Self::ErrorOutOfGasEXTCODECOPY,
+            Self::ErrorOutOfGasRETURNDATACOPY,
+            Self::ErrorOutOfGasLOG,
+            Self::ErrorOutOfGasCALL,
+            Self::ErrorOutOfGasCALLCODE,
+            Self::ErrorOutOfGasDELEGATECALL,
```
DreamWuGit, line 223: missing error  `ErrCodeStoreOutOfGas` or `ErrorOutOfGasCREATE2' represents it ?

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,552 @@
...
+impl<F: FieldExt> Step<F> {
+    pub(crate) fn new(
+        meta: &mut ConstraintSystem<F>,
+        qs_byte_lookup: Column<Advice>,
+        advices: [Column<Advice>; STEP_WIDTH],
+        is_next_step: bool,
+    ) -> Self {
+        let state = {
+            let mut cells =
+                VecDeque::with_capacity(ExecutionResult::amount() + 10);
```
DreamWuGit, line 422: why plus 10 for cells capacity ?

zkevm-circuits/src/evm_circuit/step.rs
han0110: It's missing, thanks for catching! On my way adding this.

zkevm-circuits/src/evm_circuit/step.rs
han0110: For other stuff in `StepState`.

zkevm-circuits/src/evm_circuit/step.rs
han0110: Fixed in [`3e1241f`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/3e1241f2316bb35b370a7602da2f059a545cb09e).

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+                    self.idx += 1;
+
+                    Some([
+                        self.hash,
+                        F::from_u64(idx as u64),
+                        F::from_u64(byte as u64),
+                        F::from_u64(is_code as u64),
+                    ])
+                }
+            }
```
han0110, line 258: Bytecode table `is_code` annotation has been moved to here as iterator for `bytecode_table` assignment (previously partially done by [`assgin_byte_table_step`](https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit.rs#L1220)).

@DreamWuGit Please have a look if possible, thanks!

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
DreamWuGit: I understand  `SameContextGadget` is only used for op code successful case as in failure case execution_result may be not op code itself but errors , right ?

han0110: Yes, `ExecutionResult` it not mapping 1-1 to opcodes, it aims to flatten the possible nested errors and reduce the redundancy. And `SameContextGadget` here means the `StepState` is not going to switch to another context, and this is the case in most execution results.

icemelon (commented):
I made a few suggestions about the naming of execution result cases. I think as we separate these cases from their responsible opcodes, we don't need to follow exact opcode names and could give them a better name to summarize the responsible opcodes

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+                });
+            }
+
+            vec![0.expr()]
+        });
+
+        Self { state, rows }
+    }
+
+    pub(crate) fn q_execution_result(
```
icemelon, line 474:
```suggestion
    pub(crate) fn execution_result_selector(
```

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+    }
+}
+use bus_mapping_tmp::{Block, Call, ExecStep, Transaction};
+
+pub(crate) trait ExecutionGadget<F: FieldExt> {
+    const NAME: &'static str;
+
+    const EXECUTION_RESULT: ExecutionResult;
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self;
```
icemelon, line 375: I have a question here. So all constraints in a `ConstraintBuilder` share the same `condition`. What if we need constraints with different conditions, e.g., `CallDataCopy`? The current trait cannot support such use case.

han0110: There is a function `condition` in `ConstraintBuilder` for nested branching, I think it's enough to support complicated opcode like `CALLDATACOPY`. Would you like to show some part of the spec that you are concern that current `ConstraintBuilder` cannot support?

icemelon: Ah, I see. I missed the `condition` function. That'd be enough.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -160,35 +258,28 @@ pub struct LtGadget<F, const NUM_BYTES: usize> {
...
-    ) -> Expression<F> {
-        let diff = from_bytes::expr(self.diff.to_vec());
+    ) -> Self {
+        assert!(NUM_BYTES <= MAX_BYTES_FIELD, "unsupported number of bytes");
+
+        let lt = cb.query_bool();
+        let diff = cb.query_bytes();
+        let range = pow_of_two(NUM_BYTES * 8);
+
         // The equation we require to hold: `lhs - rhs == diff - (lt * range)`.
```
DreamWuGit, line 272: why not `lhs - rhs == diff + (lt * range) ?

han0110: When `lt`, the `lhs - rhs` would be negative value, which never fits in the `NUM_BYTES`. So adding it with the `range` we are targeting should makes `lhs - rhs + range` fit again. The equation here just moves the `+ range` to the right hand side, so it becomes `lhs - rhs == diff - (lt ⋅ range)`

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+    COINBASE,
+    TIMESTAMP,
+    NUMBER,
+    DIFFICULTY,
+    GASLIMIT,
+    CHAINID,
+    SELFBALANCE,
+    BASEFEE,
+    POP,
+    MLOAD, // MLOAD, MSTORE, MSTORE8
```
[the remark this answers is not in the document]
han0110: That's the way it's implemented currently, see [`memory.rs`](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-7fdb3d6f56c60547f56665227b9aebb5e74bc13b0e923448ff033d3bcbb65c77) for more details

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+use bus_mapping::evm::OpcodeId;
+use halo2::{
+    arithmetic::FieldExt,
+    circuit::Region,
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression},
+};
+use std::collections::VecDeque;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum ExecutionResult {
```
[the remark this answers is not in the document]
han0110: `ExecutionState` sounds much better since it's acting as a finite state machine.

zkevm-circuits/src/evm_circuit/step.rs
han0110: Fixed in [`22fd394`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/22fd39411d7083466a40fe62b2622e6b192b354d).

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+    MOD,
+    SMOD,
+    ADDMOD,
+    MULMOD,
+    EXP,
+    SIGNEXTEND,
+    LT,  // LT, GT, EQ
+    SLT, // SLT, SGT
+    ISZERO,
+    AND, // AND, OR, XOR
```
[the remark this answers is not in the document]
han0110: Fixed in [`dba3630`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/dba36302da81b8eb5da6767d441adf9a57f6d8db)

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression},
+};
+use std::collections::VecDeque;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum ExecutionResult {
+    BeginTx,
+    // Opcodes
+    STOP,
+    ADD, // ADD, SUB
```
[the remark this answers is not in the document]
han0110: I think `ADD_SUB` sounds confusing, I would still prefer just `ADD` here.

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+    MUL,
+    DIV,
+    SDIV,
+    MOD,
+    SMOD,
+    ADDMOD,
+    MULMOD,
+    EXP,
+    SIGNEXTEND,
+    LT,  // LT, GT, EQ
```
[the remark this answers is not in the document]
han0110: Fixed in [`dba3630`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/dba36302da81b8eb5da6767d441adf9a57f6d8db)

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,553 @@
...
+    SLT, // SLT, SGT
```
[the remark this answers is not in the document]
han0110: Fixed in [`dba3630`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/dba36302da81b8eb5da6767d441adf9a57f6d8db)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -0,0 +1,553 @@
...
+}
+
+impl<F> Default for Transition<F> {
+    fn default() -> Self {
+        Self::Persistent
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct StateTransition<F: FieldExt> {
```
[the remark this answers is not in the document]
han0110: Fixed in [`e91cb04`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/e91cb048ab5c717af2202aeae397f9111b9291f3)

zkevm-circuits/src/evm_circuit/step.rs
han0110: Fixed in [`57b71f2`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/57b71f29bafa5234bbc8671cc2f7d079e93e7769)

zkevm-circuits/src/evm_circuit/step.rs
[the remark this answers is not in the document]
han0110: Fixed in [`426786e`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/426786e23c0eb5d69ca65df86fe67d91b751e1d9)

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
DreamWuGit: oh, it is the moving , right !  thanks

zkevm-circuits/src/evm_circuit/step.rs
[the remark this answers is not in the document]
icemelon: sure

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,555 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct StepState<F> {
+    pub(crate) execution_state: Vec<Cell<F>>,
+    pub(crate) rw_counter: Cell<F>,
+    pub(crate) call_id: Cell<F>,
+    pub(crate) is_root: Cell<F>,
+    pub(crate) is_create: Cell<F>,
+    pub(crate) opcode_source: Cell<F>,
```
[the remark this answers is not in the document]
han0110: In most situation, it would be `bytecode_hash`, but for root creation call, the `opcode_source` is actually from transaction calldata, so it would be `tx_id`. I was planning to lookup different table (`bytecode_table` and `tx_table`) respectively, but I think it deserve more discussion and then decide how to do. So I leave a todo here at line 373.

https://github.com/appliedzkp/zkevm-circuits/blob/640757ff223ad6891850ad83b6d80efe1d9112d2/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs#L365-L386

One approach is to just lookup different table in different condition, but the `is_code` annotation then needs to be applied in both `bytecode_table` and `tx_table`, which I am not sure if acceptable.

Another one is to have `bytecode_table` to copy calldata from `tx_table` and create a actually unused `bytecode_hash` for this calldata as `opcode_source`, then we only need to do `opcode_lookup` to `bytecode_table` and `is_code` annotation only needs to be done in `bytecode_table`. However, I'm also afraid this would blow up `bytecode_table` because root creation call could be cheap to be huge, unless EIPs like [`EIP3680`](https://eips.ethereum.org/EIPS/eip-3860) is adopted, I think it somehow opens a DOS vector to prover.

Will keep the name `opcode_source` for now, and will file an issue to explain the consideration add some comment above line 395 to link to the issue. Do you think it's enough, or we should rename right now to avoid ambiguity it?

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+
+        for (cell, value) in self
+            .presets_map
+            .get(&step.execution_state)
+            .expect("not implemented")
+        {
+            cell.assign(region, offset, Some(*value))?;
+        }
+
+        macro_rules! assign_exec_step {
```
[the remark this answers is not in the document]
han0110: This is trying to reduce wordy codes here, if using generic function or closure we need to pass everything, then it doesn't help :(

Maybe using trait object here would be a final solution, but I would like to do this in further PR.

zkevm-circuits/src/evm_circuit/step.rs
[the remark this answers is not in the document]
han0110: Issue created appliedzkp/zkevm-specs#73 and commented added in [`4229f84`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/4229f841f47a8795345066430cc85873bd031005)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -0,0 +1,555 @@
...
+const LOOKUP_DEGREE: usize = 3;
+
+#[derive(Clone, Debug, Default)]
+struct StepRowUsage {
+    next_idx: usize,
+    is_byte_lookup_enabled: bool,
+}
+
+pub(crate) enum Transition<T> {
+    Persistent,
```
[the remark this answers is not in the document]
han0110: All naming suggestion applied in [`4229f84`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/4229f841f47a8795345066430cc85873bd031005)

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,555 @@
...
+                opcode_source: cells.pop_front().unwrap(),
+                program_counter: cells.pop_front().unwrap(),
+                stack_pointer: cells.pop_front().unwrap(),
+                gas_left: cells.pop_front().unwrap(),
+                memory_size: cells.pop_front().unwrap(),
+                state_write_counter: cells.pop_front().unwrap(),
+            }
+        };
+
+        let mut rows = Vec::with_capacity(STEP_HEIGHT - 4);
```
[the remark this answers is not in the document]
han0110: Fixed in [`646bc28`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/646bc28281b9de3f80c5b070fd33cc8d164c0126)

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+    use halo2::arithmetic::FieldExt;
+    use sha3::{Digest, Keccak256};
+
+    #[derive(Debug, Default)]
+    pub struct Block<F> {
+        // randomness for random linear combination
+        pub randomness: F,
+        pub txs: Vec<Transaction<F>>,
+        pub rws: Vec<Rw>,
+        pub bytecodes: Vec<Bytecode>,
```
DreamWuGit, line 73: do you plan to add block related properties (block number,  difficulty ,  coinbase, timestamp .etc.)  to support block relevant op codes?

han0110: yeah, but I don't think I should keep adding things in this PR, I can add this right after this PR and the changes should be easier to review. Do you think it's okay to add in next PR? Or is this blocking any current progress?

HablistonChris: I am reviewing those op codes about block and planning to implement ,  I can add this part later.

DreamWuGit: I am reviewing those op codes about block and planning to implement ,  I can add this part later.

han0110: Thanks, that will be great :)

zkevm-circuits/src/evm_circuit/step.rs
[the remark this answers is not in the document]
HAOYUatHZ: Isn't it corresponding to `ExecutionResult` in `go-ethereum/internal/ethapi/api.go`. If that's the case, we should keep the naming consistent with **geth**?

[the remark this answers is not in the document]
han0110: I think the enum here is more like **specify which kind of constraints the step enables** instead of **`ExecutionResult` of a call** in geth.

Currently EVM circuit is built to be like a finite state machine, each step verifies a opcode in most time (except multi-step opcode or precompiled), and the `ExecutionState` here specified which kind of constraint to enable to verify the step.

Not sure if it makes sense, any further suggestions are appreciated.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -0,0 +1,555 @@
...
+        self.query_cells::<N>(true)
+    }
+
+    fn query_cells<const N: usize>(&mut self, is_byte: bool) -> [Cell<F>; N] {
+        let mut cells = Vec::with_capacity(N);
+
+        for (row, usage) in
+            self.curr.rows.iter().zip(self.row_usages.iter_mut())
+        {
+            if usage.is_byte_lookup_enabled != is_byte {
```
icemelon, line 193: The code here requires all byte cells to be in the same row. Is this too restricted? There could be the case that you only allocate one byte cell instead of a word.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+                    for (acc, expr) in input_exprs[*index]
+                        .iter_mut()
+                        .zip(lookup.input_exprs().into_iter())
+                    {
+                        *acc = acc.clone() + expr;
+                    }
+                }
+                *index += 1;
+            }
+        }
```
Brechtpd, line 604: This took me a while to understand what exactly this does and why it works. Perhaps some comments for posterity pointing in the right direction could help speed that up. :) Something like (if I even understand correctly): lookups are reused between execution states when possible, and lookup expressions from different execution states can be added together because they are conditioned on their execution state.

zkevm-circuits/src/evm_circuit/execution/comparator.rs
```diff
@@ -0,0 +1,316 @@
...
+    },
+    util::Expr,
+};
+use bus_mapping::{eth_types::ToLittleEndian, evm::OpcodeId};
+use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Error};
+
+// ComparatorGadget verifies ADD and SUB at the same time by an extra swap flag,
+// when it's ADD, we annotate stack as [a, b, ...] and [c, ...],
+// when it's SUB, we annotate stack as [a, c, ...] and [b, ...].
+// Then we verify if a + b is equal to c.
```
Brechtpd, line 26: Some incorrect parts in the comment for the gadget.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+
+                    let idx = self.idx;
+                    let byte = self.bytes[self.idx];
+                    let mut is_code = true;
+
+                    if self.push_data_left > 0 {
+                        is_code = false;
+                        self.push_data_left -= 1;
+                    } else if (0x60..0x80).contains(&byte) {
+                        self.push_data_left = byte as usize - 0x60 + 1;
```
Brechtpd, line 246: Any reason the opcode IDs aren't used here?

zkevm-circuits/src/evm_circuit/execution.rs
DreamWuGit: sorry , just realize this mention, will have a quick look

DreamWuGit: good to me! just think using OpId maybe better

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,727 @@
...
+                    }
+                }
+                *index += 1;
+            }
+        }
+
+        macro_rules! lookup {
+            ($id:path, $table:ident) => {
+                if let Some(input_exprs) = input_exprs_map.remove(&$id) {
+                    for input_exprs in input_exprs {
```
DreamWuGit, line 609: for input_expr in input_exprs ?

zkevm-circuits/src/evm_circuit/execution.rs
han0110: Thanks, not intended to use raw byte, let me fix it.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
han0110: Yeah there will be the case, but when further bytes are queried, those unused cells will be returned by `ConstraintBuilder`, which would try to use all available cells queried (it panics when cells are not enough).

This is a small optimization to use an extra advice column to not blow up amount of byte lookups (that already exist in previously), let me document it now.

zkevm-circuits/src/evm_circuit/execution.rs
han0110: Sorry for the spaghetti code here, will doc this now.

You are totally right, the lookup here is try to add up all **independent lookup** input expression together, to use as less lookup argument as possible (but still we are abusing lookup). And the reason they are independent is because we have constraint one and only one of execution state would be enabled at a step.

zkevm-circuits/src/evm_circuit/execution.rs
han0110: Not intended to use raw byte here, let me fix it now.

zkevm-circuits/src/evm_circuit/execution/comparator.rs
han0110: Thanks for catching this, removed in [`665b9d1`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/665b9d10d3af47b6a1b6b7dbdcbb3158a070bf02)

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,750 @@
...
+        tx_table: TxTable,
+        rw_table: RwTable,
+        bytecode_table: BytecodeTable,
+        independent_lookups: Vec<Vec<Lookup<F>>>,
+    ) where
+        TxTable: LookupTable<F, 4>,
+        RwTable: LookupTable<F, 8>,
+        BytecodeTable: LookupTable<F, 4>,
+    {
+        // Because we constraint one and only one ExecutionState will be enabled
```
icemelon, line 597:
```suggestion
        // Because one and only one ExecutionState is enabled
```

zkevm-circuits/src/evm_circuit/execution.rs
icemelon: But there could be the case that none of the **independent lookup** is enabled for this table at a step, right? So does it mean that we should add a tuple of all zeros to each lookup table?

zkevm-circuits/src/evm_circuit/execution/memory.rs
```diff
@@ -0,0 +1,378 @@
...
+    is_mstore8: IsEqualGadget<F>,
+}
+
+impl<F: FieldExt> ExecutionGadget<F> for MemoryGadget<F> {
+    const NAME: &'static str = "MEMORY";
+
+    const EXECUTION_STATE: ExecutionState = ExecutionState::MEMORY;
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let opcode = cb.query_cell();
```
icemelon, line 43: One question here. Why not put `opcode` into the `MemoryGadget` struct while everything else is in the struct? I also didn't the assignment to this `opcode` cell.

zkevm-circuits/src/evm_circuit/execution.rs
han0110: I think there are 2 questions mixed here:

> But there could be the case that none of the independent lookup is enabled for this table at a step

Yeah, in this case, the lookup argument will then not configured.

> does it mean that we should add a tuple of all zeros to each lookup table?

This is already done [here](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-b4e5f73a1df679165a0103a62054eeff42c1bcbba32eff0ba8287c52740f2738R65-R70), [here](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-b4e5f73a1df679165a0103a62054eeff42c1bcbba32eff0ba8287c52740f2738R158-R165), [here](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-b4e5f73a1df679165a0103a62054eeff42c1bcbba32eff0ba8287c52740f2738R197-R204) and [here](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-b4e5f73a1df679165a0103a62054eeff42c1bcbba32eff0ba8287c52740f2738R237-R244).

zkevm-circuits/src/evm_circuit/execution/memory.rs
han0110: This is done by `SameContextGadget` ([here](https://github.com/appliedzkp/zkevm-circuits/pull/196/files#diff-08d58ebc59ab6c11bf7c049e859441eed2319a03feb3fdffaddaba73f8a8c829R90-R91)), was intended to reduce the redudancy. Let me try to fix this ambiguity by move the opcode query into the `SameContextGadget`.

zkevm-circuits/src/evm_circuit/execution.rs
han0110: Fixed in [`0b68140`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/0b681404db2130a86297f2fb325fd69ba3101501)

zkevm-circuits/src/evm_circuit/execution/memory.rs
han0110: Just realized why I left opcode in `ExecutionGadget`, because some of step state transition depends on the opcode, so we need to query the opcode first to generate such expression.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1,1257 +1,284 @@
...
-                    )?;
-                    for (idx, column) in
-                        self.fixed_table[2..].iter().enumerate()
-                    {
+                for (offset, row) in std::iter::once([
+                    F::zero(),
+                    F::zero(),
+                    F::zero(),
+                    F::zero(),
+                ])
```
spartucus, line 70:
```suggestion
                for (offset, row) in std::iter::once([F::zero(); 4])
```

han0110: Thanks, fixed in [`13efaecd`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/13efaecd5548c6c15b7f60221368d5954421a6df).

zkevm-circuits/src/evm_circuit/execution/pop.rs
```diff
@@ -0,0 +1,170 @@
...
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(1.expr()),
+            program_counter: Delta(1.expr()),
+            stack_pointer: Delta(1.expr()),
+            ..Default::default()
+        };
+        let opcode = cb.query_cell();
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 48:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/signextend.rs
```diff
@@ -0,0 +1,386 @@
...
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(3.expr()),
+            program_counter: Delta(1.expr()),
+            stack_pointer: Delta(1.expr()),
+            ..Default::default()
+        };
+        let opcode = cb.query_cell();
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 148:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/byte.rs
```diff
@@ -0,0 +1,273 @@
...
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(3.expr()),
+            program_counter: Delta(1.expr()),
+            stack_pointer: Delta(1.expr()),
+            ..Default::default()
+        };
+        let opcode = cb.query_cell();
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 91:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/jump.rs
```diff
@@ -0,0 +1,205 @@
...
+        let opcode = cb.query_cell();
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(1.expr()),
+            program_counter: To(from_bytes::expr(&destination.cells)),
+            stack_pointer: Delta(1.expr()),
+            ..Default::default()
+        };
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 59:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/jumpdest.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        // State transition
+        let step_state_transition = StepStateTransition {
+            program_counter: Delta(1.expr()),
+            ..Default::default()
+        };
+        let opcode = cb.query_cell();
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 38:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/jumpi.rs
```diff
@@ -0,0 +1,273 @@
...
+        let opcode = cb.query_cell();
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(2.expr()),
+            program_counter: To(next_program_counter),
+            stack_pointer: Delta(2.expr()),
+            ..Default::default()
+        };
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 80:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/execution/memory.rs
icemelon: I see.

zkevm-circuits/src/evm_circuit/execution/pc.rs
```diff
@@ -0,0 +1,168 @@
...
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(1.expr()),
+            program_counter: Delta(1.expr()),
+            stack_pointer: Delta((-1).expr()),
+            ..Default::default()
+        };
+        let opcode = cb.query_cell();
+        let same_context = SameContextGadget::construct(
+            cb,
+            opcode,
```
icemelon, line 55:
```suggestion
            cb.query_cell(),
```

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -0,0 +1,559 @@
...
+                self.curr.state.call_id.expr(),
+                memory_address,
+                byte,
+                0.expr(),
+                0.expr(),
+            ],
+        );
+    }
+
+    pub(crate) fn memory_lookup_inner(
```
icemelon, line 480: I found the func `memory_lookup_inner` can be confusing. It has quite subtle difference from `memory_lookup` as it doesn't increase the `rw_counter_offset`. So are the `rw_lookup` and `rw_lookup_inner`. I wonder if you could come up with a better design or name for this? Exposing inner function to public is also a bit weird. Probably you can rename it to `memory_lookup_with_counter` and add doc with something like "the func doesn't increase rw_counter_offset".

zkevm-circuits/src/evm_circuit/table.rs
```diff
@@ -0,0 +1,284 @@
...
+    Fixed,
+    Tx,
+    Rw,
+    Bytecode,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum Lookup<F> {
+    Fixed {
+        tag: Expression<F>,
```
icemelon, line 196: Add some doc to each field in these lookup table?

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -1276,24 +316,83 @@ mod test {
...
 
-            // load bytecode source from test sequence
-            let bytecode_table = assgin_byte_table_step(&self.execution_steps);
+    pub(crate) fn run_test_circuit<F: FieldExt>(
+        block: Block<F>,
+        fixed_table_tags: Vec<FixedTableTag>,
+    ) -> Result<(), Vec<VerifyFailure>> {
+        let log2_ceil = |n| {
+            u32::BITS - (n as u32).leading_zeros() - (n & (n - 1) == 0) as u32
+        };
```
spartucus, line 348: Since we are using `nightly` Rust, maybe can use [log](https://doc.rust-lang.org/std/primitive.u32.html#method.log) or [log2](https://doc.rust-lang.org/std/primitive.u32.html#method.log2)?

zkevm-circuits/src/evm_circuit/execution/pop.rs
han0110: I tried but the borrow checker complained that the `ConstraintBuilder` is mutably borrwoed previously (the first argument of `SameContextGadget::construct`, do I miss something to make this work?

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
han0110: Previously it's `memory_lookup_at`, I forgot why I changed it to `inner`...   Let me update it to `memory_lookup_with_counter` with some comments, thanks!

zkevm-circuits/src/evm_circuit/execution/pop.rs
icemelon: hmm, I see. I didn't compile it. I'll take a look.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -0,0 +1,749 @@
...
+
+                vec![
+                    qs_byte_lookup.clone() * FixedTableTag::Range256.expr(),
+                    qs_byte_lookup * advice,
+                    0.expr(),
+                    0.expr(),
+                ]
+                .into_iter()
+                .zip(fixed_table.table_exprs(meta).to_vec().into_iter())
+                .map(|(input, table)| (input, table))
```
spartucus, line 489:
```suggestion
```

zkevm-circuits/src/evm_circuit/execution/add.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        // ADD: Pop a and b from the stack, push c on the stack
+        // SUB: Pop c and b from the stack, push a on the stack
+        cb.stack_pop(select::expr(is_sub.expr().0, c.expr(), a.expr()));
+        cb.stack_pop(b.expr());
+        cb.stack_push(select::expr(is_sub.expr().0, a.expr(), c.expr()));
+
+        // State transition
+        let step_state_transition = StepStateTransition {
+            rw_counter: Delta(3.expr()),
```
spartucus, line 62: Do we have a plan to handle these constant numbers in future PR?

zkevm-circuits/src/evm_circuit/table.rs
han0110: Added in [`3c57d83`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/3c57d83839d68064d81fea52ef9a55ec82a3ef5b).

zkevm-circuits/src/evm_circuit/execution.rs
han0110: Fixed in [`3c57d83`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/3c57d83839d68064d81fea52ef9a55ec82a3ef5b)

zkevm-circuits/src/evm_circuit.rs
han0110: I think using `u32::BITS - (n as u32).leading_zeros()` as `log2` would be enough (I am afraid introducing `#![feature(...)]` would also introduce other issues), furthermore the new feature `log2` seems to round down instead of ceiling, so I would prefer keep it unchanged here.

icemelon (approved):
LGTM

zkevm-circuits/src/evm_circuit/execution/add.rs
han0110: You mean a mapping from opcode to these constant like the constant gas? I think that will be helpful for stack only opcodes.

DreamWuGit (approved):
LGTM, thanks the work!

zkevm-circuits/src/evm_circuit/execution/add.rs
spartucus: Yes

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
```diff
@@ -0,0 +1,154 @@
...
+        // Check if the memory address is too large
+        let address_in_range =
+            IsZeroGadget::construct(cb, address_high::expr(&address));
+        // Check if the amount of gas available is less than the amount of gas
+        // required
+        let (_, memory_gas_cost) = memory_expansion.expr();
+        let insufficient_gas = LtGadget::construct(
+            cb,
+            cb.curr.state.gas_left.expr(),
+            GasCost::FASTEST.expr() + memory_gas_cost,
```
spartucus, line 74: I don't understand here why need a extra `GasCost::FASTEST`?

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -0,0 +1,565 @@
...
+            Self::ErrorOutOfGasCREATE2,
+            Self::ErrorOutOfGasSTATICCALL,
+        ]
+        .iter()
+        .copied()
+    }
+
+    pub(crate) fn amount() -> usize {
+        Self::iterator().count()
+    }
```
spartucus, line 235: A little nitpick:
There is a crate called [strum](https://crates.io/crates/strum), it has macros `EnumIter` and `EnumCount` do the exact functions.
You can just add these two attributes, it can save a lot codes.

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
han0110: That is the constant gas cost of `MLOAD/MSTORE/MSTORE8`, the name of `memory_gas_cost` seem confusing, `memory_expansion_gas_cost` might be better. Fixed in [`9cc8bc7`](https://github.com/appliedzkp/zkevm-circuits/pull/196/commits/9cc8bc77726149d1b6abb225c11c409bbe1d54d8).

(In successful case, the constant gas is added in `SameContextGadget`)

spartucus: Cool!

ChihChengLiang (approved):
Untested approval to unblock the merge

zkevm-circuits/src/evm_circuit/step.rs
han0110: Thanks! Will try it in further PR.

PR #198, opened 2021-11-25
Mitigate the long testing time by allowing partial fixed table

han0110:
This PR aims to mitigate the long testing time due to large table required by bitwise opcode gadget. It adds a flag `including_large_tables` when `load_fixed_tables` to decide if to include large table or not.


icemelon (approved):
LGTM. The long testing time is quite annoying.

DreamWuGit (approved):
just suffer it at noon ;). thanks for mitigating it!

Issue #194, opened 2021-11-23
bus-mapping: Add eth_getCode JSON-RPC


ed255:
Implement this RPC method in our client: https://eth.wiki/json-rpc/API#eth_getcode
This is required because the circuit needs the bytecode of each smart contract executed in the trace.

PR #202, opened 2021-11-29
Issue/194

NoCtrlZ:
Hi @ed255 
I drafted `eth_getCode` method issue https://github.com/appliedzkp/zkevm-circuits/issues/194.
I also setup the test env with `ganache` which supports same RPC with geth.
By indicating same mnemonic, we can produce same contract addresses and block hashes.
You can check my code with `$ docker-compose up` and `cd bus-mapping && cargo test -- --ignored`.

Thank you!


bus-mapping/src/rpc.rs
```diff
@@ -119,6 +119,17 @@ impl<P: JsonRpcClient> GethClient<P> {
             .map_err(|e| Error::JSONRpcError(e.into()))?;
         Ok(resp.0.into_iter().map(|step| step.result).collect())
     }
+
+    /// Calls `eth_getCode` via JSON-RPC returning a contract code
+    pub async fn get_code_by_address(
+        &self,
+        contract_address: String,
```
spartucus, line 126: nitpick: For contract address, use `Address` instead of `String` would be better, what do you think?

NoCtrlZ: Thank you for the review!
I think that's more beautiful.
I am going to change.

bus-mapping/src/rpc.rs
```diff
@@ -119,6 +119,20 @@ impl<P: JsonRpcClient> GethClient<P> {
             .map_err(|e| Error::JSONRpcError(e.into()))?;
         Ok(resp.0.into_iter().map(|step| step.result).collect())
     }
+
+    /// Calls `eth_getCode` via JSON-RPC returning a contract code
+    pub async fn get_code_by_address(
+        &self,
+        contract_address: Address,
+    ) -> Result<String, Error> {
```
ed255, line 127: This endpoint returns the code as a hex string.  Could you decode the hex string and return a `Vec<u8>`?  This way we get the exact type we expect.

bus-mapping/src/rpc.rs
```diff
@@ -199,4 +213,17 @@ mod rpc_tests {
...
         assert!(!trace_by_hash[0].struct_logs.is_empty());
     }
+
+    #[ignore]
+    #[tokio::test]
+    async fn test_get_contract_code() {
+        let transport = Http::new(Url::parse("http://localhost:8545").unwrap());
+        let prov = GethClient::new(transport);
+        let contract_address =
+            Address::from_slice(b"0xD5F110B3E81dE87F22Fa8c5e668a5Fc541c54E3d");
```
ed255, line 223: Can you confirm if you've this test locally successfully?
The method `Address::from_slice` is not supposed to take a hex string.  For that you can use `Address::from_str` or the `address` macro from `crate::address`.

ed255 (changes requested):

ed255:
> Hi @ed255 I drafted `eth_getCode` method issue #194. I also setup the test env with `ganache` which supports same RPC with geth. By indicating same mnemonic, we can produce same contract addresses and block hashes. You can check my code with `$ docker-compose up` and `cd bus-mapping && cargo test -- --ignored`.

I think ganache will be very limiting for our tests, because it doesn't support all the RPC endpoints that geth does.  Are you familiar with operating geth?  If so, do you think it's feasible to replace ganache with geth, so that the docker-compose starts a geth instance in development mode, and the js scripts deploy the contract there?  This way we could lay a path towards an integration test that will allow us to test complete flows.

bus-mapping/src/rpc.rs
NoCtrlZ: Yeah I fixed.

bus-mapping/src/rpc.rs
NoCtrlZ: I changed `String` to `Vec<u8>`.

NoCtrlZ:
> If so, do you think it's feasible to replace ganache with geth, so that the docker-compose starts a geth instance in development mode, and the js scripts deploy the contract there?

Yeah it's piece of cake 🧁

> This way we could lay a path towards an integration test that will allow us to test complete flows.

I agree with that!

bus-mapping/src/rpc.rs
ed255: I think you forgot to decode the hex!  You can do that very easily using this type https://docs.rs/ethers-core/0.5.6/ethers_core/types/struct.Bytes.html like:
```rust
let code: Bytes = ...
// ...
code.to_vec()
```

NoCtrlZ:
There are two problems happening.
- `deployer` container can't access through host `geth`.
- web3 transaction is not confirmed.

```
➜  geth git:(issue/194) ✗ yarn deploy
yarn run v1.21.1
$ ts-node utils/deploy.ts
^C
➜  geth git:(issue/194) ✗ yarn deploy
yarn run v1.21.1
$ ts-node utils/deploy.ts
(node:39548) UnhandledPromiseRejectionWarning: Error: Returned error: already known
```
The assumption is that the transaction pool is not included to block.

NoCtrlZ:
Okay!

PR #230, opened 2021-12-15
Implement `ExecutionState::BeginTx`

han0110:
This PR aims to implement `ExecutionGadget` of `ExecutionState::BeginTx`.

The spec it follows could be found [here](https://github.com/han0110/zkevm-specs/blob/feature%2Fuse-legacy-tx-format/src/zkevm_specs/evm/execution/begin_tx.py).

Integration with `witness.rs` is left for further PRs, or left for `CALL` implementation PR

### DIFF
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -106,11 +106,12 @@ mod test {
     };
     use bus_mapping::eth_types::Word;
     use halo2::{
-        arithmetic::FieldExt,
+        arithmetic::{BaseExt, FieldExt},
         circuit::{Layouter, SimpleFloorPlanner},
         dev::{MockProver, VerifyFailure},
         plonk::{Advice, Circuit, Column, ConstraintSystem, Error},
     };
+    use pairing::bn256::Fr as Fp;
     use rand::{
         distributions::uniform::{SampleRange, SampleUniform},
         random, thread_rng, Rng,
@@ -134,6 +135,10 @@ mod test {
 
     pub(crate) fn rand_word() -> Word {
         Word::from_big_endian(&rand_bytes_array::<32>())
+    }
+
+    pub(crate) fn rand_fp() -> Fp {
+        Fp::rand()
     }
 
     #[derive(Clone)]
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -94,6 +94,15 @@ impl<F: FieldExt> EvmCircuit<F> {
     ) -> Result<(), Error> {
         self.execution.assign_block(layouter, block)
     }
+
+    /// Assign steps in incomplete block for unit test purpose
+    pub fn assign_incomplete_block(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        block: &Block<F>,
+    ) -> Result<(), Error> {
+        self.execution.assign_incomplete_block(layouter, block)
+    }
 }
 
 #[cfg(test)]
@@ -340,7 +349,9 @@ mod test {
                 &self.block.bytecodes,
                 self.block.randomness,
             )?;
-            config.evm_circuit.assign_block(&mut layouter, &self.block)
+            config
+                .evm_circuit
+                .assign_incomplete_block(&mut layouter, &self.block)
         }
     }
 
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -95,13 +95,13 @@ impl<F: FieldExt> EvmCircuit<F> {
         self.execution.assign_block(layouter, block)
     }
 
-    /// Assign steps in incomplete block for unit test purpose
-    pub fn assign_incomplete_block(
+    /// Assign exact steps in block without padding for unit test purpose
+    pub fn assign_block_exact(
         &self,
         layouter: &mut impl Layouter<F>,
         block: &Block<F>,
     ) -> Result<(), Error> {
-        self.execution.assign_incomplete_block(layouter, block)
+        self.execution.assign_block_exact(layouter, block)
     }
 }
 
@@ -351,7 +351,7 @@ mod test {
             )?;
             config
                 .evm_circuit
-                .assign_incomplete_block(&mut layouter, &self.block)
+                .assign_block_exact(&mut layouter, &self.block)
         }
     }
 
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -379,8 +379,6 @@ impl<F: FieldExt> ExecutionConfig<F> {
         layouter: &mut impl Layouter<F>,
         block: &Block<F>,
     ) -> Result<(), Error> {
-        // TODO: Assign region upto the desired capacity
-
         layouter.assign_region(
             || "Execution step",
             |mut region| {
@@ -408,11 +406,15 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 }
                 Ok(())
             },
-        )
+        )?;
+
+        // TODO: Pad leftover region to the desired capacity
+
+        Ok(())
     }
 
-    /// Assign steps in incomplete block for unit test purpose
-    pub fn assign_incomplete_block(
+    /// Assign exact steps in block without padding for unit test purpose
+    pub fn assign_block_exact(
         &self,
         layouter: &mut impl Layouter<F>,
         block: &Block<F>,
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -78,6 +78,7 @@ pub(crate) trait ExecutionGadget<F: FieldExt> {
 #[derive(Clone, Debug)]
 pub(crate) struct ExecutionConfig<F> {
     q_step: Selector,
+    q_step_first: Selector,
     step: Step<F>,
     presets_map: HashMap<ExecutionState, Vec<Preset<F>>>,
     add_gadget: AddGadget<F>,
@@ -116,6 +117,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
         BytecodeTable: LookupTable<F, 4>,
     {
         let q_step = meta.complex_selector();
+        let q_step_first = meta.complex_selector();
         let qs_byte_lookup = meta.advice_column();
         let advices = [(); STEP_WIDTH].map(|_| meta.advice_column());
 
@@ -135,28 +137,44 @@ impl<F: FieldExt> ExecutionConfig<F> {
 
         meta.create_gate("Constrain execution state", |meta| {
             let q_step = meta.query_selector(q_step);
+            let q_step_first = meta.query_selector(q_step_first);
 
             // Only one of execution_state should be enabled
-            let sum_to_one = step_curr
-                .state
-                .execution_state
-                .iter()
-                .fold(1.expr(), |acc, cell| acc - cell.expr());
+            let sum_to_one = (
+                "Only one of execution_state should be enabled",
+                step_curr
+                    .state
+                    .execution_state
+                    .iter()
+                    .fold(1.expr(), |acc, cell| acc - cell.expr()),
+            );
 
             // Cells representation for execution_state should be bool.
-            let bool_checks = step_curr
-                .state
-                .execution_state
-                .iter()
-                .map(|cell| cell.expr() * (1.expr() - cell.expr()));
+            let bool_checks =
+                step_curr.state.execution_state.iter().map(|cell| {
+                    (
+                        "Representation for execution_state should be bool",
+                        cell.expr() * (1.expr() - cell.expr()),
+                    )
+                });
+
+            // TODO: ExecutionState transition needs to be constraint to avoid
+            // transition from non-terminator to BeginTx.
+
+            let first_step_check = {
+                let begin_tx_selector =
+                    step_curr.execution_state_selector(ExecutionState::BeginTx);
+                std::iter::once((
+                    "First step should be BeginTx",
+                    q_step_first * (begin_tx_selector - 1.expr()),
+                ))
+            };
 
             std::iter::once(sum_to_one)
                 .chain(bool_checks)
-                .map(move |poly| q_step.clone() * poly)
+                .map(move |(name, poly)| (name, q_step.clone() * poly))
+                .chain(first_step_check)
         });
-
-        // TODO: ExecutionState transition needs to be constraint to avoid
-        // transition from non-terminator to BeginTx.
 
         // Use qs_byte_lookup as selector to do byte range lookup on each advice
         // column. In this way, ExecutionGadget could enable the byte range
@@ -184,6 +202,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 Self::configure_gadget(
                     meta,
                     q_step,
+                    q_step_first,
                     &randomness,
                     &step_curr,
                     &step_next,
@@ -195,6 +214,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
 
         let config = Self {
             q_step,
+            q_step_first,
             add_gadget: configure_gadget!(),
             mul_gadget: configure_gadget!(),
             bitwise_gadget: configure_gadget!(),
@@ -231,9 +251,11 @@ impl<F: FieldExt> ExecutionConfig<F> {
         config
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn configure_gadget<G: ExecutionGadget<F>>(
         meta: &mut ConstraintSystem<F>,
         q_step: Selector,
+        q_step_first: Selector,
         randomness: &Expression<F>,
         step_curr: &Step<F>,
         step_next: &Step<F>,
@@ -249,20 +271,26 @@ impl<F: FieldExt> ExecutionConfig<F> {
 
         let gadget = G::configure(&mut cb);
 
-        let (constraints, lookups, presets) = cb.build();
+        let (constraints, constraints_first_step, lookups, presets) =
+            cb.build();
         assert!(
             presets_map.insert(G::EXECUTION_STATE, presets).is_none(),
             "execution state already configured"
         );
 
-        if !constraints.is_empty() {
-            meta.create_gate(G::NAME, |meta| {
-                let q_step = meta.query_selector(q_step);
+        for (selector, constraints) in [
+            (q_step, constraints),
+            (q_step_first, constraints_first_step),
+        ] {
+            if !constraints.is_empty() {
+                meta.create_gate(G::NAME, |meta| {
+                    let selector = meta.query_selector(selector);
 
-                constraints.into_iter().map(move |(name, constraint)| {
-                    (name, q_step.clone() * constraint)
-                })
-            });
+                    constraints.into_iter().map(move |(name, constraint)| {
+                        (name, selector.clone() * constraint)
+                    })
+                });
+            }
         }
 
         // Push lookups of this ExecutionState to independent_lookups for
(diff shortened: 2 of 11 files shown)


zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -107,6 +111,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
         BytecodeTable: LookupTable<F, 4>,
     {
         let q_step = meta.complex_selector();
+        let q_step_first = meta.complex_selector();
```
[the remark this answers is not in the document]
icemelon: Ah. I found that `q_step_first` is set to 1 only for the first step in the first transaction of the block. Am I right?

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -1,6 +1,6 @@
 // Step dimension
 pub(crate) const STEP_WIDTH: usize = 32;
-pub(crate) const STEP_HEIGHT: usize = 10;
+pub(crate) const STEP_HEIGHT: usize = 14;
```
icemelon, line 3: The increase in height is due to `BeginTx` circuit?

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -200,6 +200,110 @@ impl<F: FieldExt, const N: usize> AddWordsGadget<F, N> {
     }
 }
 
+/// Construction of 256-bit product by 256-bit multiplicand * 64-bit multiplier.
+#[derive(Clone, Debug)]
+pub(crate) struct MulWordByU64Gadget<F> {
```
icemelon, line 205: There's a duplicate Multiply gadget in #255 . Probably @han0110 you can also take a look

icemelon: @noel2004 Please also take a look at this part.

noel2004: Seems this Gadget is specialized for the mutiply of `U256 * U64 mod 2^256` while our gadget can be applied for two full (U)256 integer (the I256 can also be applicable). So maybe they are for different usage caes ...

I just wonder is it a good designation to keep the constraint of mutiplier outside of the gadget？

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -24,19 +26,17 @@ pub struct Block<F> {
 
 #[derive(Debug, Default)]
 pub struct Transaction<F> {
-    // Context
     pub id: usize,
     pub nonce: u64,
     pub gas: u64,
-    pub gas_tip_cap: Word,
-    pub gas_fee_cap: Word,
+    pub gas_price: Word,
```
DreamWuGit, line 32: so where to handle gas_limit of tx ?

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -24,19 +26,17 @@ pub struct Block<F> {
...
     pub caller_address: Address,
     pub callee_address: Address,
     pub is_create: bool,
     pub value: Word,
-    pub call_data_length: usize,
     pub call_data: Vec<u8>,
-
+    pub call_data_length: usize,
+    pub call_data_gas_cost: u64,
```
DreamWuGit, line 39: what is the intention of call_data_gas_cost here?  I
 think the `gas` field counts it , or for call data out of gas purpose ?

zkevm-circuits/src/evm_circuit/execution.rs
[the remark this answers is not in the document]
han0110: Yes, the first step of the proof

zkevm-circuits/src/evm_circuit/param.rs
han0110: Yes

zkevm-circuits/src/evm_circuit/witness.rs
han0110: Do you mean checking if block's gas pool is still enough for this tx? If so, I forgot that in the spec and I will add that asap, then we can implement it in the further PRs.

zkevm-circuits/src/evm_circuit/witness.rs
han0110: It makes `BeginTx` more complicated for EVM circuit to checks `call_data` gas cost, because we need to iterate over it byte to byte to check if it's zero or not. So currently we assume this information to be prepared in `tx_table`, and it could be calculated by verifier, or verified by some other constraint enforced on `tx_table`.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
han0110: yes this is specialized for multiplication of `U256 * U64 mod 2^256`, and I think the one in #255 is more general, maybe we can extend the one in #255 to support such special case, which would not carry overflow 64-bit and could save 2 cells.

If #255 is merged first, I can do so, or we can also leave that for further PRs.

zkevm-circuits/src/evm_circuit/witness.rs
DreamWuGit: yea, that is . on the other hand, each transaction have gas limit setting when making by user, if a transaction successfully, need to constrain consumed  gas value <= gas limit

han0110: The checking of tx gas limit is not at the beginning (we charge first no matter it succeeds or not to avoid any kinds of DoS).

Ignore EIP-1559, a valid tx in a block must pass:

1. The block gas pool has sufficient `gas` for the `gas` specified in tx.
2. The signer recovered from signature has enough balance `gas * gas_price` to buy the `gas`.
3. The `gas` is enough for intrinsic usage, that is `53000 if is_create else 21000 + call_data_gas_cost`
4. ... (not gas related)

Then the next step will have `gas_left` set to `gas - intrinsic_gas_cost`. Does it make sense in this way?

DreamWuGit: sounds good , the tx table itself should have corresponding  circuit constraint like byte code table, right ?

han0110: Yes, if we decide to instantiate the `tx_table` by something like Tx circuit, then we can verify `call_data_gas_cost` is correct by some constraints (should be super easy to implement).

But currently we just let verifier build everything directly (including calculating `call_data_gas_cost` and `hash`) and instantiate the `tx_table` with only instance columns, without additional advice columns or constraint enforced.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -0,0 +1,612 @@
...
+            cb,
+            tx_gas_price.clone(),
+            tx_gas.clone(),
+            true,
+        );
+
+        // TODO: Take gas cost of access list (EIP 2930) into consideration.
+        // Use intrinsic gas
+        let intrinsic_gas_cost =
+            select::expr(tx_is_create.expr(), 53000.expr(), 21000.expr())
```
CPerezz, line 107: Should we move these numbers to a `const`. Will be easier to parametrize I think.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -0,0 +1,612 @@
...
+                tx_call_data_length.expr(),
+            ),
+            (CallContextFieldTag::Value, tx_value.expr()),
+            (CallContextFieldTag::IsStatic, 0.expr()),
+        ] {
+            cb.call_context_lookup(Some(call_id.expr()), field_tag, value);
+        }
+
+        cb.require_step_state_transition(StepStateTransition {
+            rw_counter: Delta(16.expr()),
```
CPerezz, line 171: Why `16` in particular?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -0,0 +1,612 @@
...
+            call_id: To(call_id.expr()),
+            is_root: To(1.expr()),
+            is_create: To(0.expr()),
+            opcode_source: To(code_hash.expr()),
+            program_counter: To(0.expr()),
+            stack_pointer: To(1024.expr()),
```
CPerezz, line 177: Probably using a `const` for the stack len would be better as mentioned avobe.

zkevm-circuits/src/evm_circuit/param.rs
CPerezz: Then couldn't we have `STEP_HEIGHT` as the sum of the constants that build it? On that case is easier to track this kind of modifications.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -99,3 +100,91 @@ impl<F: FieldExt> SameContextGadget<F> {
...
+        Self {
+            sub_sender_balance,
+            add_receiver_balance,
+        }
+    }
+
+    pub(crate) fn assign(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
```
CPerezz, line 170: Why using `Region` and `offset` if you can simplify with the `Layouter`?

Is it easier to manage gadgets with the region + offset approach?

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
CPerezz: Does it save constraints the fact of having the specialised gadget? 

IF so, we could maybe keep both. And use the U64 in the cases where we know the input has to be below 64bits.

WDYT?

zkevm-circuits/src/evm_circuit/witness.rs
CPerezz: Should we add this kind of work as a `bus-mapping` responsability?

CPerezz (commented):
A couple of questions.

Aside from that, brilliant and amazing work @han0110 :partying_face:

zkevm-circuits/src/evm_circuit/param.rs
han0110: The desired cells are `STEP_WIDTH * STEP_HEIGHT`, and we could tune the dimension to optimize the verifier cost.

For example (assume extra rotation is almost free for verifier thanks for SHPLONK), if we find out the extra column is required to have less rotation to verify more steps, then we can have a wider circuit. In the contrary, if we find we don't need so many steps, we can have a narrower circuit.

So I think it's more like a parameter than a value inferred from the gadget, as for why width is `32`, it's a naively value to have at least 1 million steps.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
han0110: It indeed saves some, but even with same struct, I think we can still enforce different constraint by some const generic flag. But not sure if it's worthy to add such complexity to avoid re-implementation on similar logic.

zkevm-circuits/src/evm_circuit/witness.rs
han0110: Yes eventually we should let `bus-mapping` to calculate it, but it would change a lot on `bus-mapping` to fully support generating the witness for `BeginTx`, which is already explored in further PR of `CALL` implementation. Is it acceptable to do this after that PR?

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -135,6 +145,10 @@ mod test {
         Word::from_big_endian(&rand_bytes_array::<32>())
     }
 
+    pub(crate) fn rand_fp() -> Fp {
+        Fp::rand()
+    }
+
```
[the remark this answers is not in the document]
han0110: Ah, this is trying to make creating randomness less code. If we want to use `random` directly, then we need to import the trait `ff:Field` and `rand::rngs::OsRng` every time we use it. Not sure if this is worthy for such extra function.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -338,6 +370,44 @@ impl<F: FieldExt> ExecutionConfig<F> {
         &self,
         layouter: &mut impl Layouter<F>,
         block: &Block<F>,
+    ) -> Result<(), Error> {
+        // TODO: Assign region upto the desired capacity
```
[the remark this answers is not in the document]
han0110: Even we have already assinged all steps in a block, we still need to pad the circuit to match the capacity, which is decided at keygen, right?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: Because it has `16` read/write in such step (creation tx is not yet implemented), so the `rw_counter` should increase `16` and transit to next step.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: Fixed in [`bb2f101`](https://github.com/appliedzkp/zkevm-circuits/pull/230/commits/bb2f1012ec8dd8b6d5f91f87bba2757180fdef60).

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: Fixed in [`bb2f101`](https://github.com/appliedzkp/zkevm-circuits/pull/230/commits/bb2f1012ec8dd8b6d5f91f87bba2757180fdef60).

zkevm-circuits/src/evm_circuit/param.rs
CPerezz: Hmmm I see. 
Seems that it's too complex then to compute these constants adding the contributions of the gadget compositions they represent.

I said it as it would be useful and much easier if `14` doesn't appear due to "magic" but instead appears from the computation: 
```rust
pub(crate) const STEP_HEIGHT: usize = BeginTxGadget::HEIGHT_CONTRIBUTION + OtherGadget::HEIGHT_CONTRIBUTION;
```

This was what I meant. But I can be missing some stuff so no worries.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
CPerezz: Well if it only depends on adding const-generics I think we're fine. 

The only thing I see is that the const_generics should not be `ANY` value and just `64` or `256`.
We can maybe file an issue to discuss about it.

zkevm-circuits/src/evm_circuit/witness.rs
CPerezz: Of course!! Otherways it would become massive! Makes complete sense!

zkevm-circuits/src/evm_circuit.rs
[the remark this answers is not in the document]
CPerezz: I see..
Well this is fine then I guess if the purpose is just to allow a non-verbose randomness generator mechanism.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
CPerezz: Is it padded or anything similar? What if a step requires less R/W ops?

Also, should we also parametrize this as public constants so that is easier to use them  (and change if needed at any point) much easier?

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
han0110: Ah I have missed this one. Yes using `Region` and `offset` is easier here because we are sharing a lot of cells, if using layouter we might need to calculate offset back, and leads to more complicated codes when assignment.

zkevm-circuits/src/evm_circuit/param.rs
han0110: I see, but the gadgets actually care how many cells they need than the height. Will think about how to manage this parameter more properly

CPerezz: But they're known to use some `offset` parameter which will determine the height no? Or at least a max_height that they can go up to.

Anyway, it's not needed for this PR. Just wanted to give the idea at least. But we can address it later! :)

zkevm-circuits/src/evm_circuit/execution.rs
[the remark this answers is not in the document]
CPerezz: I thought that halo2 when building behind the polynomials that need to be committed will create them with a fixed key length.

So we don't actually need to do anything in particular. It's like if dummy constraints were added to pad the circuit to the KEY size choosen.

@therealyingtong anyway will know the correct behaviour. Will ping her to see if can provide some insight.

Anyway, don't take this as a blocker to merge! :)

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
CPerezz: OOOOOH!! I see!!

Thanks for the explanation!

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
CPerezz: Hmmm. I see. You don't want to enable equalty constraints for all the things you copy right?

Makes sense then.  :)

CPerezz (approved):
LGTM now that all the topics have been clarified!

Awesome job @han0110 !!!!

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
ChihChengLiang: Can we at least add a comment of the explanation there?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -0,0 +1,621 @@
...
+        }
+    }
+
+    #[test]
+    fn begin_tx_gadget_simple() {
+        // Transfer 1 ether, successfully
+        test_ok(mock_tx(None, None, None, vec![]), true);
+
+        // Transfer 1 ether, tx reverts
+        test_ok(mock_tx(None, None, None, vec![]), false);
```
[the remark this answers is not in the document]
ChihChengLiang: Oh, I see. This test is independent of the previous one. And we arbitrarily ask it to revert.

ChihChengLiang (approved):
LGTM. This is epic work!

zkevm-circuits/src/evm_circuit/execution.rs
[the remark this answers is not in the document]
therealyingtong: Are we discussing padding the circuit with dummy inputs that still pass the gates? I do think that will need to be done at production.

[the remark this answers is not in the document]
CPerezz: We're discussing about the gadget that will contain all the tx block step's constraints. 

This has to have a maximum limit which is related to the `k` provided to the preprocessing. And we were wondering if we need to padd with dummy constraints when we have less steps than the maximum or we don't.

We say all of that because the number cannot be arbitrary. Otherwise, we will end up with different circuit descriptions for each block.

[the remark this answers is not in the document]
han0110: > So we don't actually need to do anything in particular. It's like if dummy constraints were added to pad the circuit to the KEY size choosen.

In `halo2` we always need to enable the selector manually IIUC, and the comment is a reminder of that, once we have decided how to pad the leftover steps.

> This has to have a maximum limit which is related to the `k` provided to the preprocessing.

Yes, we can do that by passing `k` as input for this function, or with const generic for this whole gadget.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: Sure, added in [`cdf00f4`](https://github.com/appliedzkp/zkevm-circuits/pull/230/commits/cdf00f4c9d047d36f7ccd5830616bd69987d90fc). Thanks!

PR #234, opened 2021-12-16
Radical Pi Gate

ChihChengLiang:
we don't need 25 columns for the pi gate.
We can use a function that takes input [(Cell, F); 25] and output [(Cell, F); 25]. Permute them at the right order. The copy constraint in the next gate will enforce what Pi intended to do.


CPerezz (approved):
Looks much better than #223 

Let's go for this then! Nice one CC!!

ChihChengLiang:
Thank you for the review Carlos!

PR #236, opened 2021-12-16
Fix `JUMP` and `JUMPI` to support 8 bytes for destination

han0110:
This PR aims to fix the implementation of `JUMP` and `JUMPI` to support 8 bytes destination just like current `geth` implementation.

It previously assumes execution bytecode has maximum size <code>2<sup>24</sup>-1</code> and fit 3 bytes, because transaction size could only be up to 128kB. But the assumption would be broken if any change made to ethereum p2p protocol to allow a larger transaction size.

Also it simplify the check `is_zero` of condition to check the random linear combination directly instead of sum of bytes. (We already assume the random linear combination is collision resistent, so if random linear combination is zero, the decompose bytes should also sum to zero.)

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/jump.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/jump.rs
@@ -1,7 +1,7 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
-        param::MAX_CODE_SIZE_IN_BYTES,
+        param::NUM_BYTES_PROGRAM_COUNTER,
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
@@ -22,7 +22,7 @@ use std::convert::TryInto;
 #[derive(Clone, Debug)]
 pub(crate) struct JumpGadget<F> {
     same_context: SameContextGadget<F>,
-    destination: RandomLinearCombination<F, MAX_CODE_SIZE_IN_BYTES>,
+    destination: RandomLinearCombination<F, NUM_BYTES_PROGRAM_COUNTER>,
 }
 
 impl<F: FieldExt> ExecutionGadget<F> for JumpGadget<F> {
@@ -80,7 +80,11 @@ impl<F: FieldExt> ExecutionGadget<F> for JumpGadget<F> {
         self.destination.assign(
             region,
             offset,
-            Some(destination.to_le_bytes()[..3].try_into().unwrap()),
+            Some(
+                destination.to_le_bytes()[..NUM_BYTES_PROGRAM_COUNTER]
+                    .try_into()
+                    .unwrap(),
+            ),
         )?;
 
         Ok(())
--- a/zkevm-circuits/src/evm_circuit/param.rs
+++ b/zkevm-circuits/src/evm_circuit/param.rs
@@ -1,7 +1,10 @@
 // Step dimension
-pub const STEP_WIDTH: usize = 32;
-pub const STEP_HEIGHT: usize = 10;
-pub const NUM_CELLS_STEP_STATE: usize = 10;
+pub(crate) const STEP_WIDTH: usize = 32;
+pub(crate) const STEP_HEIGHT: usize = 10;
+pub(crate) const NUM_CELLS_STEP_STATE: usize = 10;
+
+// The number of bytes an u64 has.
+pub(crate) const NUM_BYTES_U64: usize = 8;
 
 /// The maximum number of bytes that a field element
 /// can be broken down into without causing the value it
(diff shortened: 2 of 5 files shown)


HAOYUatHZ (approved):

DreamWuGit:
> already

So it is intended to prepare for future geth change ?

han0110:
> So it is intended to prepare for future geth change ?

No, it's just trying not to assume too much that is not specified in protocol (destination is not restricted to be in any size).

As for deployed code size, it would be verified at `RETURN`; as for tx size, it would just follow geth's p2p rule.

DreamWuGit:
> geth's p2p rule.

Got but could you share the  p2p rule more info ?  or link ?

han0110:
Here https://github.com/ethereum/go-ethereum/blob/bc6bf1e1937829b5d5b0d431a9333d47c3e08082/core/tx_pool.go#L596

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -13,9 +16,11 @@ pub(crate) const MAX_GAS_SIZE_IN_BYTES: usize = 8;
 // Number of bytes that will be used of the address word.
 // If any of the other more signficant bytes are used it will
 // always result in an out-of-gas error.
-pub const NUM_ADDRESS_BYTES_USED: usize = 5;
-pub const MAX_MEMORY_SIZE_IN_BYTES: usize = 5;
-// Number of bytes that will be used of the JUMP* destination or code copy range
-// check. Although the deployed code has maximum size of 0x6000, the size of
-// a creation transaction could be 128KB, which needs 3 bytes to cover.
-pub const MAX_CODE_SIZE_IN_BYTES: usize = 3;
+pub(crate) const NUM_ADDRESS_BYTES_USED: usize = 5;
```
DreamWuGit, line 19: irrelevant question:  does this address stands for wallet address ?  if so it is 160 bit length, 20 bytes

HAOYUatHZ: it's memory address

DreamWuGit: how to calculate 5 bytes of memory address ?

han0110: It's explained [here](https://github.com/appliedzkp/zkevm-specs/blob/master/specs/opcode/51MLOAD_52MSTORE_53MSTORE8.md#circuit-behavior), but let me quote the main reason why it's 5 bytes:

> The memory expansion calculations and the memory addressing use a partial address value consisting of the 5 LSBs of address. If any other of the higher bytes are used an out-of-gas exception is always thrown.

zkevm-circuits/src/evm_circuit/execution/jump.rs
```diff
@@ -80,7 +80,11 @@ impl<F: FieldExt> ExecutionGadget<F> for JumpGadget<F> {
         self.destination.assign(
             region,
             offset,
-            Some(destination.to_le_bytes()[..3].try_into().unwrap()),
+            Some(
+                destination.to_le_bytes()[..NUM_BYTES_PROGRAM_COUNTER]
```
HAOYUatHZ, line 84: since we use `NUM_BYTES_PROGRAM_COUNTER` to avoid hardcode magic number here.

I think we could also do the same for https://github.com/appliedzkp/zkevm-circuits/blob/8d2eea1df59d4375901be5e901b32ee62ef06d61/zkevm-circuits/src/evm_circuit/execution/memory.rs#L167 in this PR. 

(Because it's just such a tiny modification, so I have been trying to find a PR to put it into 😂 )

han0110: Sounds great, fixed in [`f6de529`](https://github.com/appliedzkp/zkevm-circuits/pull/236/commits/f6de529c2778489c7f6c8dcc783a8bb25be8b0fd)

zkevm-circuits/src/evm_circuit/param.rs
DreamWuGit: thanks I read it through  and think https://github.com/ethereum/go-ethereum/blob/master/core/vm/gas_table.go#L29#L40 constraint essentially , correct me if I am wrong .

zkevm-circuits/src/evm_circuit/execution/jump.rs
DreamWuGit: So  update ../evm_circuit/execution/pc.rs as well ?

han0110: Which part do you mean? I think assignment from `program_counter` is already 8 bytes because it's an `u64`.

DreamWuGit: yeah,  I am not sure if need limitation in configure constraint https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/pc.rs#L30 , but looks trivial!

DreamWuGit (approved):
LGTM, Thanks!

PR #238, opened 2021-12-17
gitignore circuit_benchmarks bench_params

HAOYUatHZ:
currently circuit-benchmarks/src/bench_params.rs (if ever generated) will be easily added to git by mistake. But this should depend on the ENV config of different people


CPerezz (approved):
LGTM. Will be useful for #235!! Thanks!

ed255 (approved):

PR #245, opened 2021-12-17
bus-mapping: Handle code_hash in CircuitInputBuilder, Define more Operations

### LINKED ISSUE #173: bus-mapping: build TxAccessListAccountStorage
Introduce the `AccountStorageAccess` Op in the bus-mapping: https://hackmd.io/1RXDCPagQNOAT04yA_zPKA#Read-write-unit-grouping

Every time a storage read/write happens, an access op is generated.  The value can be cold, warm.  Initially it's cold, and after the first access it becomes warm.  This can be reverted.

ed255:
NOTE: this PR contains 2 commits. They can be reviewed independently.

# bus-mapping: Handle code_hash in CircuitInputBuilder
- circuit_input_builder:
    - Add `code_source` in Call to distinguish the executed contract
      code from the contract address.
    - In every `*CALL*/CREATE*`, properly set the `code_hash` on the
      resutling `Call` object.  To do this we make use of the StateDB to
      query the `code_hash` of an address at a given execution step.
    - Require a partially-populated `StateDB` in the
      `CircuitInputBuilder` to `handle_tx`.  The accounts that are
      accessed in the block must be present in the `StateDB`.  Now the
      `CircuitInputBuilder` constructor takes a `StateDB` and `CodeDB`.
    - Move all methods of `CircuitInputStateRef` to the same impl block.
- state_db:
    - Add `CodeDB` to store contract code indexed by code hash.
- mock:
    - Introduce a helper method for the `mock::BlockData` to create a
      `CircuitInputBuilder` with the context and StateDB from the
      `BlockData`.  This way, tests don't need to manually populate the
      `StateDB` to work with the `CircuitInputBuilder`.
- zkevm-circuits:
    - Replace hardcoded geth trace in `state_circuit` test by dynamic
      trace generation using the `external_tracer`.

# bus-mapping: Define more Operations
- Define TxAccessListAccount, TxAccessListStorageSlot, TxRefund,
  Account, AccountDestructed in bus-mapping.  The only missing one now
  is CallContext.
- Resolve #173

### DIFF
--- a/bus-mapping/src/bytecode.rs
+++ b/bus-mapping/src/bytecode.rs
@@ -19,7 +19,7 @@ impl Bytecode {
     }
 
     /// Get the generated code
-    pub fn to_bytes(&self) -> Vec<u8> {
+    pub fn to_vec(&self) -> Vec<u8> {
         self.code.clone()
     }
 
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -10,7 +10,7 @@ use crate::geth_errors::*;
 use crate::operation::container::OperationContainer;
 use crate::operation::RW;
 use crate::operation::{Op, Operation};
-use crate::state_db::StateDB;
+use crate::state_db::{CodeDB, StateDB};
 use crate::Error;
 use core::fmt::Debug;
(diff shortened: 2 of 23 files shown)


han0110 (approved):
LGTM!

PR #247, opened 2021-12-18
bus_mapping: add `jump` and `jumpi` opcode

scroll-dev:


bus-mapping/src/evm/opcodes/jumpi.rs
```diff
@@ -0,0 +1,122 @@
...
+        eth_types::Word,
+        evm::StackAddress,
+        mock,
+    };
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn jumpi_opcode_impl() -> Result<(), Error> {
+        let condition = 1;
+        let destination = 35;
```
han0110, line 54: Tho it doesn't affect the zero chunk write in line 63-65, but maybe with a larger value or at least `68` makes more sense?

han0110 (approved):
LGTM! Only one unimportant nitpick.

bus-mapping/src/evm/opcodes/jumpi.rs
HAOYUatHZ: make sense! 9c97d140efc5b67bf0fd4905bde276d8f6d201c4

CPerezz (approved):
LGTM

PR #249, opened 2021-12-20
WIP op(mul)

noel2004:
Replace PR #34:

some comments by @lispc :

>I think the function result_generate should be moved out of test mod into as a method of the gadget.
This method should be called inside assign so values in ExecutionStep will only contain a,b,c. BTW maybe it can be renamed like witness_generate? This type of function(generating mid / temp values) is used widely in zk-proof, which in fact is an important part of the whole witgen/proving process. https://www.cairo-lang.org/docs/how_cairo_works/hints.html gives a great theoretical explanation of this type of computation. pow5 and pow5th give another example
remove some prints
suma/sumb/sumc: these are related to testing. They should be moved out of result_generate. btw rlc_a / ... may be better naming. Since codes of this PR are written two months ago, it is likely the summing logic is no longer consistent with main branch. I find in current main https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/add.rs#L184 just using a/b/c is ok.
add some comments of tmp vars like t0 t1 t2 t3 v0 v1
this PR is for reviewing only. I think checking out a new branch from main and copying&refactoring codes is better compared with merging this branch with main and solving lots of conflict

- [x] base on current main branch
- [x] well-form witness generating route (as 'assign_witness')
- [x] unittests
- [x] comments for tmp vars and how the proof works
- [ ] try resuming suma/b/c calculations into testing

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -28,6 +28,7 @@ mod jump;
 mod jumpdest;
 mod jumpi;
 mod memory;
+mod mul;
 mod pc;
 mod pop;
 mod push;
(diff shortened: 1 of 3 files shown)


noel2004:
sorry, wrong target

PR #252, opened 2021-12-21
bus-mapping: Add call_id in StackOp and MemoryOp

ed255:
Rename TxAccessListStorageSlot to TxAccessListAccountStorage

Rename GlobalCounter to RWCounter and gc to rwc

Since StackOp and MemoryOp now have the `call_id` field, I've added the helper functions `push_memory_op` and `push_stack_op` so that the correct `call_id` is set automatically without requiring passing the value.

I've also modified the style of the Ord implementations for the Ops to a more clean and explicit way.

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -4,12 +4,14 @@ use crate::eth_types::{
     self, Address, ChainConstants, GethExecStep, GethExecTrace, Hash,
     ToAddress, ToBigEndian, Word,
 };
-use crate::evm::{Gas, GasCost, GlobalCounter, OpcodeId, ProgramCounter};
+use crate::evm::{
+    Gas, GasCost, MemoryAddress, OpcodeId, ProgramCounter, RWCounter,
+    StackAddress,
+};
 use crate::exec_trace::OperationRef;
 use crate::geth_errors::*;
 use crate::operation::container::OperationContainer;
-use crate::operation::RW;
-use crate::operation::{Op, Operation};
+use crate::operation::{MemoryOp, Op, Operation, StackOp, RW};
 use crate::state_db::{CodeDB, StateDB};
 use crate::Error;
 use core::fmt::Debug;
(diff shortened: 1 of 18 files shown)


bus-mapping/src/circuit_input_builder.rs
```diff
@@ -452,17 +462,57 @@ pub struct CircuitInputStateRef<'a> {
...
+    /// the stored operation ([`OperationRef`]) inside the bus-mapping
+    /// instance of the current [`ExecStep`].  Then increase the `block_ctx`
+    /// [`RWCounter`] by one.
+    pub fn push_memory_op(
+        &mut self,
+        rw: RW,
+        address: MemoryAddress,
+        value: u8,
+    ) {
+        let call_id = self.call().call_id;
```
han0110, line 487: We might need to add memory read/write to caller/callee's memory, maybe in the future we need to implement `push_caller_memory_op` and `push_callee_memory_op`, or just use `push_op(MemoryOp { ... })` to set `call_id` manually.

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -1175,7 +1228,7 @@ mod tracer_tests {
                 builder,
                 tx,
                 tx_ctx: TransactionContext::new(&block.eth_tx),
-                step: ExecStep::new(geth_step, 0, GlobalCounter(0), 0),
+                step: ExecStep::new(geth_step, 0, RWCounter(0), 0),
```
han0110, line 1231: This seems not updated, tho it doesn't affect the testing result.
```suggestion
                step: ExecStep::new(geth_step, 0, RWCounter::new(), 0),
```

han0110 (approved):
LGTM!!

bus-mapping/src/circuit_input_builder.rs
ed255: > We might need to add memory read/write to caller/callee's memory, maybe in the future we need to implement `push_caller_memory_op` and `push_callee_memory_op`, or just use `push_op(MemoryOp { ... })` to set `call_id` manually.

I like `push_caller_memory_op` and `push_callee_memory_op`!  We can add them once we need them :)

PR #255, opened 2021-12-23
Circuit for opcode `MUL`

scroll-dev:
The corresponding spec is https://github.com/appliedzkp/zkevm-specs/pull/81

### DIFF
--- a/zkevm-circuits/src/evm_circuit/util/math_gadget.rs
+++ b/zkevm-circuits/src/evm_circuit/util/math_gadget.rs
@@ -11,7 +11,6 @@ use crate::{
 use bus_mapping::eth_types::{ToLittleEndian, Word};
 use halo2::plonk::Error;
 use halo2::{arithmetic::FieldExt, circuit::Region, plonk::Expression};
-use std::convert::TryInto;
 
 /// Returns `1` when `value == 0`, and returns `0` otherwise.
 #[derive(Clone, Debug)]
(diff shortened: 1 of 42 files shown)


ed255:
I have a quick question.  For the ADD and SUB opcode, we reuse the same `AddWordsGadget`: the gadget proves that `c = a + b mod 2**256`.
- ADD => inputs: a, b; output: c
- SUB => inputs: c, b, output: a

You introduce `MulWordsGadget` which proves that `c = a * b mod 2**256`.  Can the same trick be applied to also implement `DIV`?

HAOYUatHZ:
> I have a quick question. For the ADD and SUB opcode, we reuse the same `AddWordsGadget`: the gadget proves that `c = a + b mod 2**256`.
> 
> * ADD => inputs: a, b; output: c
> * SUB => inputs: c, b, output: a
> 
> You introduce `MulWordsGadget` which proves that `c = a * b mod 2**256`. Can the same trick be applied to also implement `DIV`?

@noel2004

icemelon:
@ed255 I think these two opcodes are different and probably couldn't be implemented in the same circuit. `DIV` needs to prove `a = b * c + r` where `r` is the residue.  `a, b, c, r` are all `u256` and there's no overflow.

noel2004:
> I have a quick question. For the ADD and SUB opcode, we reuse the same `AddWordsGadget`: the gadget proves that `c = a + b mod 2**256`.
> 
> * ADD => inputs: a, b; output: c
> * SUB => inputs: c, b, output: a
> 
> You introduce `MulWordsGadget` which proves that `c = a * b mod 2**256`. Can the same trick be applied to also implement `DIV`?

We considered that but found it may be not suitable, as metioned by @icemelon,  many extra constraints would be involved which just not need for MUL. For DIV we do not only to prove `a = b * c + r` but also `b * c ∈[0, 2^256)`. To prove the later costs a bunch of gates.

Of course DIV (and SDIV) would make use of the `MulWordsGadget` that is why we have wrapped it into matchgadgets.

ed255:
@icemelon @noel2004 thanks for the explanation!

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -200,6 +201,232 @@ impl<F: FieldExt, const N: usize> AddWordsGadget<F, N> {
...
+    //here we execute a multi-limbs multiplication, see spec or
+    //https://hackmd.io/HL0QhGUeQoSgIBt2el6fHA
+    // a, b and product is divided into 4 64-bit digits, call them a0 ~ a3, b0
+    // ~ b3 ... a * b = a0 * b0 + a1 * b0 ...
+    t0: Cell<F>, // a0 * b0, contribute to 0 ~ 128 bit
+    t1: Cell<F>, /* a0 * b1 + a1 * b0, contribute to 64 ~ 193 bit (include
+                  * the carry) */
+    t2: Cell<F>, // a0 * b2 + a2 * b0 + a1 * b1, contribute to above 128 bit
+    t3: Cell<F>, /* a0 * b3 + a3 * b0 + a2 * b1 + a1 * b2, contribute to
+                  * above 192 bit */
```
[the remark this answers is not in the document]
noel2004: Thx. I will update the circuit according to this advice

han0110 (approved):
LGTM!

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -200,6 +200,196 @@ impl<F: FieldExt, const N: usize> AddWordsGadget<F, N> {
...
+    ) -> Result<(), Error> {
+        use num::BigUint;
+
+        let a = BigUint::from_bytes_le(&wa.to_le_bytes());
+        let b = BigUint::from_bytes_le(&wb.to_le_bytes());
+        let c = BigUint::from_bytes_le(&wc.to_le_bytes());
+        let constant_64 = BigUint::from(1u128 << 64);
+        let constant_128 = constant_64.clone() * constant_64.clone();
+        /*let constant_256 = constant_128.clone() * constant_128.clone();
+        assert_eq!(a.clone() * b.clone() % constant_256, c);*/
```
ChihChengLiang, line 331: Do we want these?
```suggestion
```

ChihChengLiang (approved):
LGTM with a nitpick on comments.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
noel2004: yes, these code is not needed any more

PR #257, opened 2021-12-27
Add initial integration test with circuits

ed255:
Add an integration tests for the evm and state circuits, using the input
generated by the circuit input builder and data queried from geth.

bus-mapping: Introduce `BuilderClient` which wraps a GethClient and
contains methods to generate the complete circuit inputs by doing all
the necessary steps: query geth, build partial state_db, etc.  Use this
`BuilderClient` in both the circuits and circuit_input_builder
integration tests.

---

For the state circuit I've used the `pub StateCircuit`.

For the evm circuit, I didn't find a pub circuit, so I copied the `TestCircuit` used in the evm circuit tests.  I don't know if this is the best approach, please let me know if you have a better idea!

I haven't included the state_circuit integration tests into the CI because, even though the evm_circuit is passing, the state_circuit is not working.
The test tries to build the proof based on the block in which the [Greeter.sol](https://github.com/appliedzkp/zkevm-circuits/blob/main/integration-tests/contracts/greeter/Greeter.sol) contract is deployed.
Here's a log that includes the generated memory, stack and storage ops, as well as the verification error that the MockProver outputs: https://gist.github.com/ed255/ba4991427bb4cee0fa909f498e9e5786
Is the state circuit supposed to fail?  Does anybody have an idea of what may be failing, based on the generated ops by the circuit input builder?

The state circuit integration test can be reproduced following these steps:
### Setup
```sh
cd integration-tests
./run.sh --steps "setup"
./run.sh --steps "gendata"
```
### Test
```sh
RUST_LOG=circuits=trace cargo test --profile release --test circuits --features circuits test_state_circuit_block_a -- --nocapture
```
---

With the addition of the circuit integration tests, the integration tests github action take between 16m an 20m. For this reason I have disabled the trigger on pull request and left only the trigger on push to main.

### DIFF
--- a/integration-tests/tests/circuits.rs
+++ b/integration-tests/tests/circuits.rs
@@ -1,143 +1,37 @@
-//! The EVM circuit implementation.
+#![cfg(feature = "circuits")]
 
-#![allow(missing_docs)]
-use halo2::{arithmetic::FieldExt, circuit::Layouter, plonk::*};
+use bus_mapping::circuit_input_builder::BuilderClient;
+use halo2::dev::MockProver;
+use integration_tests::{get_client, log_init, GenDataOutput};
+use lazy_static::lazy_static;
+use log::trace;
+use zkevm_circuits::evm_circuit::witness::block_convert;
+use zkevm_circuits::state_circuit::StateCircuit;
 
-mod execution;
-pub mod param;
-mod step;
-mod util;
-
-pub mod table;
-pub mod witness;
-
-use execution::ExecutionConfig;
-use table::{FixedTableTag, LookupTable};
-use witness::Block;
-
-/// EvmCircuit implements verification of execution trace of a block.
-#[derive(Clone, Debug)]
-pub struct EvmCircuit<F> {
-    fixed_table: [Column<Fixed>; 4],
-    execution: ExecutionConfig<F>,
+lazy_static! {
+    pub static ref GEN_DATA: GenDataOutput = GenDataOutput::load();
 }
 
-impl<F: FieldExt> EvmCircuit<F> {
-    /// Configure EvmCircuit
-    pub fn configure<TxTable, RwTable, BytecodeTable>(
-        meta: &mut ConstraintSystem<F>,
-        randomness: Column<Instance>,
-        tx_table: TxTable,
-        rw_table: RwTable,
-        bytecode_table: BytecodeTable,
-    ) -> Self
-    where
-        TxTable: LookupTable<F, 4>,
-        RwTable: LookupTable<F, 8>,
-        BytecodeTable: LookupTable<F, 4>,
-    {
-        let fixed_table = [(); 4].map(|_| meta.fixed_column());
-
-        let execution = ExecutionConfig::configure(
-            meta,
-            randomness,
-            fixed_table,
-            tx_table,
-            rw_table,
-            bytecode_table,
-        );
-
-        Self {
-            fixed_table,
-            execution,
-        }
-    }
-
-    /// Load fixed table
-    pub fn load_fixed_table(
-        &self,
-        layouter: &mut impl Layouter<F>,
-        fixed_table_tags: Vec<FixedTableTag>,
-    ) -> Result<(), Error> {
-        layouter.assign_region(
-            || "fixed table",
-            |mut region| {
-                for (offset, row) in std::iter::once([F::zero(); 4])
-                    .chain(
-                        fixed_table_tags
-                            .iter()
-                            .map(|tag| tag.build())
-                            .flatten(),
-                    )
-                    .enumerate()
-                {
-                    for (column, value) in self.fixed_table.iter().zip(row) {
-                        region.assign_fixed(
-                            || "",
-                            *column,
-                            offset,
-                            || Ok(value),
-                        )?;
-                    }
-                }
-
-                Ok(())
-            },
-        )
-    }
-
-    /// Assign block
-    pub fn assign_block(
-        &self,
-        layouter: &mut impl Layouter<F>,
-        block: &Block<F>,
-    ) -> Result<(), Error> {
-        self.execution.assign_block(layouter, block)
-    }
-}
-
-#[cfg(test)]
-mod test {
-    use crate::evm_circuit::{
+/// This module contains a definition of a Circuit for the EVM Circuit that can
+/// be used for testing.  This is required because there's no public directly
+/// usable EVM Circuit yet.  The code in this module is copied from
+/// `zkevm_circuits::evm_circuit::test` at `zkevm-circuits/src/evm_circuit.rs`.
+mod test_evm_circuit {
+    use halo2::{
+        arithmetic::FieldExt,
+        circuit::{Layouter, SimpleFloorPlanner},
+        dev::{MockProver, VerifyFailure},
+        plonk::*,
+    };
+    use zkevm_circuits::evm_circuit::{
         param::STEP_HEIGHT,
         table::FixedTableTag,
         witness::{Block, Bytecode, Rw, Transaction},
         EvmCircuit,
     };
-    use bus_mapping::eth_types::Word;
-    use halo2::{
-        arithmetic::FieldExt,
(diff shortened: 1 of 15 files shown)


integration-tests/tests/circuits.rs
```diff
@@ -0,0 +1,344 @@
...
+        vec![
+            FixedTableTag::Range16,
+            FixedTableTag::Range32,
+            FixedTableTag::Range256,
+            FixedTableTag::Range512,
+            FixedTableTag::SignByte,
+            FixedTableTag::ResponsibleOpcode,
+        ],
+    )
+}
```
han0110, line 239: This is intended for faster unit-test purpose, so it could be removed.
```suggestion
```

han0110 (approved):
LGTM!

The reason module `evm_circuit` doesn't expose a directly usable circuit is because we haven't decide how to connect all circuits together. A direct approach is to replace the `rw_table` with `StateCircuit` and `bytecode_table` with `BytecodeCircuit` and force them to implement the `LookupTable` trait. But not sure if this is a desired approach (it hasn't taken aggregation into consideration).

But if the "integration" is for integration of bus-mapping and circuits, I think using mocking advice table is enough.

integration-tests/tests/circuits.rs
```diff
@@ -0,0 +1,344 @@
...
+use log::trace;
+use zkevm_circuits::evm_circuit::{
+    // param::STEP_HEIGHT,
+    table::FixedTableTag,
+    witness::{block_convert, Block, Bytecode, Rw, Transaction},
+    EvmCircuit,
+};
+use zkevm_circuits::state_circuit::StateCircuit;
+
+const STEP_HEIGHT: usize = 10;
```
han0110, line 21: We should use the one in `param` (need to export it) to set the enough randomness in instance column. The reason is because if we set instance column full of randomness value, it complains about there is no space for blinding factors, even we already disable the feature at the underlying proof system.

Another better approach is to handle this in `EvmCircuit` to avoid this setting, will fix this asap

ed255: I've exposed the original `STEP_HEIGHT` as pub and used that in the test for now :)

Once the `STEP_HEIGHT` is used handled in the EvmCircuit we can update this.

ed255:
> The reason module `evm_circuit` doesn't expose a directly usable circuit is because we haven't decide how to connect all circuits together. A direct approach is to replace the `rw_table` with `StateCircuit` and `bytecode_table` with `BytecodeCircuit` and force them to implement the `LookupTable` trait. But not sure if this is a desired approach (it hasn't taken aggregation into consideration).

Thanks for the explanation!  I've moved the `TestCircuitConfig` and `TestCircuit` in a module inside the test so that it's easier to see which part of the integration test is temporary and copied from the `evm_circuit` module.

> But if the "integration" is for integration of bus-mapping and circuits, I think using mocking advice table is enough.

I wanted to see how far the pieces we have could be connected between them at this stage.  The truth is that I copied the `run_test_circuit` and `TestCircuit` from `evm_circuit` without paying to much attention, and I didn't realize it was using mock advice.  I think it would be much better to use the advice provided by the circuit input builder as much as possible, but I'm not exactly sure how to proceed to achieve that.

han0110:
> I think it would be much better to use the advice provided by the circuit input builder as much as possible.

The mocking advice table is using the witness from circuit input builder, the only different is that they are not enforced with some relation (read/write consistency of `rw_table`, validity of hash of `bytecode_table`).

I think after #170 is merged, we can try to connect it here, and State circuit might need some more extra effort for integration.

lispc:
The test of the evm citcuit **should not pass too**.    There was [a bug](https://github.com/appliedzkp/zkevm-circuits/pull/261) due to which evm circuit tests can never fail.  After the fix, the evm circuit test here will output `thread 'test_evm_circuit_block_a' panicked at 'not implemented: unimplemented opcode CALLVALUE', zkevm-circuits/src/evm_circuit/witness.rs` correctly.   

More details:
the evm trace being tested here is the deploy related bytecodes. The YUL representation of the bytecode executed in this integration test is [this constructor YUL IR](https://gist.github.com/lispc/f0f0b5e4a110fbd4b0f7e31faabe4b2a#file-greeter-yul-L11-L144). Opcode like CALLVALUE is not implemented in evm circuit, so the evm circuit test should fail.

miha-stopar:
Sorry, I totally overlooked this review was assigned to me. But it looks all good to me! I think it can be merged once conflicts and what @lispc mentioned are resolved.

ed255:
With @miha-stopar we analyzed why the state circuit test was failing.  The data generated by the circuit input builder for StackOps and MemoryOps contained some cases of a READ following a WRITE at the same address with different values.

## StackOps
After analyzing the cases I realized that this is expected in the current state because we haven't implemented all opcodes in the bus-mapping, so by skipping an opcode that writes to the stack, we skip generating the StackOp that does the WRITE.  If a following step does a READ we will get a READ which doesn't match the previous generated WRITE.

## MemoryOps
I found out that the traces returned by geth never returned memory contents.  After some digging I found that we were affected by this issue https://github.com/ethereum/go-ethereum/issues/24146 which I fixed by introducing a configuration parameter for the relevant rpc methods to request memory dumpss in the traces.   With that fixes the state circuit still fails on MemoryOps constraints: this is again because we don't have all the opcodes implemented in the bus-mapping that touch memory.  In particular, in our trace we have a `CODECOPY`, which writes to memory, but we don't have it implemented, so we generate a later READ that doesn't match the previous WRITE.

ed255:
> @ed255 Hi~ Just curious the above comment I found without some opcode like callvalue, the evm circuit test using a block with contract constructor should have failed. But i think finally the test runs well when being merged. So i wonder maybe some updates/changes are applied here? Can you give me some information? Thank you!

I just ran an integration test on `main` with a block that runs a contract deployment.  I did it like this:
```sh
cd integration-tests
$ RUST_LOG=circuits,bus_mapping=trace cargo test --profile release --test circuits --features circuits test_evm_circuit_block_deploy_greeter -- --nocapture
```

This is the result:
```
    Finished release [optimized] target(s) in 0.25s
     Running tests/circuits.rs (/home/dev/git/zkevm/zkevm-circuits/target/release/deps/circuits-eb262837322ee84e)

running 1 test
[2022-03-01T12:11:12Z WARN  bus_mapping::evm::opcodes] Using dummy gen_associated_ops for opcode CODESIZE
[2022-03-01T12:11:12Z WARN  bus_mapping::evm::opcodes] Using dummy gen_associated_ops for opcode CODECOPY
[2022-03-01T12:11:12Z WARN  bus_mapping::evm::opcodes] Using dummy gen_associated_ops for opcode SSTORE
[2022-03-01T12:11:12Z WARN  bus_mapping::evm::opcodes] Using dummy gen_associated_ops for opcode CODECOPY
[2022-03-01T12:11:12Z WARN  bus_mapping::evm::opcodes] Using dummy gen_associated_ops for opcode RETURN
thread 'test_evm_circuit_block_deploy_greeter' panicked at 'not implemented', zkevm-circuits/src/evm_circuit/witness.rs:1109:26
test test_evm_circuit_block_deploy_greeter ... FAILED

failures:

failures:
    test_evm_circuit_block_deploy_greeter

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 5 filtered out; finished in 0.04s

error: test failed, to rerun pass '--test circuits'
```

The warnings come from the `bus_mapping` because there are opcodes which are not implemented, so their associated ops are not generated.
And then the panic comes from the witness converter found in the evm circuit: it doesn't yet support code source from tx input (only from account).

So currently the integration test for evm circuit with contract deployment is not passing.

On the other hand, all the integration tests with circuits are currently disabled in the Github Actions: https://github.com/appliedzkp/zkevm-circuits/blob/a07200fb2675cada12bbcc3abc5fdaf21d101b55/.github/workflows/integration.yml#L62 So they are not triggered in PRs.

lispc:
ok got it. I thought the tests could pass. Thank you!

PR #259, opened 2022-01-03
Circuit for opcode COINBASE

scroll-dev:
The PR includes the following features:
- Add block table & lookup
- Coinbase opcode circuit
- Add tests

The spec can be found at https://github.com/appliedzkp/zkevm-specs/pull/83


DreamWuGit:
@han0110  Could you please have a look ?

zkevm-circuits/src/evm_circuit/execution/coinbase.rs
```diff
@@ -0,0 +1,174 @@
...
+impl<F: FieldExt> ExecutionGadget<F> for CoinbaseGadget<F> {
+    const NAME: &'static str = "COINBASE";
+
+    const EXECUTION_STATE: ExecutionState = ExecutionState::COINBASE;
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let coinbase_address = cb.query_cell();
+
+        // Pop the value from the stack
+        cb.stack_push(coinbase_address.expr());
```
ChihChengLiang, line 35: This should be push?
```suggestion
        // Push the value to the stack
        cb.stack_push(coinbase_address.expr());
```

DreamWuGit: thanks for catching this, updated!

bus-mapping/src/eth_types.rs
```diff
@@ -137,8 +137,9 @@ impl ToWord for Address {
 impl<F: FieldExt> ToScalar<F> for Address {
     fn to_scalar(&self) -> Option<F> {
         let mut bytes = [0u8; 32];
-        bytes[32 - Self::len_bytes()..].copy_from_slice(self.as_bytes());
-        F::from_bytes(&bytes).into()
+        bytes[..Self::len_bytes()].copy_from_slice(self.as_bytes());
+        let bytes_f = F::from_bytes(&bytes);
```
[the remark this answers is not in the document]
CPerezz: That's correct! `reverse` should do. Although you should be able to transform `Address` to `Word`. And with `Word` choose the endianness directly instead of reversing by yourself.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -389,6 +389,20 @@ impl<'a, F: FieldExt> ConstraintBuilder<'a, F> {
...
+    // block
+    pub(crate) fn block_lookup(
+        &mut self,
+        tag: Expression<F>,
+        number: Option<Expression<F>>,
+        val: Expression<F>,
+    ) {
+        self.add_lookup(Lookup::Block {
+            field_tag: tag,
+            number: number.unwrap_or_else(|| 0.expr()),
```
CPerezz, line 401: If the block comes without a number provided shouldn't we panic directly? As it means that some deep internal logic must be wrong.

CPerezz (changes requested):
Aside from @han0110's comment on the Lookup of previous blocks by hash and the panic question, the rest looks good to me!

Nice job!!! :smile:

bus-mapping/src/eth_types.rs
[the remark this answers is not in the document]
DreamWuGit: cool, I would prefer `reverse` does here :)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
DreamWuGit: if the `block` construct not initialize explicitly with a number, the default value will set to zero,  so will not panic directly, does it make sense?

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -20,6 +20,81 @@ pub struct Block<F> {
...
+                    ),
+                ],
+            ],
+            self.previous_block_hashes
+                .iter()
+                .enumerate()
+                .map(|(idx, hash)| {
+                    [
+                        F::from(BlockContextFieldTag::BlockHash as u64),
+                        F::from(idx as u64),
```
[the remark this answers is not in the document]
DreamWuGit: thanks for catching , fixed in 7ad3636

bus-mapping/src/eth_types.rs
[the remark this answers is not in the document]
DreamWuGit: fixed in 7ad3636

zkevm-circuits/src/evm_circuit/witness.rs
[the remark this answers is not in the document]
han0110: Sorry, I just found the spec is not correct... Could you update it with the fix in [`b2aa171`](https://github.com/appliedzkp/zkevm-specs/pull/84/commits/b2aa1710f7fcf1f7c0be2ee0d6be680d5abf9d0d)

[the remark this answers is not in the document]
DreamWuGit: oh, I also didn't check there very carefully :) ,  b2aa171 matches logic in https://github.com/ethereum/go-ethereum/blob/master/core/evm.go#L86L88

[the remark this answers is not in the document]
han0110: Thanks for fixing :)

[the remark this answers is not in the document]
DreamWuGit: updated in 7429811

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
CPerezz: But 0 would probbaly not be the correct value no? So we will be passing a wrong block to the circuit condtructor IIUC.

DreamWuGit: the block constants include block number will be pub input of circuit eventually as planned, so it will be easy to figure out if it is incorrect .

CPerezz: Yes, what I meant is that if you check the code, if there's any unexpected error/behavior and we receive a `None/Err`, putting a 0 will not be correct. 

Therefore the question of paniking.

DreamWuGit: Ah, that is it , at first I used `number` parameter type `Expression` directly without `Option` wrapper, but considering maybe zero in cases  , change to `Option` as now.  what is your ideas ?

bus-mapping/src/evm/opcodes/coinbase.rs
```diff
@@ -0,0 +1,91 @@
...
+            0,
+        );
+        let mut state_ref =
+            test_builder.state_ref(&mut tx, &mut tx_ctx, &mut step);
+
+        // Add the last Stack write
+        state_ref.push_stack_op(
+            RW::WRITE,
+            StackAddress::from(1024 - 1),
+            block.ctants.coinbase.to_word(),
```
DreamWuGit, line 74: I want to refactor  to  use `BlockConstants`'s coinbase field  here and remove `ChainConstants` coinbase field, not sure why exists in chain constant of bussmapping before, want to hear from you guys first @ChihChengLiang  @han0110 @CPerezz

CPerezz: Maybe @ed255 can provide some insight on the bus-mapping related stuff.
From my perspective, the coinbase will always be the same. And therefore is only passed once via `ChainConstants` instead of passing it on each tx/block.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
CPerezz: To me it looks like `number` should not be an `Option` or `Result` and instead a pure value directly. 
In that case, it's impossible to construct a non-valid `number`.

Maybe @han0110 has some insight on that. I'm just wondering that this cannot be the source of errors in the future. If it cannot, we can simply resolve the conversation.

bus-mapping/src/evm/opcodes/coinbase.rs
DreamWuGit: coinbase is only block's field not related to tx， each block can have different `coinbase` address,  chainid is global which make sense exists in `ChainConstants`

ed255: I originally put the `coinbase` in `ChainConstants`.  I had the wrong idea of what the `coinbase` was, so it was a mistake.  `coinbase` should be removed from `ChainConstants`.

The `coinbase` that you need here I think should be `block.eth_block.author`. See documentation here https://docs.rs/ethers/latest/ethers/core/types/struct.Block.html#structfield.author
Block `miner`, `author`, `coinbase` are all the same thing.

The struct `BlockConstants` defined in `external_tracer` has a name which is a bit misleading.  That struct is really a configuration parameter to setup an environment for the external tracer.

CPerezz: Ohh my bad... 

I understood coinbase on a different way...

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
CPerezz: Then you can proceed as you wish!!!

CPerezz (approved):
LGTM!!

bus-mapping/src/evm/opcodes/coinbase.rs
DreamWuGit: thanks @ed255 , I will try remove `coinbase` in `ChainConstants` at this PR!

DreamWuGit: fixed in 21ad52f

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -439,8 +515,15 @@ pub fn block_convert(
     let mut storage_ops = b.container.sorted_storage();
     storage_ops.sort_by_key(|s| usize::from(s.rwc()));
 
+    // converting to block context
+    let context = BlockContext {
+        coinbase: b.constants.coinbase,
+        ..Default::default()
```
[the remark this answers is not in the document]
DreamWuGit: changed to use BlockConstant which holds all fields now.

ed255:
@DreamWuGit it seems the branch has conflicts with main.

DreamWuGit:
>  it seems the branch has conflicts with main.
yes, will rebase to main

DreamWuGit:
> > it seems the branch has conflicts with main.
> > yes, will rebase to main

rebased in 723912d

CPerezz:
Tests are failing @DreamWuGit 

Maybe was introduced some inconistency in tbhe rebase?

han0110:
Need 3 fixes of the doc in crate `bus-mapping`

1. 
https://github.com/appliedzkp/zkevm-circuits/blob/40cc46fd044de225a07c381d1df255bc5dad1190/bus-mapping/src/lib.rs#L55-L58

```diff
+ //! use bus_mapping::external_tracer::BlockConstants;
```

2. 
https://github.com/appliedzkp/zkevm-circuits/blob/40cc46fd044de225a07c381d1df255bc5dad1190/bus-mapping/src/lib.rs#L109-L112

```diff
- //!     coinbase: Address::zero(),
```

3.
https://github.com/appliedzkp/zkevm-circuits/blob/40cc46fd044de225a07c381d1df255bc5dad1190/bus-mapping/src/lib.rs#L120-L121

```diff
- //! let mut builder =
- //!     CircuitInputBuilder::new(sdb, CodeDB::new(), &eth_block, ctants);
+ //! let mut builder = CircuitInputBuilder::new(
+ //!     sdb,
+ //!     CodeDB::new(),
+ //!     &eth_block,
+ //!     ctants.clone(),
+ //!     BlockConstants::from_eth_block(
+ //!         &eth_block,
+ //!         &Word::from(ctants.chain_id),
+ //!     ),
+ //! );
```

DreamWuGit:
thanks @han0110 and @CPerezz ! should be fixed by 776bc42,  just be curious that why need  commented code snippet in bus-mapping/src/lib.rs

DreamWuGit:
understood, THX！

PR #261, opened 2022-01-05
fix tx_convert in witness

lispc:
in a [previous commit](https://github.com/appliedzkp/zkevm-circuits/pull/240/files#diff-09bf944a4f9216c1053a271fea0ff2f7a985380789b990f8e4ac530159214df2R419-R421), in witness.rs, the tx steps for evm citcuit are incorrectly converted to empty vec, as a result, no steps are really tested.


han0110 (approved):
Thanks for catching this!

ChihChengLiang (approved):
LGTM

PR #262, opened 2022-01-05
public mod table

Lawliet-Chan:
We need use `table::FixedTableTag` to build circuits in external project, so I think we should make `table mod` public

### DIFF
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -6,8 +6,9 @@ use halo2::{arithmetic::FieldExt, circuit::Layouter, plonk::*};
 mod execution;
 mod param;
 mod step;
-mod table;
 mod util;
+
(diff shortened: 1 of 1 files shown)


HAOYUatHZ:
To complement, in order to run a standalone prover (which imports zkevm-circuit and proves some witnesses), we need to export `table::FixedTableTag` (according to https://github.com/scroll-tech/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit.rs#L270)

han0110 (approved):
Make sense! (#257 does the same thing [here](https://github.com/appliedzkp/zkevm-circuits/pull/257/files#diff-b4e5f73a1df679165a0103a62054eeff42c1bcbba32eff0ba8287c52740f2738L9-R9))

ChihChengLiang (approved):
LGTM

PR #265, opened 2022-01-08
bring back ethers deps to crates.io

scroll-dev:
close https://github.com/appliedzkp/zkevm-circuits/issues/233


CPerezz (approved):
LGTM!

Thanks for being paying attention to that!

PR #271, opened 2022-01-12
Fix MSIZE opcode and add doc to step

scroll-dev:
Previous the `MSIZE` circuit pushes the memory size in words to the stack. However, the memory size that needs to be pushed to stack should be the memory size in bytes (see [here](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L571) and [here](https://github.com/ethereum/go-ethereum/blob/master/core/vm/memory.go#L102)).

In addition, the memory size stored in the `StepState` is actually word size. So I change the field name from `memory_size` to `memory_word_size` to reduce the confusion in the future. This PR also adds the document to fields in the `StepState` and `ExecStep`.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -386,7 +386,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 let mut offset = 0;
                 for transaction in &block.txs {
                     for step in &transaction.steps {
-                        let call = &transaction.calls[step.call_idx];
+                        let call = &transaction.calls[step.call_id];
 
                         self.q_step.enable(&mut region, offset)?;
                         if offset == 0 {
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -386,7 +386,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 let mut offset = 0;
                 for transaction in &block.txs {
                     for step in &transaction.steps {
-                        let call = &transaction.calls[step.call_id];
+                        let call = &transaction.calls[step.call_index];
 
                         self.q_step.enable(&mut region, offset)?;
                         if offset == 0 {
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -426,7 +426,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 let mut offset = 0;
                 for transaction in &block.txs {
                     for step in &transaction.steps {
-                        let call = &transaction.calls[step.call_idx];
+                        let call = &transaction.calls[step.call_index];
 
                         self.q_step.enable(&mut region, offset)?;
                         self.assign_exec_step(
--- a/zkevm-circuits/src/evm_circuit/step.rs
+++ b/zkevm-circuits/src/evm_circuit/step.rs
@@ -387,10 +387,15 @@ impl ExecutionState {
 
 #[derive(Clone, Debug)]
 pub(crate) struct StepState<F> {
+    /// The execution state for the step
     pub(crate) execution_state: Vec<Cell<F>>,
+    /// The Read/Write counter
     pub(crate) rw_counter: Cell<F>,
+    /// The call index
     pub(crate) call_id: Cell<F>,
+    /// Whether the call is root call
     pub(crate) is_root: Cell<F>,
+    /// Whether the call is a create call
     pub(crate) is_create: Cell<F>,
     // This is the identifier of current executed bytecode, which is used to
     // lookup current executed opcode and used to do code copy. In most time,
@@ -401,10 +406,15 @@ pub(crate) struct StepState<F> {
     // issue https://github.com/appliedzkp/zkevm-specs/issues/73 for more
     // discussion.
     pub(crate) opcode_source: Cell<F>,
+    /// The program counter
     pub(crate) program_counter: Cell<F>,
+    /// The stack pointer
     pub(crate) stack_pointer: Cell<F>,
+    /// Gas left in the transaction
     pub(crate) gas_left: Cell<F>,
-    pub(crate) memory_size: Cell<F>,
+    /// Memory size in words (32 bytes)
+    pub(crate) memory_word_size: Cell<F>,
+    /// The counter for state write
     pub(crate) state_write_counter: Cell<F>,
 }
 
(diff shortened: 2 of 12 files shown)


zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -392,7 +413,7 @@ fn step_convert(
         gas_left: step.gas_left.0,
         gas_cost: step.gas_cost.as_u64(),
         opcode: Some(step.op),
-        memory_size: step.memory_size as u64 / 32, /* memory size in word */
+        memory_size: step.memory_size as u64,
```
han0110, line 416: I think we can have `memory_word_size` directly here, then save the function above.

icemelon: I would prefer to have `memory_size` (in bytes) in the witness so that in the `MSIZE` gadget we can directly assign `step.memory_size` to the msize `value` cell (see [here](https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/msize.rs#L79)) without conversion (e.g., multiply by 32). Let me know if this makes sense.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -349,7 +349,7 @@ impl<F: FieldExt> ExecutionConfig<F> {
                 let mut offset = 0;
                 for transaction in &block.txs {
                     for step in &transaction.steps {
-                        let call = &transaction.calls[step.call_idx];
+                        let call = &transaction.calls[step.call_id];
```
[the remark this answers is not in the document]
icemelon: updated

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -387,10 +387,15 @@ impl ExecutionState {
 
 #[derive(Clone, Debug)]
 pub(crate) struct StepState<F> {
+    /// The execution state for the step
     pub(crate) execution_state: Vec<Cell<F>>,
+    /// The Read/Write counter
     pub(crate) rw_counter: Cell<F>,
+    /// The call index
```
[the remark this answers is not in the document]
icemelon: updated

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -136,19 +137,38 @@ pub struct Call<F> {
 
 #[derive(Clone, Debug, Default)]
 pub struct ExecStep {
-    pub call_idx: usize,
+    /// The call index
+    pub call_id: usize,
```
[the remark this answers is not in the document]
icemelon: updated

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -202,40 +226,78 @@ impl<F: FieldExt> Transaction<F> {
...
+    /// The return data offset in the memory
     pub return_data_offset: usize,
+    /// The length of return data
     pub return_data_length: usize,
+    /// The ether amount of the transaction
     pub value: Word,
+    /// TBD, Han will update this field
     pub result: Word,
+    /// Indicate if this call and all its caller have `is_success == true`
     pub is_persistent: bool,
```
ChihChengLiang, line 262: Is `is_persistent` for the caller or for the callee? @han0110

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -202,40 +226,78 @@ impl<F: FieldExt> Transaction<F> {
...
+    /// The length of call data
     pub call_data_length: usize,
+    /// The return data offset in the memory
     pub return_data_offset: usize,
+    /// The length of return data
     pub return_data_length: usize,
+    /// The ether amount of the transaction
     pub value: Word,
+    /// TBD, Han will update this field
     pub result: Word,
```
ChihChengLiang, line 260: Can we create an issue for this? @icemelon

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -103,18 +114,31 @@ impl<F: FieldExt> BlockContext<F> {
 
 #[derive(Debug, Default)]
 pub struct Transaction<F> {
+    /// The transaction index in the block
     pub id: usize,
+    /// The sender account nounce of the transaction
```
ChihChengLiang, line 119: nitpick
```suggestion
    /// The sender account nonce of the transaction
```

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -202,40 +226,78 @@ impl<F: FieldExt> Transaction<F> {
...
+    /// The counter for state writes
     pub state_write_counter: usize,
+    /// The opcode corresponds to the step
     pub opcode: Option<OpcodeId>,
 }
 
+impl ExecStep {
+    pub fn memory_word_size(&self) -> u64 {
+        // The memory size must be multiple of words (32 bytes)
+        assert_eq!(self.memory_size % NUM_BYTES_WORD as u64, 0);
```
ChihChengLiang, line 296: @icemelon Does the `memory_size` have to be a multiple of 32 bytes? If we perform `MSTORE8`, does it result in a memory size that's not a multiple of 32 bytes?

ChihChengLiang (commented):
Looks good to me. Added 1 nitpick and some questions.

zkevm-circuits/src/evm_circuit/witness.rs
han0110: It means if state write will be persistent or not, for **each call**. See [this note](https://hackmd.io/OVFxN25cQoKrHCkrHd0Fww) for further explanation about fields `is_success`, `is_persistent` and `rw_counter_end_of_reversion`.

ChihChengLiang: Thanks for the explanation, the document makes sense to me now.

zkevm-circuits/src/evm_circuit/witness.rs
icemelon: EVM always pads the memory to word size (see [here](https://github.com/ethereum/go-ethereum/blob/master/core/vm/interpreter.go#L212-L216)). That's why `memory_size` has to be multiple of 32 bytes. I can elaborate it in the comment to address the confusion.

zkevm-circuits/src/evm_circuit/witness.rs
icemelon: Fixed in 2465e30

zkevm-circuits/src/evm_circuit/witness.rs
ChihChengLiang: The comment looks good to me

ChihChengLiang (approved):
LGTM

PR #279, opened 2022-01-14
Extend memory gadgets to handle dynamic offset and length

han0110:
This PR is part of #278, which extends memory gadgets to add/extend

- `MemoryAddressGadget` - To handle dynamic offset and length expansion. It's also useful for memory expansion opcode like `CALLDATACOPY`. 
- `MemoryExpansionGadget` - To support multiple expansion size. It's useful for `*CALL*` which assign 2 memory chunks as expansion targets.

It also refactors some frequently used constant and methods. For main changes, please refer to [`memory_gadget.rs`](https://github.com/appliedzkp/zkevm-circuits/pull/279/files#diff-19306e3baf34b1f4a23a09ee03e0f5bb5187d2e885af70cab30d81b4419c62a9)

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
@@ -130,7 +130,7 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
             region,
             offset,
             step.memory_word_size(),
-            [address_low::value::<F>(address.to_le_bytes())
+            [address_low::value(address.to_le_bytes())
                 + if is_mstore8 == F::one() { 1 } else { 32 }],
         )?;
 
--- a/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
@@ -131,8 +131,7 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
             offset,
             step.memory_word_size(),
             [address_low::value::<F>(address.to_le_bytes())
-                + 1
-                + if is_mstore8 == F::one() { 0 } else { 31 }],
+                + if is_mstore8 == F::one() { 1 } else { 32 }],
         )?;
 
         // Gas insufficient check
--- a/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
@@ -1,11 +1,11 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
-        param::MAX_MEMORY_SIZE_IN_BYTES,
+        param::{N_BYTES_GAS, N_BYTES_MEMORY_WORD_SIZE},
         step::ExecutionState,
         util::{
             constraint_builder::ConstraintBuilder,
-            math_gadget::{IsEqualGadget, IsZeroGadget, LtGadget},
+            math_gadget::{IsEqualGadget, IsZeroGadget, RangeCheckGadget},
             memory_gadget::{address_high, address_low, MemoryExpansionGadget},
             Cell, Word,
         },
@@ -21,9 +21,15 @@ pub(crate) struct ErrorOOGPureMemoryGadget<F> {
     opcode: Cell<F>,
     address: Word<F>,
     address_in_range: IsZeroGadget<F>,
+    // Allow memory size to expand to 5 bytes, because memory address could be
+    // at most 2^40 - 1, after constant division by 32, the memory word size
+    // then could be at most 2^35 - 1.
     memory_expansion:
-        MemoryExpansionGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 - 1 }>,
-    insufficient_gas: LtGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 - 1 }>,
+        MemoryExpansionGadget<F, 1, { N_BYTES_MEMORY_WORD_SIZE + 1 }>,
+    // Even memory size at most could be 2^35 - 1, the qudratic part of memory
+    // expansion gas cost could be at most 2^61 - 2^27, due to the constant
+    // division by 512, which still fits in 8 bytes.
+    insufficient_gas: RangeCheckGadget<F, N_BYTES_GAS>,
     is_mstore8: IsEqualGadget<F>,
 }
 
@@ -52,9 +58,9 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
         let memory_expansion = MemoryExpansionGadget::construct(
             cb,
             cb.curr.state.memory_word_size.expr(),
-            address_low::expr(&address)
+            [address_low::expr(&address)
                 + 1.expr()
-                + (is_not_mstore8 * 31.expr()),
+                + (is_not_mstore8 * 31.expr())],
         );
 
         // Check if the memory address is too large
@@ -62,20 +68,14 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
             IsZeroGadget::construct(cb, address_high::expr(&address));
         // Check if the amount of gas available is less than the amount of gas
         // required
-        let insufficient_gas = LtGadget::construct(
-            cb,
-            cb.curr.state.gas_left.expr(),
-            OpcodeId::MLOAD.constant_gas_cost().expr()
-                + memory_expansion.gas_cost(),
-        );
-
-        // Make sure we are out of gas
-        // Out of gas when either the address is too big and/or the amount of
-        // gas available is lower than what is required.
-        cb.require_zero(
-            "Either the address is too big or insufficient gas",
-            address_in_range.expr() * (1.expr() - insufficient_gas.expr()),
-        );
+        let insufficient_gas = cb.condition(address_in_range.expr(), |cb| {
+            RangeCheckGadget::construct(
+                cb,
+                OpcodeId::MLOAD.constant_gas_cost().expr()
+                    + memory_expansion.gas_cost()
+                    - cb.curr.state.gas_left.expr(),
+            )
+        });
 
         // Pop the address from the stack
         // We still have to do this to verify the correctness of `address`
@@ -130,9 +130,9 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
             region,
             offset,
             step.memory_word_size(),
-            address_low::value::<F>(address.to_le_bytes())
+            [address_low::value::<F>(address.to_le_bytes())
                 + 1
-                + if is_mstore8 == F::one() { 0 } else { 31 },
+                + if is_mstore8 == F::one() { 0 } else { 31 }],
         )?;
 
         // Gas insufficient check
@@ -140,8 +140,7 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
         self.insufficient_gas.assign(
             region,
             offset,
-            F::from(step.gas_left),
-            F::from(step.gas_cost),
+            F::from(step.gas_cost - step.gas_left),
         )?;
 
         Ok(())
--- a/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
@@ -24,6 +24,9 @@ pub(crate) struct ErrorOOGPureMemoryGadget<F> {
     // Allow memory size to expand to 5 bytes, because memory address could be
     // at most 2^40 - 1, after constant division by 32, the memory word size
     // then could be at most 2^35 - 1.
+    // So generic N_BYTES_MEMORY_WORD_SIZE for MemoryExpansionGadget needs to
+    // be larger by 1 than normal usage (to be 5), to be able to contain
+    // number up to 2^35 - 1.
     memory_expansion:
         MemoryExpansionGadget<F, 1, { N_BYTES_MEMORY_WORD_SIZE + 1 }>,
     // Even memory size at most could be 2^35 - 1, the qudratic part of memory
--- a/zkevm-circuits/src/evm_circuit/util.rs
+++ b/zkevm-circuits/src/evm_circuit/util.rs
@@ -1,4 +1,4 @@
-use crate::{evm_circuit::param::MAX_MEMORY_SIZE_IN_BYTES, util::Expr};
+use crate::{evm_circuit::param::N_BYTES_MEMORY_ADDRESS, util::Expr};
 use bus_mapping::eth_types::U256;
 use halo2::{
     arithmetic::FieldExt,
@@ -140,7 +140,7 @@ impl<F: FieldExt, const N: usize> Expr<F> for RandomLinearCombination<F, N> {
 
 pub(crate) type Word<F> = RandomLinearCombination<F, 32>;
 pub(crate) type MemoryAddress<F> =
-    RandomLinearCombination<F, MAX_MEMORY_SIZE_IN_BYTES>;
+    RandomLinearCombination<F, N_BYTES_MEMORY_ADDRESS>;
 
 /// Returns the sum of the passed in cells
 pub(crate) mod sum {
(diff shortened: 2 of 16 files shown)


zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -533,33 +536,36 @@ impl<F: FieldExt, const NUM_BYTES: usize> RangeCheckGadget<F, NUM_BYTES> {
 }
 
 /// Returns `1` when `lhs < rhs`, and returns `0` otherwise.
-/// lhs and rhs `< 256**NUM_BYTES`
-/// `NUM_BYTES` is required to be `<= MAX_BYTES_FIELD` to prevent overflow:
-/// values are stored in a single field element and two of these
-/// are added together.
+/// lhs and rhs `< 256**N_BYTES`
+/// `N_BYTES` is required to be `<= MAX_N_BYTES_INTEGER` to prevent overflow:
```
CPerezz, line 540: I'm confused here.
`256**N_BYTES`  would be the equivalent to `32**(N_BYTES *8)`. But not entirely sure what this value represents TBH.

CPerezz: Ohhh you do it for the range over 256. Is that the case? If so, now all makes sense..

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -533,33 +536,36 @@ impl<F: FieldExt, const NUM_BYTES: usize> RangeCheckGadget<F, NUM_BYTES> {
...
     pub(crate) fn construct(
         cb: &mut ConstraintBuilder<F>,
         lhs: Expression<F>,
         rhs: Expression<F>,
     ) -> Self {
-        assert!(NUM_BYTES <= MAX_BYTES_FIELD, "unsupported number of bytes");
+        assert!(
+            N_BYTES <= MAX_N_BYTES_INTEGER,
+            "Too many bytes to compose an integer in field"
+        );
```
CPerezz, line 564: ditto

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -1,57 +1,166 @@
...
-        from_bytes::expr(&address.cells[0..NUM_ADDRESS_BYTES_USED])
+        from_bytes::expr(&address.cells[..N_BYTES_MEMORY_ADDRESS])
     }
 
-    pub(crate) fn value<F: FieldExt>(address: [u8; 32]) -> Address {
-        from_bytes::value::<F>(&address[0..NUM_ADDRESS_BYTES_USED])
-            .get_lower_128() as Address
+    pub(crate) fn value<F: FieldExt>(address: [u8; 32]) -> u64 {
+        from_bytes::value::<F>(&address[..N_BYTES_MEMORY_ADDRESS])
+            .get_lower_128() as u64
```
CPerezz, line 38: How do you know that this will not fail? Since `get_lower_128()` returns the lower 128 bits. 
See: https://github.com/appliedzkp/pairing/blob/3f668ba2eab444ad9f4c6396db8292fada6feb42/src/arithmetic/fields.rs#L117

Therefore this should be able to fail.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -1,57 +1,166 @@
...
+        let memory_length_bytes = memory_length.to_le_bytes();
+        let memory_length_is_zero = memory_length.is_zero();
+        self.memory_offset.assign(
+            region,
+            offset,
+            Some(Word::random_linear_combine(memory_offset_bytes, randomness)),
+        )?;
+        self.memory_offset_bytes.assign(
+            region,
+            offset,
```
CPerezz, line 115: While all of the offsets in the same region contain the same value? Is everything done within the same origin row or something similar?

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -173,23 +290,32 @@ impl<F: FieldExt, const MAX_QUAD_COST_IN_BYTES: usize>
...
                 region,
                 offset,
-                F::from(address_memory_size),
-                F::from(curr_memory_size),
-            )?
-            .get_lower_128() as MemorySize;
+                F::from(next_memory_size as u64),
+                F::from(*memory_size),
+            )?;
+            next_memory_size = max.get_lower_128() as u64;
```
CPerezz, line 317: Another conversion which could in theory fail no? Is this checked previously somewhere?

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -173,23 +290,32 @@ impl<F: FieldExt, const MAX_QUAD_COST_IN_BYTES: usize>
...
+    ) -> Result<(u64, u64), Error> {
         // Calculate the active memory size
-        let address_memory_size =
-            self.address_memory_size.assign(region, offset, address)?;
+        let memory_sizes = self
+            .memory_sizes
+            .iter()
+            .zip(addresses.iter())
+            .map(|(memory_size, address)| {
+                memory_size.assign(region, offset, *address)
```
CPerezz, line 302: It's a bit strange that within the same region each size item is placed at same offset. 
Whouldn't be needed to update the offset after each assignment? 

Just a question. Haven't checked this in depth.

CPerezz (commented):
Some questions and nits.

Amazing job as always @han0110 :+1:

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
```diff
@@ -21,9 +21,14 @@ pub(crate) struct ErrorOOGPureMemoryGadget<F> {
     opcode: Cell<F>,
     address: Word<F>,
     address_in_range: IsZeroGadget<F>,
-    memory_expansion:
-        MemoryExpansionGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 - 1 }>,
-    insufficient_gas: LtGadget<F, { MAX_MEMORY_SIZE_IN_BYTES * 2 - 1 }>,
+    // Allow memory size to expand to 5 bytes, because memory address could be
+    // at most 2^40 - 1, after constant division by 32, the memory size then
+    // could be at most 2^35 - 1.
+    memory_expansion: MemoryExpansionGadget<F, 1, { N_BYTES_MEMORY_SIZE + 1 }>,
```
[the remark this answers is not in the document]
han0110: The `N_BYTES_MEMORY_SIZE` is `4`, and it's useful in successful cases. But in such OOG case, because we actually use `5` bytes for memory address, after division of `32`, it still could be `5` bytes, so here it's `4 + 1 = 5`.

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
```diff
@@ -130,18 +129,17 @@ impl<F: FieldExt> ExecutionGadget<F> for ErrorOOGPureMemoryGadget<F> {
             region,
             offset,
             step.memory_size,
-            address_low::value::<F>(address.to_le_bytes())
+            [address_low::value::<F>(address.to_le_bytes())
                 + 1
-                + if is_mstore8 == F::one() { 0 } else { 31 },
+                + if is_mstore8 == F::one() { 0 } else { 31 }],
```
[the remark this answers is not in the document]
han0110: Yes!

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -209,11 +209,14 @@ pub(crate) mod select {
...
-    use crate::{evm_circuit::param::MAX_BYTES_FIELD, util::Expr};
+    use crate::{evm_circuit::param::MAX_N_BYTES_INTEGER, util::Expr};
     use halo2::{arithmetic::FieldExt, plonk::Expression};
 
     pub(crate) fn expr<F: FieldExt, E: Expr<F>>(bytes: &[E]) -> Expression<F> {
-        assert!(bytes.len() <= MAX_BYTES_FIELD, "number of bytes too large");
+        assert!(
+            bytes.len() <= MAX_N_BYTES_INTEGER,
+            "number of bytes too large"
+        );
```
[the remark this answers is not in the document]
han0110: Yes there are, but we never know if it fits, until we check the whole number is less than the modulus. This is more like a warning to developer, if you are combining more than `31` bytes, you should know it's possible to wrap around. If we really need such feature, we can add function like `expr_unchecked` or mod like `from_bytes_unchecked`.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -493,18 +493,21 @@ impl<F: FieldExt> MulWordByU64Gadget<F> {
...
+impl<F: FieldExt, const N_BYTES: usize> RangeCheckGadget<F, N_BYTES> {
     pub(crate) fn construct(
         cb: &mut ConstraintBuilder<F>,
         value: Expression<F>,
     ) -> Self {
-        assert!(NUM_BYTES <= 31, "number of bytes too large");
+        assert!(
+            N_BYTES <= MAX_N_BYTES_INTEGER,
+            "Too many bytes to compose an integer in field"
+        );
```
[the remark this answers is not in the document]
han0110: Yes, I assume we will have field less than 256 bits.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
han0110: Yes, it verifies the range by doing `N` times `Range256` range lookup. We can further optimize it with larger chunk if needed.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: I was assuming the witness is already valid, to escape the `Option` or `Result` for such small function. There are many other places I use such methods which don't panic.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: The `offset` here is "offset of the step", and the column and rotation information is stored in `Cell` itself, to assign to correct `offset` of its value.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: The `assign` function might answer you question. Currently we don't care about `column` or `rotation` too much, to boost development and explore as much as possible, so the `column` and `rotation` are stored directly inside the struct `Cell`, for correct assignment.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: Not really, let me try to review all and propagate the `Result` out.

han0110: Ah, just realized the `lhs` and `rhs` or `MinMaxGadget` are both `u64`, so the conversion should not fail here.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: Just realized the bytes are already capped to 5 bytes, so it should not fail. But I fixed it by using `u64::from`, to avoid such confusing conversion.

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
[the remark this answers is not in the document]
CPerezz: Now it makes sense! Thanks!
Maybe we can declare a const for the OOG method or ad a comment explaining the `+1`

zkevm-circuits/src/evm_circuit/util.rs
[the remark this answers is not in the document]
CPerezz: If it's an assertion for the developer, shouldn't we make it correct for the entire ring using for example:
https://github.com/appliedzkp/pairing/blob/3f668ba2eab444ad9f4c6396db8292fada6feb42/src/bn256/fr.rs#L649

And failing if the `CtOption` is none. Which means the number is not canonical or over the modulus.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
CPerezz: Ohhh that's right! My bad!!

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
CPerezz: Yep, now I see the approach and definitely makes all the work much more easy!

zkevm-circuits/src/evm_circuit/util.rs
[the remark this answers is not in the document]
han0110: But this is more like assertion for verifier, to warn it that with such usage, the value might wrap around, and we should never allow it.

zkevm-circuits/src/evm_circuit/execution/error_oog_pure_memory.rs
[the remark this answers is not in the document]
han0110: Added in [`3909ef6`](https://github.com/appliedzkp/zkevm-circuits/pull/279/commits/3909ef67a4d14f00621818534b4bb25468b6d55f).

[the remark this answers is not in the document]
CPerezz: Ohhh I see. Maybe then just add a small explanation or a new constant for the OOG case would be sufficient :)

zkevm-circuits/src/evm_circuit/util.rs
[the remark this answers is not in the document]
CPerezz: Hmmm. If we want to warn then probable we should use `debug_assert_eq!` maybe?

See: https://doc.rust-lang.org/std/macro.debug_assert.html

I say it because this is also failing for a small set of valid values.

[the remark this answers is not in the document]
CPerezz: Just had a pairing session with @ed255 and we hit this assert on a case where the value was correct.

The workarround was to constrain the MSB to be 0 (as we were working with an address) and pass 31 bytes to the fn's with the assert.

Edu likes the assertion. So then we can keep it and close this discussion.

Apologies for the noise.

CPerezz (approved):
LGTM

zkevm-circuits/src/evm_circuit/util.rs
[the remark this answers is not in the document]
han0110: > Just had a pairing session with @ed255 and we hit this assert on a case where the value was correct.

Ah, if you are saying about #287, in such case, the address is already known 20 bytes to verifier, so we only need 20 bytes for decomposition, and then use random linear combination on it. Then we don't need to query extra 12 cells and specially constrain them to be zero.

> If we want to warn then probable we should use debug_assert_eq! maybe?

There are many raw `assert!` is used in the circuit, and I think using `debug_assert!` makes much more sense. I have filed #288 and will try to resolve in future PRs.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -1,49 +1,161 @@
...
-        sum::value::<F>(&address[NUM_ADDRESS_BYTES_USED..32])
+        sum::value::<F>(&address[N_BYTES_MEMORY_ADDRESS..])
+    }
+}
+
+/// Convert the dynamic memory offset and length from random linear combiation
+/// to integer. It handles the "no expansion" feature when length is zero.
+#[derive(Clone, Debug)]
+pub(crate) struct MemoryAddressGadget<F> {
+    memory_offset: Cell<F>,
```
DreamWuGit, line 66: `memory_offset` is starting point of address for expansion ?

han0110: Yes, more precisely, the starting point of address for `call_data` or `return_data` specified by caller.

ed255: My understanding of this gadget is that it aims to provide the max used address after a memory access by an opcode that takes `offset` and `length` like `CALLDATACOPY`, `CODECOPY`, `EXTCODECOPY`, `RETURNDATACOPY`.  And I think this max address would then be used in the `MemoryExpansionGadget` to calculate the `gas_cost` of the possible expansion.

And `MemoryAddressGadget` returns `max_address = 0` when the `length = 0` so that no memory expansion is performed in such case, even if the offset is past the current max used address.

ed255 (approved):
LGTM!

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
han0110: Exactly, thanks for explanation!

PR #281, opened 2022-01-14
Handle base conversion nicer

ChihChengLiang:
### Problem

The current codebase uses many patterns that are difficult to reason and debug.

For example, 

#### pattern 1

```
b13_chunks.iter().fold(0, |acc, x| { acc * B9 + convert_b13_coef(*x)})
```

This requires `b13_chunks` to be in big endianness, but a reader might not be aware of it.

Solution: We collect them in the same function `f_from_radix_be` so that we have to reason only once. And from the function name, we know the input must be in big-endian.

#### pattern 2

We do a lot of manual base conversion like

```
    for _ in 0..64 {
        let remainder: u64 = mod_u64(&raw, B9);
        acc += coef_transform(remainder) * base.clone();
        raw /= from_base;
        base *= to_base;
    }
```
Solution: The `bigUint` has an expressive `to_radix_be` API to allow us to perform base conversion easily.

### DIFF
--- a/keccak256/src/arith_helpers.rs
+++ b/keccak256/src/arith_helpers.rs
@@ -1,14 +1,14 @@
 use crate::common::State;
 use itertools::Itertools;
 use num_bigint::BigUint;
-use num_traits::{One, Zero};
+use num_traits::Zero;
 use pairing::arithmetic::FieldExt;
 use pairing::bn256::Fr as Fp;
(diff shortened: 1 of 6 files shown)


CPerezz (approved):
LGTM!

Thanks for this. As we discussed will make us easier to debug the error in #268

PR #282, opened 2022-01-14
test: enable state circuit test for opcodes

lispc:
* enable state circuit test for each opcode
* fix a state circuit bug: state circuit used RWValue::to_scalar() for word,  in fact random linear combination should be used here to be consistent with evm circuit
* add vanity check for state circuit to make debugging easier: if memory_size / global_counter exceed the circuit size limit, error msg will be reported.

### DIFF
--- a/.github/proverCiScripts/deployToProver.sh
+++ b/.github/proverCiScripts/deployToProver.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+#set -e 
+#set -x
+
+prnumber=$1
+base_dir="/home/ubuntu/CI_Prover_Benches/"
+target_dir="$base_dir"PR"$1"
+
+source_dir=$2
+
+cd $source_dir && scp -r * prover:$target_dir
+
--- a/.github/proverCiScripts/execBench.sh
+++ b/.github/proverCiScripts/execBench.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -e
+#set -x
+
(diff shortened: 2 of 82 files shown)


lispc:
this PR is most related to state circuit and test, can @ChihChengLiang and @miha-stopar have a look? Thank you!

miha-stopar (approved):
LGTM! Thanks!

Only two minor things, but could be ignored: perhaps add a short description in state.rs what kind of check VANITY_CHECK enables & perhaps name BytecodeTestConfig without "Bytecode", I thought it means something else (something more directly related to the byte code) when I first saw it, but it doesn't really matter.

lispc:
totally agree with the first. For the second, I think 'TestConfig' may be too general. Is there any suggestion on its name? SimpleTxTestConfig / OpcodesTestConfig / etc? Thanks

miha-stopar:
> totally agree with the first. For the second, I think 'TestConfig' may be too general. Is there any suggestion on its name? SimpleTxTestConfig / OpcodesTestConfig / etc? Thanks

TestConfig sounds cool to me, but don't worry, you can leave it as it is, it's totally OK. Sorry for bothering you :)

ChihChengLiang (approved):
LGTM. The branch needs a rebase.

lispc:
add comments of SANITY_CHECK ( and apply the degree-fix patch to make sure CI passes)

PR #283, opened 2022-01-14
Remove biguint to field

ChihChengLiang:
### Problem
`big_uint_to_field` and `biguint_to_f` are duplicated.

Keep `biguint_to_f` since it has 32 bytes check and `big_uint_to_field` uses the concrete type `Fp`.

Also removed unused `Lane<F>`


CPerezz (approved):
LGTM! Thanks for keep simplifying the codebase!

PR #286, opened 2022-01-18
Refactor rho witness

ChihChengLiang:
### Problem

The current Rho witness uses an object that constantly mutates its state to compute the rotational base conversion, which is hard to reason and debug.

### Solution

Create a new object that computes all the rotational base conversion in one shot. It can also print out the output of 64-bit output, which is helpful to check against the plain Keccak implementation.


CPerezz (approved):
LGTM!!!

PR #291, opened 2022-01-19
Add integration test for ETH transfer

ed255:
In the integration testing setup we already had a block with an ETH transfer, but we just didn't use it as inputs for the state and evm circuit.

This is a very small PR that test (via integration) the state and evm circuit with an ETH transfer.  Currently the CircuitInputBuilder doesn't generate the BeginTx, EndTx, EndBlock circuit steps, so the circuits in this test will have mostly empty inputs; but this will soon change and I thought it was a good idea to already have an integration test for the moment we have those 3 steps generated by the CircuitInputBuilder.


CPerezz (approved):
LGTM

HAOYUatHZ (approved):

PR #297, opened 2022-01-21
Circuit for opcode calldatacopy

scroll-dev:
We verify the `CALLDATACOPY` using two gadgets: `CallDataCopyGadget` and `CopyToMemoryGadget`. 

- The `CallDataCopyGadget` checks the stack values including memory offset, data offset, and length, and then set the proper state in the `CopyToMemoryGadget`. 
- The `CopyToMemoryGadget`, which associates with an internal ExecutionState `CopyToMemory`, can span over multiple steps, depending on the amount of data to be copied. It can copy data from either transaction call data (root call) or memory (internal call) to the memory in the callee. Since `CopyToMemoryGadget` is an internal state, it only increases the rw counter while keeps the same for the rest of context, such as pc, sp, gas.

Here is a diagram showing how `CALLDATACOPY` is handled with these two gadgets. 

Untitled Diagram

Spec: https://github.com/appliedzkp/zkevm-specs/pull/111

We leave the bus-mapping implementation of `CALLDATACOPY` to next PR as it is non-trivial and needs to generate multiple steps (related to https://github.com/appliedzkp/zkevm-circuits/issues/309).


DreamWuGit (approved):
Overall LGTM!

DreamWuGit:
@han0110 @ChihChengLiang could you have a look ? we have coming PR reply on part of this !

ChihChengLiang:
We can prioritize merging this. Can we do rebase again?

ChihChengLiang (approved):
LGTM

PR #300, opened 2022-01-21
Feature/test new bech workflow

AronisAt79:


AronisAt79:
not needed, here for testing purposes

Issue #306, opened 2022-01-25
Bug: Keccak base conversion


ChihChengLiang:
### Problem

Same input, but the circuit computed one more ghost chunk.

```
outstate 0 expect len 55 b13be  [1, 1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0]
outstate 0 circuit len 56 b13be [1, 1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0]
```
Reproduction here
https://github.com/appliedzkp/zkevm-circuits/compare/base-conversion-bug?expand=1

### Solution

Investigating

ChihChengLiang:
closed by #308

PR #308, opened 2022-01-26
fix base conversion

ChihChengLiang:
### Problem

#306

The input has 64 chunks. The lookup can process n chunks. We split the input chunks into 64/n groups and perform lookups and running sums for them.
The running sum is in big-endian style `accumulation = accumulation*radix**n + coef` to save columns.
This works fine when n is a multiple of 64, say 16.
But for base 9 we lookup n=5 chunks at a time. There would be one group that has 4 chunks.
Since we use big-endian style running sums, the 4 chunks group must be on the big-endian side. The current implementation put the 4 chunks group on the little-endian side.

### Solution

Move the 4 chunks group to the big-endian side.


CPerezz (approved):
LGTM

PR #326, opened 2022-02-09
Feat/prover: add prover command line utilities

pinkiebell:
The first version includes two commands:
###### gen_params
For generating circuit commitment parameters
###### prover_cmd
that generates circuit proofs and prints them as json to stdout.

Another rpc daemon is needed for a later stage so that it becomes possible to defer generating/requesting proofs over http.

### DIFF
--- a/prover/src/bin/prover_cmd.rs
+++ b/prover/src/bin/prover_cmd.rs
@@ -82,11 +82,6 @@ async fn main() {
         let mut transcript = Blake2bWrite::<_, _, Challenge255<_>>::init(vec![]);
         create_proof(&params, &pk, &[circuit], &[], &mut transcript).expect("evm proof");
         evm_proof = transcript.finalize();
-
-        // verify
-        // let params = Setup::<Bn256>::verifier_params(&params, 0).unwrap();
-        // let mut transcript = Blake2bRead::<_, _, Challenge255<_>>::init(&evm_proof[..]);
-        // verify_proof(&params, pk.get_vk(), &[], &mut transcript).unwrap();
     }
 
     {
@@ -127,11 +122,6 @@ async fn main() {
         let mut transcript = Blake2bWrite::<_, _, Challenge255<_>>::init(vec![]);
         create_proof(&params, &pk, &[circuit], &[], &mut transcript).expect("state proof");
         state_proof = transcript.finalize();
-
-        // verify
-        // let params = Setup::<Bn256>::verifier_params(&params, 0).unwrap();
-        // let mut transcript = Blake2bRead::<_, _, Challenge255<_>>::init(&state_proof[..]);
-        // verify_proof(&params, pk.get_vk(), &[], &mut transcript).unwrap();
     }
 
     serde_json::to_writer(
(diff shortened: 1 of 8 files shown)


prover/src/bin/prover_cmd.rs
```diff
@@ -0,0 +1,135 @@
...
+        let code_hash = builder.block.txs()[0].calls()[0].code_hash;
+        let bytecode = builder
+            .code_db
+            .0
+            .get(&code_hash)
+            .expect("code_hash not found");
+        let block = block_convert(Fr::rand(), bytecode, &builder.block);
+        let circuit = TestCircuit::<Fr>::new(block, FixedTableTag::iterator().collect());
+
+        // TODO: can this be pre-generated to a file?
```
HAOYUatHZ, line 77: related: 
+ https://github.com/zcash/halo2/issues/443
+ https://github.com/zcash/halo2/issues/449

prover/src/bin/prover_cmd.rs
```diff
@@ -0,0 +1,135 @@
...
+    let builder = BuilderClient::new(geth_client).await.unwrap();
+    let builder = builder.gen_inputs(block_num).await.unwrap();
+
+    // TODO: only {evm,state}_proof are implemented right now
+    let evm_proof;
+    let state_proof;
+    {
+        // generate evm_circuit proof
+        //
+        // TODO: why blocks.txs()[0]?
```
HAOYUatHZ, line 67: related: 
+ https://github.com/scroll-tech/zkevm-circuits/pull/65#issuecomment-1010832650
+ https://github.com/scroll-tech/zkevm-circuits/issues/68

ed255 (approved):
LGTM!  Good work :)

I have two comments though:
1. I saw a few `unwrap()`, maybe you could replace them with `expect()` for a more friendly message in case of panic.
2. I noticed that `gen_params` uses arguments while `prover_cmd` uses env variables.  Why is that?

For the future I suggest taking a look at https://github.com/clap-rs/clap for argument parsing (it also supports setting arguments from env vars), so that we get help messages and nice errors on invalid arguments for free.  I think this will be useful once we have binaries that require more options.

pinkiebell:
> LGTM! Good work :)
> 
> I have two comments though:
> 
> 1. I saw a few `unwrap()`, maybe you could replace them with `expect()` for a more friendly message in case of panic.

Will do :)

> 2. I noticed that `gen_params` uses arguments while `prover_cmd` uses env variables.  Why is that?

`gen_params` has limited utility and thus will be used rarely on the cmdline - that's why its reading arguments.
the `prover_cmd` is used in test environments and all the configuration lives as env variables. This fits naturally in the prod & developer sandbox setups without having to pass any arguments. 

> For the future I suggest taking a look at https://github.com/clap-rs/clap for argument parsing (it also supports setting arguments from env vars), so that we get help messages and nice errors on invalid arguments for free. I think this will be useful once we have binaries that require more options.

Sounds good.

ed255:
@pinkiebell there seems to be a small formatting code issue: https://github.com/appliedzkp/zkevm-circuits/runs/5234903891?check_suite_focus=true

pinkiebell:
> @pinkiebell there seems to be a small formatting code issue: https://github.com/appliedzkp/zkevm-circuits/runs/5234903891?check_suite_focus=true

sry, should run rustfmt in the root folder next time 😅

Lizy87 (approved):

PR #328, opened 2022-02-10
Circuit for opcode SELFBALANCE

scroll-dev:
Implement the SelfbalanceGadget for opcode `SELFBALANCE`.

Specs: https://github.com/appliedzkp/zkevm-specs/pull/113

Will update the bus mapping for SELFBALANCE once https://github.com/appliedzkp/zkevm-circuits/pull/292 is merged in upstream.


ChihChengLiang:
@scroll-dev Mason Liang Can we do a rebase for this?

bus-mapping/src/evm/opcodes.rs
```diff
@@ -116,7 +117,7 @@ fn fn_gen_associated_ops(opcode_id: &OpcodeId) -> FnGenAssociatedOps {
         // OpcodeId::DIFFICULTY => {},
         // OpcodeId::GASLIMIT => {},
         // OpcodeId::CHAINID => {},
-        // OpcodeId::SELFBALANCE => {},
+        OpcodeId::SELFBALANCE => StackOnlyOpcode::<0>::gen_associated_ops,
```
ed255, line 120: Shouldn't there be an CallContextOp and AccountOp READ in this opcode?  If you plan on adding those once #292 is merged, maybe you could set this PR as draft until it's ready?

ed255 (changes requested):

zkevm-circuits/src/evm_circuit/execution/selfbalance.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let callee_address = cb.query_cell();
+        cb.call_context_lookup(
+            None,
+            CallContextFieldTag::CalleeAddress,
+            callee_address.expr(),
+        );
+
+        let self_balance = cb.query_rlc();
```
ed255, line 38: Considering that the `self_balance` value is only used in the RLC encoded form in this gadget, maybe we can avoid using `Word` (which requires 32 cells), and just use a `Cell` for `self_balance`?
I think the current implementation is correct nevertheless.

z2trillion: updated the PR to use `Cell` instead.

ed255 (approved):
LGTM! Thanks, good work!

z2trillion:
@ed255, I've rebased and added the bus mapping.

ed255 (approved):

ed255:
@z2trillion There seems to be a linting error coming from clippy.

z2trillion:
@ed255 it should be fixed now.

PR #329, opened 2022-02-10
Gadget for opcode calldatasize

scroll-dev:
Spec: https://github.com/appliedzkp/zkevm-specs/pull/96

### DIFF
--- a/bus-mapping/src/evm/opcodes.rs
+++ b/bus-mapping/src/evm/opcodes.rs
@@ -4,6 +4,7 @@ use crate::Error;
 use core::fmt::Debug;
 use eth_types::GethExecStep;
 
+mod calldatasize;
 mod caller;
 mod callvalue;
 mod coinbase;
@@ -27,6 +28,7 @@ use crate::evm::OpcodeId;
 use log::warn;
 
 use self::push::Push;
+use calldatasize::Calldatasize;
 use caller::Caller;
 use callvalue::Callvalue;
 use dup::Dup;
@@ -104,7 +106,7 @@ fn fn_gen_associated_ops(opcode_id: &OpcodeId) -> FnGenAssociatedOps {
         OpcodeId::CALLER => Caller::gen_associated_ops,
         OpcodeId::CALLVALUE => Callvalue::gen_associated_ops,
         // OpcodeId::CALLDATALOAD => {},
-        // OpcodeId::CALLDATASIZE => {},
+        OpcodeId::CALLDATASIZE => Calldatasize::gen_associated_ops,
         // OpcodeId::CALLDATACOPY => {},
         // OpcodeId::CODESIZE => {},
         // OpcodeId::CODECOPY => {},
--- a/zkevm-circuits/src/evm_circuit/execution/calldatasize.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/calldatasize.rs
@@ -8,7 +8,7 @@ use crate::{
         util::{
             common_gadget::SameContextGadget,
             constraint_builder::{ConstraintBuilder, StepStateTransition, Transition},
-            Cell, Word,
+            Word,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
@@ -20,7 +20,7 @@ use super::ExecutionGadget;
 #[derive(Clone, Debug)]
 pub(crate) struct CallDataSizeGadget<F> {
     same_context: SameContextGadget<F>,
-    call_data_size: Cell<F>,
+    call_data_size: Word<F>,
 }
 
 impl<F: FieldExt> ExecutionGadget<F> for CallDataSizeGadget<F> {
@@ -32,7 +32,13 @@ impl<F: FieldExt> ExecutionGadget<F> for CallDataSizeGadget<F> {
         let opcode = cb.query_cell();
 
         // Add lookup constraint in the call context for the calldatasize field.
-        let call_data_size = cb.call_context(None, CallContextFieldTag::CallDataLength);
+        let call_data_size = cb.query_rlc();
+        cb.call_context_lookup(
+            false.expr(),
+            None,
+            CallContextFieldTag::CallDataLength,
+            call_data_size.expr(),
+        );
 
         // The calldatasize should be pushed to the top of the stack.
         cb.stack_push(call_data_size.expr());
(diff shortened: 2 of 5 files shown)


ChihChengLiang:
@roynalnaruto Can we rebase this?

bus-mapping/src/evm/opcodes.rs
```diff
@@ -99,7 +100,7 @@ fn fn_gen_associated_ops(opcode_id: &OpcodeId) -> FnGenAssociatedOps {
         // OpcodeId::CALLER => {},
         // OpcodeId::CALLVALUE => {},
         // OpcodeId::CALLDATALOAD => {},
-        // OpcodeId::CALLDATASIZE => {},
+        OpcodeId::CALLDATASIZE => StackOnlyOpcode::<0>::gen_associated_ops,
```
ed255, line 103: I think this can't be just a `StackOnlyOpcode` because it does a `CallContextOp` READ to the `rw_table`, similar to this (from https://github.com/appliedzkp/zkevm-circuits/pull/287): https://github.com/appliedzkp/zkevm-circuits/blob/0e7c1791f849fcb76d6f3367298313e0e5854c4f/bus-mapping/src/evm/opcodes/caller.rs#L21-L26

The `CallContextOp` is pending to be introduced in #292

ed255 (changes requested):

bus-mapping/src/evm/opcodes.rs
roynalnaruto: Yes that's true, so let's wait until #292 is merged. Thanks!

zkevm-circuits/src/evm_circuit/execution/calldatasize.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        // Add lookup constraint in the call context for the calldatasize field.
+        cb.call_context_lookup(
+            None,
+            CallContextFieldTag::CallDataLength,
+            call_data_size.expr(),
+        );
+
+        // The calldatasize should be pushed to the top of the stack.
+        cb.stack_push(call_data_size.expr());
```
ed255, line 44: I believe that the CallDataLength from the CallContext is the actual value, but in the stack it must be RLC encoded:
https://github.com/appliedzkp/zkevm-specs/blob/feb9d0e9a42cd6ef3a62ae0d3e87cd9224100bb6/src/zkevm_specs/evm/execution/calldatasize.py#L13-L22

ed255 (changes requested):

bus-mapping/src/evm/opcodes.rs
ed255: #292 has been merged :)

roynalnaruto: Thanks, will update my PR tomorrow 👍

zkevm-circuits/src/evm_circuit/execution/calldatasize.rs
roynalnaruto: @ed255 I've made changes to resolve the above comments and discussions.

ed255: I think this implementation is still different than the spec.  Currently the value used for the stack WRITE is not RLC encoded.

Take a look for example at the COINBASE implementation: https://github.com/appliedzkp/zkevm-circuits/blob/b1e368ec7812ca7db9a016b780d586e247f634a3/zkevm-circuits/src/evm_circuit/execution/coinbase.rs#L32  The coinbase value is an ethereum address so it fits in the field.  In the RwTable.CallContext it's raw, and then in the RwTable.Stack it's RLC encoded.

zkevm-circuits/src/evm_circuit/execution/calldatasize.rs
```diff
@@ -0,0 +1,184 @@
...
+            }],
+        );
+        rw_map.insert(
+            RwTableTag::Stack,
+            vec![Rw::Stack {
+                rw_counter: 10,
+                is_write: true,
+                call_id,
+                stack_pointer: 1023,
+                value: Word::from(call_data_size),
```
ed255, line 126: Following the spec, this would be RLC encoded. See https://github.com/appliedzkp/zkevm-specs/blob/feb9d0e9a42cd6ef3a62ae0d3e87cd9224100bb6/tests/evm/test_calldatasize.py#L39

roynalnaruto: @ed255 I've used the `query_rlc` for calldatasize in the latest commit. Although this `value` field in the `Rw::Stack` is later converted into RLC-encoded value here:
https://github.com/appliedzkp/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/witness.rs#L631

ed255: Oh ok! Got it!  Sorry for the confusion.

roynalnaruto: Although the points you made earlier (regarding the fact that `cb.stack_pop` for the `call_data_size` value) were absolutely correct, and I've made the necessary changes for that.

ed255: In the current state of this PR the opcode expects that in the RwTable.Stack and RwTable.CallContext.CallDataLength the values are RLC encoded.  This is because you're using `call_data_size.expr()`.  The correct implementation should be that in RwTable.Stack it's RLC encoded, but in RwTable.CallContext.CallDataLength it's the actual value.  The actual value can be obtained by using `from_bytes::expr(&call_data_size.cells)`.  Check the coinbase opcode implementation for reference.

Now, In the current state of this PR the tests are passing, so I was trying to figure out why.  And it turns out that for values between `0` and `255`, `RLC(x) == x`, which is unfortunate, because it means that when we use small values in tests we can't catch bugs when comparing RLC encoded to non-RLC encoded values.  This is because the RLC encoding does `byte0 + byte1 * rnd + byte2 * rnd**2 + ...`.  So if only byte0 is filled, the rnd is not used. 

To validate that the current implementation is not correct, please add a test case for a value of `300`, and you'll see that it fails.

roynalnaruto: My bad, I think I should have also added a test case with a random calldatasize with upper bound on bit length. And this bug would have been caught. Thanks so much :) Will fix it today.

roynalnaruto: @ed255 The issue should now be resolved. I've also added another constant `N_BYTES_CALLDATASIZE` which is the same as `u64` (or 8 bytes), since that's what [`go-ethereum` limits to](https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L294-L297) as well.

ed255 (approved):
LGTM! Great work :)

ed255:
@roynalnaruto this PR needs one last rebase with main and it will be ready to be merged :)

roynalnaruto:
> @roynalnaruto this PR needs one last rebase with main and it will be ready to be merged :)

Done! 👍

PR #333, opened 2022-02-16
feat: add error handling in `impl From<&bus_mapping::circuit_input_builder::ExecStep> for ExecutionState`

scroll-dev:


zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -557,11 +558,46 @@ impl Rw {
...
+        match error {
+            ExecError::ExecutionReverted => ExecutionState::ErrorReverted,
+            ExecError::InvalidOpcode => ExecutionState::ErrorInvalidOpcode,
+            ExecError::StackOverflow => ExecutionState::ErrorStackOverflow,
+            ExecError::StackUnderflow => ExecutionState::ErrorStackUnderflow,
+            ExecError::WriteProtection => ExecutionState::ErrorWriteProtection,
+            ExecError::Depth => ExecutionState::ErrorDepth,
+            ExecError::InsufficientBalance => ExecutionState::ErrorInsufficientBalance,
+            ExecError::ContractAddressCollision => ExecutionState::ErrorContractAddressCollision,
+            ExecError::InvalidCode => ExecutionState::ErrorInvalidCreationCode,
```
scroll-dev, line 572: in fact not sure whether `ExecError::InvalidCode` is corresonding to `ExecutionState::ErrorInvalidCreationCode`.

But would like to fix if this line wrong

ed255: I think that is correct.
- `InvalidOpcode`: happens when the execution encounters an opcode that is not valid
- `InvalidCode`: happens when a creation call returns a code that is invalid (starts with `0xef`)

Maybe we can rename `ExecError::InvalidCode` to `ExecError::InvalidCreationCode` to be consistent.

Also we can rename `ExecError::ExecutionReverted` to `ExecError::Reverted` for consistency.

ed255 (approved):
LGTM!

PR #379, opened 2022-03-07
CI: Skip dependencies for doc checks

ed255:
Resolve https://github.com/appliedzkp/zkevm-circuits/issues/377
Resolve https://github.com/appliedzkp/zkevm-circuits/issues/378

With this change we just skip building the docs for our dependencies in the CI (which is not meaningful to us), and this should resolve the race that makes this test sometimes fail when building the `bitvec` docs for different versions in the same folder.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,7 +183,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: doc
-          args: --all --document-private-items
+          args: --no-deps --all --document-private-items
 
   fmt:
     if: github.event.pull_request.draft == false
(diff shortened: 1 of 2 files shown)


CPerezz (approved):
As said privately, I'm not sure if `--no-deps` can cause errors when in the docs we reference structs from our dependencies.

In any case, let's try this solution for now and we can always revert.
The ideal solution will arrive once this is fixed in cargo anyway.

ed255:
> As said privately, I'm not sure if `--no-deps` can cause errors when in the docs we reference structs from our dependencies.
> 
> In any case, let's try this solution for now and we can always revert. The ideal solution will arrive once this is fixed in cargo anyway.

Yeah, nevertheless I don't know when will this be fixed in cargo, it make take a looong time!

I'm going to merge this now, thanks for the review!

PR #380, opened 2022-03-07
state circuit update part1: change RwTable according to latest spec; finish stack/storage constraints

lispc:
resolving issue: https://github.com/appliedzkp/zkevm-circuits/issues/356

The reason why I submit this WIP PR now is that RwTable is updated according to the latest spec here. Another key column is added, so the overall keys become: tag,key1,key2,key3,key4.

If we can merge this PR early, we may encounter less conflicts on RwTable. So i think it is better to split the whole state circuit refactor PR into several smaller ones, merge them one by one periodically


lispc:
@ed255 @miha-stopar

miha-stopar (approved):
I like it very much! Much cleaner than my implementation.

Only one question - the constraints for the address monotonicity are to be added, right? (referring to the disabled test)

lispc:
> I like it very much! Much cleaner than my implementation.
> 
> Only one question - the constraints for the address monotonicity are to be added, right? (referring to the disabled test)

yes. Old implementation uses lookup, but we need finally use range check. This modification needs some work. So I disable it temporarily

miha-stopar:
> > I like it very much! Much cleaner than my implementation.
> > Only one question - the constraints for the address monotonicity are to be added, right? (referring to the disabled test)
> 
> yes. Old implementation uses lookup, but we need finally use range check. This modification needs some work. So I disable it temporarily

Cool, thanks!

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -98,453 +75,327 @@ pub struct Config<
...
+        // 2. stack_ptr in range
+        meta.lookup_any("Stack address in allowed range", |meta| {
+            let q_stack = q_stack(meta);
+            let address_cur = meta.query_advice(address, Rotation::cur());
+            let stack_address_table_zero =
+                meta.query_fixed(stack_address_table_zero, Rotation::cur());
 
-            let padding = meta.query_advice(padding, Rotation::cur());
-            let is_not_padding = one.clone() - padding;
+            vec![(q_stack * address_cur, stack_address_table_zero)]
```
ed255, line 314: Same question as before: Usage of s_enable in the lookup?

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -605,379 +456,119 @@ impl<
...
-                let storage_mappings = self.assign_storage_ops(
-                    &mut region,
-                    randomness,
-                    storage_ops.clone(),
-                    &address_diff_is_zero_chip,
-                    &storage_key_diff_is_zero_chip,
-                    offset,
-                );
-                bus_mappings.extend(storage_mappings.unwrap());
+                let mut offset = 1;
```
ed255, line 472: Could you explain why this starts with `offset = 1`?  Where is `offset = 0` set? (which should use the `START_TAG`).
In case you're planning on doing this in a future PR, would you add a small note with a TODO stating that please?

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -605,379 +456,119 @@ impl<
...
+            }
+        }
+        region.assign_fixed(|| "enable row", self.s_enable, offset, || Ok(F::one()))?;
+        region.assign_advice(|| "rw counter", self.rw_counter, offset, || Ok(rw_counter))?;
+        region.assign_advice(|| "value", self.value, offset, || Ok(value))?;
+        region.assign_advice(|| "is_write", self.is_write, offset, || Ok(is_write))?;
+
+        for (i, diff_is_zero_chip) in diff_is_zero_chips.iter().enumerate() {
+            let (value, diff) = match i {
+                // FIXME: find a better way here
```
ed255, line 554: I think this is good enough :)
The only think that would come to my mind to improve this would be to change `RwRow` to have a field `keys: [F, 5]`.

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -605,379 +456,119 @@ impl<
...
+                    "memory address out of range {:?} > {}",
+                    address, MEMORY_ADDRESS_MAX
+                );
+            }
+        }
+        region.assign_fixed(|| "enable row", self.s_enable, offset, || Ok(F::one()))?;
+        region.assign_advice(|| "rw counter", self.rw_counter, offset, || Ok(rw_counter))?;
+        region.assign_advice(|| "value", self.value, offset, || Ok(value))?;
+        region.assign_advice(|| "is_write", self.is_write, offset, || Ok(is_write))?;
+
```
ed255, line 551: I noticed that the following values are not assigned:
- `value_prev`
- `aux1`
- `aux2`

Why is that?

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -98,453 +75,327 @@ pub struct Config<
...
-            //          value == prev_value
-            //      else:
-            //          value == 0
-            let q_memory_not_first = q_memory_not_first(meta);
-            let address_diff = {
-                let address_prev = meta.query_advice(address, Rotation::prev());
-                let address_cur = meta.query_advice(address, Rotation::cur());
-                address_cur - address_prev
-            };
+        ///////////////////////// General constraints /////////////////////////////////
```
ed255, line 185: I noticed that the constraints `0, 1, 2, 4` are missing.  I understand that you'll do them in a future PR.  But could you add a small TODO stating that there are missing general constraints to be implemented?

ed255 (commented):
This looks very good!  Thanks for the nice work :)

I've added some comments and questions.  Maybe you're planning to address them in the next PR, that's totally OK by me!  But if that's the case I would just ask if you can add some small comments stating the missing tasks with TODO.

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -98,453 +75,327 @@ pub struct Config<
...
+        // 2. mem_addr in range
+        // TODO: rewrite this using range check gates instead of lookup
         meta.lookup_any("Memory address in allowed range", |meta| {
-            let q_memory = q_memory_first(meta) + q_memory_not_first(meta);
+            let q_memory = q_memory(meta);
             let address_cur = meta.query_advice(address, Rotation::cur());
             let memory_address_table_zero =
                 meta.query_fixed(memory_address_table_zero, Rotation::cur());
 
             vec![(q_memory * address_cur, memory_address_table_zero)]
```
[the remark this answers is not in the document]
lispc: since for mem addr 0 is a valid value, so memory_address_table_zero will contain 0. So even for s_enable == false rows, the  `q_memory * address_cur` will be 0(default value) x 0(default value), it will be allowed in the lookup table.  I will add a comment on this.

zkevm-circuits/src/state_circuit/state.rs
```diff
@@ -98,453 +75,327 @@ pub struct Config<
...
+        // 3. value is a byte
+        // Memory value is in the allowed range.
         meta.lookup_any("Memory value in allowed range", |meta| {
-            let q_memory_not_first = q_memory_not_first(meta);
+            let q_memory = q_memory(meta);
             let value = meta.query_advice(value, Rotation::cur());
             let memory_value_table = meta.query_fixed(memory_value_table, Rotation::cur());
 
-            vec![(q_memory_not_first * value, memory_value_table)]
+            vec![(q_memory * value, memory_value_table)]
```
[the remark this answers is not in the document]
lispc: same with previous

zkevm-circuits/src/state_circuit/state.rs
lispc: same

zkevm-circuits/src/state_circuit/state.rs
lispc: yes, I will add "aux1" and "aux2"!   While I think value_prev has been removed already in the spec?

lispc:
@ed255 Hi Edu, I addressed the comments above~ You can have a look. Thanks!

zkevm-circuits/src/state_circuit/state.rs
ed255: ah that's right, `value_prev` is no longer in the state circuit spec.

ed255 (approved):
LGTM!  Thanks for addressing my comments :)

PR #381, opened 2022-03-07
Circuit for NUMBER opcode

bchyl:
1. as with timestamp treatig the number u256 type as u64 tmp.
2. next step，for BlockContext word op such as difficulty should use `RandomLinearCombination<F, 32>` instead of scalar for decompress in bus mapping into v0..v31
4. and maybe also reuse code for BlockContext ops.


zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -65,10 +65,7 @@ impl BlockContext {
                 [
                     F::from(BlockContextFieldTag::Number as u64),
                     F::zero(),
-                    RandomLinearCombination::random_linear_combine(
-                        self.number.to_le_bytes(),
-                        randomness,
-                    ),
+                    self.number.to_scalar().unwrap(),
```
ed255, line 68: In the spec, the Block table, in the Number tag contains the RLC enoded value of the block number: https://github.com/appliedzkp/zkevm-specs/blob/657d0b034932ea8fa6515c3ec1f9d1e0055f43b0/src/zkevm_specs/evm/typing.py#L63

I think it's better to assign the value without RLC encoding, like you do in this PR (using the assumption that the number will always fit in the field).  If I remember correctly, Han mentioned that the yellow paper doesn't specify a bound on the block number, but I think 64 bits is a sensible bound, and there have been attempts to make these bounds official: https://eips.ethereum.org/EIPS/eip-1985 (we discussed that we may adhere to this EIP in the zkEVM implementation).

To make the implementation match the spec, I think the spec should be updated so that the block number in the Block table is not RLC encoded.

ed255 (commented):
Everything looks good!  Great work!

Nevertheless this implementation is different than the current spec as I have written in a comment.  I think the spec should be updated.

zkevm-circuits/src/evm_circuit/witness.rs
bchyl: Thank you very much for explaining the detailed RLC implementation design background.

bchyl: Thank you very much for explaining the detailed RLC implementation design background.

bchyl:
> Everything looks good! Great work!
> 
> Nevertheless this implementation is different than the current spec as I have written in a comment. I think the spec should be updated.

updated at [specs pr-140](https://github.com/appliedzkp/zkevm-specs/pull/140). pls help to review, 3gs.

CPerezz (approved):
LGTM once https://github.com/appliedzkp/zkevm-specs/pull/140 is merged!

Also, probably worth to check the nit on the tests I highlighted :)

zkevm-circuits/src/evm_circuit/execution/number.rs
```diff
@@ -0,0 +1,100 @@
...
+            #[start]
+            NUMBER
+            STOP
+        };
+        assert_eq!(run_test_circuits(bytecode), Ok(()));
+    }
+    #[test]
+    fn number_gadget_test() {
+        test_ok();
+    }
```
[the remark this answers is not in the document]
bchyl: nice catch

ed255 (approved):
LGTM!  Thanks for working on this! :)
Once the github actions pass we can merge this.

PR #382, opened 2022-03-09
[WIP]  integrate ethereum/tests StateTests

adria0:
Integrate the "official" EVM tests located in https://github.com/ethereum/tests

Since we are running our own geth executor to generate the traces, we are using the `/src` folder and not the [retesteth](https://github.com/ethereum/retesteth) _filled tests_  (located in `/BlockchainTests` and `/GeneralStateTests`).

The basic roadmap for this PR is

- [x] parse a basic .yaml tests
- [x] test StateTests, check results
- [x] compile [LLL](https://lll-docs.readthedocs.io/en/latest/lll_introduction.html) via docker
- [ ] pass the [first basic existing test](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/VMTests/vmArithmeticTest/addFiller.yml) we also need two things here:
  - [ ] `SSTORE` implemented  (scroll-tech's [WIP](https://github.com/scroll-tech/zkevm-circuits/pull/107)
  - [ ] `CALL` implemented 
- [ ] add more tests about the current implemented opcodes
- [ ] compile [YUL](https://docs.soliditylang.org/en/v0.8.12/yul.html) via docker
- [ ] use the [json](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stExample/add11Filler.json) spec (is not clear that is going to be needed, since EF tests team is moving all tests from .json to .yaml)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "evm-testvectors/tests"]
+	path = evm-testvectors/tests
+	url = https://github.com/ethereum/tests
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,9 +50,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.55"
+version = "1.0.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "159bb86af3a200e19a068f4224eae4c8bb2d0fa054c7e5d1cacd5cef95e684cd"
+checksum = "4361135be9122e0870de935d7c439aef945b9f9ddd4199a553b5270b49c82a27"
 
 [[package]]
 name = "ark-std"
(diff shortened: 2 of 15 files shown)


ed255:
>         * which ones the geth trace is too big ; I had to modify `geth-utils/lib/lib.go` to skip very big traces (geth crashed)
@adria0 do you know if geth crashes from an out of memory situation?  scroll-tech had experimented with generating traces of big mainnet blocks and geth crashed when the trace includes the memory dump at each step.  They made a PR here https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/614 where the memory dump is disabled from the trace and the memory snapshot is reconstructed in the circuit input builder.  Hopefully that should increase the number of tests we can run :)

ChihChengLiang (commented):
Do we intend to check in `code.cache` and `result.cache`? Or should we ignore them?

ChihChengLiang:
@adria0 Yeah, sure. We can have a meeting when you and Edu are back.

ed255:
> Those files are just kept to speed up testing process and to check diffs between executions. Maybe we could make a quick meeting with you and @ed255 about how we can improve and make this tests better before a review, starting september? Happy to share some ideas I have :)

Sounds good to me!

PR #391, opened 2022-03-16
CODECOPY opcode

roynalnaruto:
Specs PR: https://github.com/appliedzkp/zkevm-specs/pull/148

TODOs:
- [x] update bytecode circuit to include a `Length` row per bytecode (as per https://github.com/appliedzkp/zkevm-specs/pull/148)
- [x] refactor `CodeSize` to be fetched from the bytecode circuit
- [x] update bytecode table in specs (https://github.com/appliedzkp/zkevm-specs/pull/176)

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -11,7 +11,9 @@ use crate::state_db::{self, CodeDB, StateDB};
 use crate::Error;
 use core::fmt::Debug;
 use eth_types::evm_types::{Gas, GasCost, MemoryAddress, OpcodeId, ProgramCounter, StackAddress};
-use eth_types::{self, Address, GethExecStep, GethExecTrace, Hash, ToAddress, ToBigEndian, Word};
+use eth_types::{
+    self, Address, GethExecStep, GethExecTrace, Hash, ToAddress, ToBigEndian, Word, U256,
+};
 use ethers_core::utils::{get_contract_address, get_create2_address};
 use std::collections::{hash_map::Entry, BTreeMap, HashMap, HashSet};
 
@@ -165,8 +167,8 @@ pub enum StepAuxiliaryData {
         bytes_left: u64,
         /// Source end address
         src_addr_end: u64,
-        /// Bytecode to be copied
-        code: eth_types::Bytecode,
+        /// Hash of the bytecode to be copied
+        code_source: U256,
     },
 }
 
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -141,35 +141,43 @@ impl ExecState {
     }
 }
 
+/// Auxiliary data for CopyToMemory internal state.
+#[derive(Clone, Debug)]
+pub struct CopyToMemoryAuxData {
+    /// Source start address
+    pub src_addr: u64,
+    /// Destination address
+    pub dst_addr: u64,
+    /// Bytes left
+    pub bytes_left: u64,
+    /// Source end address
+    pub src_addr_end: u64,
+    /// Indicate if copy from transaction call data
+    pub from_tx: bool,
+}
+
+/// Auxiliary data for CopyCodeToMemory internal state.
+#[derive(Clone, Debug)]
+pub struct CopyCodeToMemoryAuxData {
+    /// Source start address
+    pub src_addr: u64,
+    /// Destination address
+    pub dst_addr: u64,
+    /// Bytes left
+    pub bytes_left: u64,
+    /// Source end address
+    pub src_addr_end: u64,
+    /// Hash of the bytecode to be copied
+    pub code_source: U256,
+}
+
 /// Auxiliary data of Execution step
 #[derive(Clone, Debug)]
 pub enum StepAuxiliaryData {
     /// Auxiliary data of Copy To Memory
-    CopyToMemory {
-        /// Source start address
-        src_addr: u64,
-        /// Destination address
-        dst_addr: u64,
-        /// Bytes left
-        bytes_left: u64,
-        /// Source end address
-        src_addr_end: u64,
-        /// Indicate if copy from transaction call data
-        from_tx: bool,
-    },
+    CopyToMemory(CopyToMemoryAuxData),
     /// Auxiliary data of Copy Code To Memory
-    CopyCodeToMemory {
-        /// Source start address
-        src_addr: u64,
-        /// Destination address
-        dst_addr: u64,
-        /// Bytes left
-        bytes_left: u64,
-        /// Source end address
-        src_addr_end: u64,
-        /// Hash of the bytecode to be copied
-        code_source: U256,
-    },
+    CopyCodeToMemory(CopyCodeToMemoryAuxData),
 }
 
 /// An execution step of the EVM.
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -142,7 +142,7 @@ impl ExecState {
 }
 
 /// Auxiliary data for CopyToMemory internal state.
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
 pub struct CopyToMemoryAuxData {
     /// Source start address
     pub src_addr: u64,
@@ -157,7 +157,7 @@ pub struct CopyToMemoryAuxData {
 }
 
 /// Auxiliary data for CopyCodeToMemory internal state.
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
 pub struct CopyCodeToMemoryAuxData {
     /// Source start address
     pub src_addr: u64,
@@ -172,7 +172,7 @@ pub struct CopyCodeToMemoryAuxData {
 }
 
 /// Auxiliary data of Execution step
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug)]
 pub enum StepAuxiliaryData {
     /// Auxiliary data of Copy To Memory
     CopyToMemory(CopyToMemoryAuxData),
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -155,6 +155,19 @@ pub enum StepAuxiliaryData {
         /// Indicate if copy from transaction call data
         from_tx: bool,
     },
+    /// Auxiliary data of Copy Code To Memory
+    CopyCodeToMemory {
+        /// Source start address
+        src_addr: u64,
+        /// Destination address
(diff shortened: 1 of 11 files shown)


roynalnaruto:
@han0110 In the specs lookup implementation [here](https://github.com/scroll-tech/zkevm-specs/blob/feat/codecopy/src/zkevm_specs/evm/table.py#L410-L430), we check if there's one and only one match and return it. But it does allow optional fields.

How can this be done in the Rust impl?

As I see it, [this](https://github.com/scroll-tech/zkevm-circuits/blob/feat/codecopy/zkevm-circuits/src/evm_circuit/table.rs#L325-L339) would have to be changed to:
```rust
            Self::Bytecode {
                hash,
                tag,
                index,
                opt_is_code,
                value,
            } => {
                if let Some(is_code) = opt_is_code {
                    vec![hash.clone(), tag.clone(), index.clone(), value.clone(), is_code.clone()]
                } else {
                    vec![hash.clone(), tag.clone(), index.clone(), value.clone()]
                }
            }
```

Also the layout of bytecode table will have to be modified so that `is_code` column follows the `value` column. Wdyt?

han0110:
In spec we try not to annoying developer to provide all of witness to pass a verification, so lookup is made to be like a normal map lookup, then we can use partial value to get the full row.

But in circuit, we have to provide all, so current approach makes sense me. I think we can stick to this now, because there is another idea that can help to make all of these more sense. (see here https://hackmd.io/Pmd8xoZWRVmko8LkRNhABQ)

roynalnaruto:
Bus-mapping PR added here: https://github.com/scroll-tech/zkevm-circuits/pull/123

roynalnaruto:
@han0110 @CPerezz @ed255 Would be great if this PR is reviewed, since the specs for `CODECOPY` have been merged already.

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -324,7 +325,7 @@ impl Bytecode {
     pub fn table_assignments<'a, F: FieldExt>(
         &'a self,
         randomness: F,
-    ) -> impl Iterator<Item = [F; 4]> + '_ {
+    ) -> impl Iterator<Item = [F; 5]> + '_ {
```
icemelon, line 328: I think you can change the return type and simplify the logic

```suggestion
    ) -> Vec<[F; 5]> {
```

zkevm-circuits/src/evm_circuit/table.rs
```diff
@@ -174,6 +174,13 @@ pub enum AccountFieldTag {
     Nonce = 1,
     Balance,
     CodeHash,
+    CodeSize,
```
icemelon, line 177: Is this tag used anywhere?

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -362,8 +368,10 @@ impl<F: Field> ExecutionConfig<F> {
             calldataload_gadget: configure_gadget!(),
             calldatasize_gadget: configure_gadget!(),
             caller_gadget: configure_gadget!(),
+            codecopy_gadget: configure_gadget!(),
             coinbase_gadget: configure_gadget!(),
             comparator_gadget: configure_gadget!(),
+            copy_code_to_memory_gadget: configure_gadget!(),
```
icemelon, line 374: move this to the internal states section

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -92,6 +102,28 @@ impl<F: Field> Config<F> {
             push_rindex_inv,
         );
 
+        // Does the current row have bytecode field tag == Length?
+        let is_row_tag_length = |meta: &mut VirtualCells<F>| {
+            let row_tag = meta.query_advice(tag, Rotation::cur());
+            row_tag.expr() * (BytecodeFieldTag::Byte.expr() - row_tag.expr())
```
icemelon, line 108: Here you limit the `tag` to be either 0 or `BytecodeFieldTag::Byte`. But based on the `BytecodeFieldTag` in the `table.rs`, the tag value should be either 1 or 2, right?

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -92,6 +102,28 @@ impl<F: Field> Config<F> {
...
+        // Does the current row have bytecode field tag == Byte?
+        let is_row_tag_byte = |meta: &mut VirtualCells<F>| {
+            let row_tag = meta.query_advice(tag, Rotation::cur());
+            row_tag.expr()
+                * (row_tag.expr() - BytecodeFieldTag::Length.expr())
+                * F::from(2).invert().unwrap()
+        };
+
+        // For a row tagged Length, is the length (value) zero?
+        let length_is_zero = IsZeroChip::configure(
```
icemelon, line 120: could the `length` of code be 0?

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -92,6 +102,28 @@ impl<F: Field> Config<F> {
             push_rindex_inv,
         );
 
+        // Does the current row have bytecode field tag == Length?
+        let is_row_tag_length = |meta: &mut VirtualCells<F>| {
```
icemelon, line 106: one potential optimization is that we just use 0/1 to serve as a selector for `length` or `byte`. So we can reduce the degree from 2 to 1

zkevm-circuits/src/evm_circuit/table.rs
roynalnaruto: This must be removed 👍

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
roynalnaruto: Yes,  based on the discussion in https://github.com/appliedzkp/zkevm-specs/issues/102:

An empty code will be represented as:
```
(EMPTY_HASH, TagLength, 0, 0, 0)
```

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
roynalnaruto: In that case, `padding` must be changed to `2` since it's used in the `tag` as well. At the moment `padding` is represented as `0`, but this point makes sense and I'll make the update.

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -155,6 +155,19 @@ pub enum StepAuxiliaryData {
...
+        /// Source start address
+        src_addr: u64,
+        /// Destination address
+        dst_addr: u64,
+        /// Bytes left
+        bytes_left: u64,
+        /// Source end address
+        src_addr_end: u64,
+        /// Bytecode to be copied
+        code: eth_types::Bytecode,
```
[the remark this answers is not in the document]
icemelon: sounds good to me.

zkevm-circuits/src/evm_circuit/execution/copy_code_to_memory.rs
```diff
@@ -0,0 +1,475 @@
...
+        let randomness = Fr::rand();
+        let call_id = 1;
+        let mut rws = RwMap::default();
+        let mut rw_counter = 1;
+        let mut steps = Vec::new();
+        let memory_size = (dst_addr + length as u64 + 31) / 32 * 32;
+
+        // generate random bytecode longer than `src_addr_end`
+        let code = bytecode! {
+            #[start]
```
han0110, line 384:
```suggestion
```

han0110 (commented):
Only got some nitpicks, others look good to me.

ChihChengLiang (approved):
LGTM modulo other review comments.

han0110 (approved):
LGTM! It will be nice to see its part of `bus-mapping` implemented next (it's fine to me to implement it in further PR), then #419 doesn't need to worry about that, thanks!

roynalnaruto:
> It will be nice to see its part of bus-mapping implemented next

@han0110 It's already in progress here: https://github.com/scroll-tech/zkevm-circuits/pull/123.

Once I'm done with the internal call cases for calldataload and calldatacopy, I will wrap this up.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
icemelon: ah, I see. In that case, I think either way looks good to me.

zkevm-circuits/src/evm_circuit/execution/codecopy.rs
```diff
@@ -0,0 +1,422 @@
...
+            rws: rws_map,
+            bytecodes: vec![code],
+            ..Default::default()
+        };
+        assert_eq!(run_test_circuit_incomplete_fixed_table(block), Ok(()));
+    }
+
+    #[test]
+    fn codecopy_gadget_single_step() {
+        test_ok(0x00, 0x00, 54);
```
icemelon, line 410: add another test case with non-zero code offset here?

roynalnaruto:
> In that case, I think either way looks good to me.

I've made the change where `Length = 0`, `Byte = 1` and `Padding = 2`

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -331,56 +332,41 @@ impl Bytecode {
...
-                let byte = self.bytes[self.idx];
+    pub fn table_assignments<F: FieldExt>(&self, randomness: F) -> Vec<[F; 5]> {
+        let n = 1 + self.bytes.len();
+        let mut rows = Vec::with_capacity(n);
+        let hash =
+            RandomLinearCombination::random_linear_combine(self.hash.to_le_bytes(), randomness);
+        let mut push_data_left = 0;
+        for idx in 0..n {
+            if idx == 0 {
+                rows.push([
```
icemelon, line 343: move this part out of the for loop and you don't need to shift the index like `let i = idx - 1;`.

zkevm-circuits/src/evm_circuit/execution/copy_code_to_memory.rs
```diff
@@ -0,0 +1,475 @@
...
+        let CopyCodeToMemoryAuxData {
+            src_addr,
+            dst_addr,
+            bytes_left,
+            src_addr_end,
+            code_source,
+        } = match step
+            .aux_data
+            .as_ref()
+            .expect("could not find aux_data for COPYCODETOMEMORY")
```
icemelon, line 190: ditto

icemelon (approved):
LGTM!

icemelon:
Thanks for the contribution @roynalnaruto

PR #398, opened 2022-03-21
MPT

### LINKED ISSUE #1416: Merge MPT to main

Creating this issue to track #398.

### Additional context

_No response_

miha-stopar:
This PR aims to provide the first version of Merkle Patricia Trie circuit. The witnesses are currently generated [here](https://github.com/miha-stopar/mpt). Lookups in the Keccak circuit are simulated using internal lookup table.

This branch replaces https://github.com/appliedzkp/zkevm-circuits/commits/merkle-patricia-trie (didn't rebase as it seems cleaner this way, will leave the old branch for a while before deleting it).


zkevm-circuits/src/mpt_circuit/storage_leaf.rs
```diff
@@ -0,0 +1,436 @@
...
+        node: &Node,
+        rlp_values: &[RLPItemWitness],
+    ) -> Result<(), Error> {
+        let storage = &node.storage.clone().unwrap();
+
+        let key_items = [
+            rlp_values[StorageRowType::KeyS as usize].clone(),
+            rlp_values[StorageRowType::KeyC as usize].clone(),
+        ];
+        let value_item = [
```
miha-stopar, line 277: @Brecht: It seems now the RLP bytes length for storage value is always 1. This actually never happens - because it's either 0 (when value is of length 1) or 2 (when value is of length bigger than 1). None of the constraints fail because it affects only the RLC of the value, which gets stored into the MPT table and currently we have no lookups into it yet.

Shall I modify this?

Brechtpd: I had the storage values wrong for a long while so it wouldn't surprise me if I still have it wrong! 

Seems like I at least documented what I think should happen above:

>  // Because the storage value is an rlp encoded string inside another rlp encoded
// string (leaves are always encoded as [key, value], with
// `value` here containing a single stored value) the stored
// value is either stored directly in the RLP encoded string if short, or stored
// wrapped inside another RLP encoded string if long.

So the RLP value stored here is the internal value, not the main RLP value (see line 114 for that one). So this one will only be used when the RLP value is long. I hope this makes sense.

miha-stopar: Ah, true! The value was correctly computed!

However, I find a bit confusing that in the case of the short RLP, the descriptor contains something like `value_rlp_bytes":[[1],[161]]`. In this example, the previous value RLP is short and the current value RLP is long. Thus, in the short case, `value_rlp_bytes` stores the actual value (being `1` in this case) and not the RLP byte, while the `value` is `[0,0,...]`.

In https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1353/commits/92ef471f51e5c70b7e3ececa7a2a3b3b34d172fb I changed to `value_rlp_bytes":[[],[161]]`
and `value` to be `[1,0,...]`. Can you have a look if it seems right to you?

Brechtpd: I think this could work, but I also think it complicates things a bit as well because it breaks the meaning of what "short" is. `value_rlp_bytes":[[1],[161]]` for me makes sense (well, RLP encoding itself is a bit crazy) because the `1` is indeed the RLP byte, it also just happens to be the value.

So moving the `1` to the value bytes normally used for long values I think works, but also a bit non-standard compared to how short values are handled elsewhere (`RLPValueGadget` functions currently would return the wrong values for most things, though I guess this would be fixable by having a special mode for no data, instead of reusing the short flag).

Maybe another argument is that the storage leaf code works similarly to how the account leaf code works before, it's just that the account leaf always stores multiple values so the values stored in `value_rlp_bytes` are always actual RLP bytes.

But I don't really have a strong opinion on it.

miha-stopar: My impression was that `value_rlp_bytes` is used for RLP bytes (the ones that specify the length, not the actual value). But I don't have a strong opinion either and thus I reverted to the previous (your) version. Thanks, Brecht!

zkevm-circuits/src/mpt_circuit/account_leaf.rs
```diff
@@ -0,0 +1,589 @@
...
+            ifx! {config.is_account_delete_mod => {
+                // Account delete
+                // We need to make sure there is no leaf when account is deleted. Two possible
+                // cases:
+                // - 1. Account leaf is deleted and there is a nil object in
+                // branch. In this case we have a placeholder leaf.
+                // - 2. Account leaf is deleted from a branch with two leaves, the remaining
+                // leaf moves one level up and replaces the branch. In this case we
+                // have a branch placeholder.
+                // TODO(Brecht): For case 2: just having the parent branch be the placeholder seems not enough
```
miha-stopar, line 293: @Brechtpd The other constraints are covered by `DriftedGadget`, right?

zkevm-circuits/src/mpt_circuit/helpers.rs
```diff
@@ -0,0 +1,1382 @@
...
+    ) {
+        self.base.store_dynamic_table(description, tag, values)
+    }
+}
+
+/// Returns `1` when `value == 0`, and returns `0` otherwise.
+#[derive(Clone, Debug, Default)]
+pub struct IsEmptyTreeGadget<F> {
+    is_in_empty_trie: IsEqualGadget<F>,
+    is_in_empty_branch: IsEqualGadget<F>,
```
miha-stopar, line 986: @Brechtpd Not that important, but perhaps we could use some other name instead of `is_in_empty_trie` and `is_in_empty_branch`? Seems a bit strange to be something in the empty something.

zkevm-circuits/src/mpt_circuit/helpers.rs
```diff
@@ -0,0 +1,1382 @@
...
+
+impl<F: Field> WrongGadget<F> {
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn construct(
+        cb: &mut MPTConstraintBuilder<F>,
+        expected_address: Expression<F>,
+        is_non_existing: Expression<F>,
+        key_value: &RLPItemView<F>,
+        key_rlc: &Expression<F>,
+        wrong_item: &RLPItemView<F>,
```
miha-stopar, line 1128: @Brechtpd Maybe `wrong_item` is a bit misleading as it actually contains the address of the enquired address? The "wrong" address is actually stored in the `key_rlc`.

zkevm-circuits/src/mpt_circuit/account_leaf.rs
```diff
@@ -0,0 +1,589 @@
...
+                &wrong_bytes,
+                config.is_in_empty_trie[true.idx()].expr(),
+                config.key_data[true.idx()].clone(),
+                &ctx.r,
+            );
+            let is_non_existing_account = not::expr(config.wrong.is_key_equal.clone().expr());
+
+            // Anything following this node is below the account
+            // TODO(Brecht): For non-existing accounts it should be impossible to prove
+            // storage leaves unless it's also a non-existing proof?
```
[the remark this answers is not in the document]
Brechtpd: It's been quite a while since I wrote this so I'm trying to reconstruct why I thought that so I may have been wrong then and I most definitely could be wrong now. :)

- Good point on the address not matching!
- For placeholder leafs the account data passed in by the prover could be anything I think, so the storage root hash could be any value passed into the circuit by the prover no?

Do we need to support `StorageDoesNotExist` for non-existing accounts?

zkevm-circuits/src/mpt_circuit/account_leaf.rs
Brechtpd: Yeah seems ok to me now, not sure if I thought of any other problems

zkevm-circuits/src/mpt_circuit/helpers.rs
Brechtpd: Any suggestions? For me it kind of makes sense, but I can understand how it could be confusing, though I can't immediately come up with a different name.

zkevm-circuits/src/mpt_circuit/account_leaf.rs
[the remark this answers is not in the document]
miha-stopar: When there is a `NonExistingAccount` proof, any `StorageChanged` lookup into `MPTTable` would fail (as `proof_type` is part of the table).

If the attacker would try to prepare a `StorageChanged` proof with a non-existing account, the address wouldn't match when there is a wrong leaf. But indeed I am not sure whether the proof would fail in the case of the placeholder leaf. I think we would need to add a constraint to allow the placeholder leaf only for the `NonExistingAccount` proof. I can add this in the PR I am preparing.

Regarding `StorageDoesNotExist` for non-existing accounts - I think we don't need it, but @ed255 would know better.

zkevm-circuits/src/mpt_circuit/helpers.rs
miha-stopar: It seems to me it would be easier to read if we would have:
```
pub struct IsPlaceholderLeafGadget<F> {
    is_empty_trie: IsEqualGadget<F>,
    is_nil_in_branch_at_mod_index: IsEqualGadget<F>,
}
```
instead of
```
pub struct IsEmptyTreeGadget<F> {
    is_in_empty_trie: IsEqualGadget<F>,
    is_in_empty_branch: IsEqualGadget<F>,
}
```

And `is_placeholder_leaf` instead of `is_in_empty_trie` in `account_leaf.rs` and `storage_leaf.rs`.

What do you think?

zkevm-circuits/src/table/mpt_table.rs
```diff
@@ -1,68 +1,134 @@
...
+    Disabled,
     /// Nonce updated
-    NonceMod = AccountFieldTag::Nonce as isize,
+    NonceChanged = AccountFieldTag::Nonce as isize,
     /// Balance updated
-    BalanceMod = AccountFieldTag::Balance as isize,
+    BalanceChanged = AccountFieldTag::Balance as isize,
     /// Code hash exists
-    CodeHashMod = AccountFieldTag::CodeHash as isize,
+    CodeHashExists = AccountFieldTag::CodeHash as isize,
```
miha-stopar, line 23: @Brechtpd At some point I thought we will need only `CodeHashExists`, but after some discussions we realised we also need `CodeHashChanged`. Is it ok to change the name?

zkevm-circuits/src/mpt_circuit.rs
```diff
@@ -0,0 +1,719 @@
...
+
+                    // Main MPT circuit
+                    // State machine
+                    cb.base.set_cell_manager(state_cm.clone());
+                    ifx! {f!(q_first) + f!(q_last) => {
+                        require!(a!(state_machine.is_start) => true);
+                    }};
+                    // Main state machine
+                    matchx! {
+                        a!(state_machine.is_start) => {
```
miha-stopar, line 295: @Brechtpd I think there are no constraints currently about which nodes can follow which nodes, but after some thought I think these are indeed not necessary as (in `account_leaf.rs`)
```
require!(config.main_data.is_below_account => false);
```
guarantees there are no two or more account leaves. Another possible attack would be to have two storage proofs one after another, but the probability is again negligible as the second storage trie would need to hash to 0 as (the first) `storage_leaf` stores the default values (`parent_data.rlc = 0`). So I think it's all good :)

zkevm-circuits/src/table/mpt_table.rs
Brechtpd: Oh yes of course, will make it part of the next batch of changes together with the name changes above!

zkevm-circuits/src/mpt_circuit.rs
Brechtpd: Yes I also believe it should be fine but could miss some case. I actually thought `is_below_account` was set to false again after the storage leaf, but looks like that isn't the case so seem like the hash being `0` is indeed the only thing stopping it from happening. Perhaps it can't hurt to also update `is_below_account` though. :)

zkevm-circuits/src/mpt_circuit/rlp_gadgets.rs
```diff
@@ -0,0 +1,770 @@
...
+#[derive(Clone, Debug, Default)]
+pub(crate) struct RLPListWitness {
+    pub(crate) is_short: bool,
+    pub(crate) is_long: bool,
+    pub(crate) is_very_long: bool,
+    pub(crate) is_string: bool,
+    pub(crate) bytes: Vec<u8>,
+}
+
+impl<F: Field> RLPListGadget<F> {
```
miha-stopar, line 77: If I am not mistaken, there are no constraints that would ensure that the byte streams are properly RLP encoded currently. So, for example for a storage leaf
`
[226 160 59 138 106 70 105 186 37 13 38 205 122 69 158 202 157 33 95 131 7 227 58 235 229 3 121 188 90 54 23 236 52 68 1]
`
the RLP gadget decodes `226` to `is_list=true` and `is_short=true`, from which it then knows how to compute for example the number of the RLP bytes. But the byte stream like
`
[227 160 59 138 106 70 105 186 37 13 38 205 122 69 158 202 157 33 95 131 7 227 58 235 229 3 121 188 90 54 23 236 52 68 1 x]
`
would be accepted as well (as far as RLP gadget is concerned), although you would still need to find the hash collision to make the attack work.

In the old code I had constraints like
`
226 - 192 = 160 - 128 + 2
`
But indeed because of the hashing these might not be necessary.

@Brechtpd  Do you think that checking that the byte streams are properly RLP encoded is not necessary?

miha-stopar: Ah, yes, they are there! Sorry, my brain was formatted too much during vacation :)

ChihChengLiang (commented):
I see the PR is getting closer to being merged.
@miha-stopar @Brechtpd @CeciliaZ030 Epic work on MPT!
Question for any of you: Is there anything you think the reviewers can focus on reviewing it? Otherwise, we can approve directly and unblock the merge.

Currently, I only quickly checked the files to see if any leftover files that shouldn't be included. The PR looks great for this part.

Brechtpd:
I think the most important thing is that the non-MPT specific changes are all good to go. If that's okay I think should be okay to merge in so it's easier to keep in sync/improve on/integrate better with the other code afterwards.

For a bit deeper review, I think also good to take a quick look at the main circuit implementation files (`mpt_circuit.rs`, `account_leaf.rs`, `storage_leaf.rs`, `extension_branch.rs`, `branch.rs`, `extension.rs`) to get a feeling how the circuit works.

I think the other files don't need much looking into for now unless to check how something works a bit more deeply. They are used only in the MPT circuit and the code there should be shared better with the other circuits once this PR is merged in (quite a bit of duplication there now to prevent merge conflicts and keep the diff clean).

ChihChengLiang (approved):
LGTM!
@Brechtpd Thanks for the reviewing guide!

Brechtpd (approved):

PR #401, opened 2022-03-22
fix "gen_state_access_trace" when handling the call stacks

ripwu:
There are two bugs when generating state access list from `GethExecTrace` by `gen_state_access_trace()`:

1. `CRAETE*` should push stack frame in a normal case.

2. `*CALL*` shouldn't push stack frame if there is no need. (e.g. CALL an EOA)

Now the program will panic and throw:

```
thread 'main' panicked at 'gen_inputs for BLOCK_NUM: InvalidGethExecStep("gen_state_access_trace: call stack will be empty", Step { pc: 0x0004, op: REVERT, gas: 170031, gas_cost: 0, depth: 2, error: None, stack: [0x0, 0x0], memory: [], storage: {} })', prover/src/bin/prover_cmd.rs:65:10
```

This PR aims to fix it.

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -1680,10 +1680,10 @@ pub fn gen_state_access_trace<TX>(
         let (contract_address, code_source) = &call_stack[call_stack.len() - 1];
         let (contract_address, code_source) = (*contract_address, *code_source);
 
-        let (mut push_stack, mut pop_stack) = (false, false);
+        let (mut push_call_stack, mut pop_call_stack) = (false, false);
         if let Some(next_step) = next_step {
-            push_stack = step.depth + 1 == next_step.depth;
-            pop_stack = step.depth - 1 == next_step.depth;
+            push_call_stack = step.depth + 1 == next_step.depth;
+            pop_call_stack = step.depth - 1 == next_step.depth;
         }
 
         match step.op {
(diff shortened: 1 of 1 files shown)


bus-mapping/src/circuit_input_builder.rs
```diff
@@ -1679,6 +1679,13 @@ pub fn gen_state_access_trace<TX>(
         let i = Some(index);
         let (contract_address, code_source) = &call_stack[call_stack.len() - 1];
         let (contract_address, code_source) = (*contract_address, *code_source);
+
+        let (mut push_stack, mut pop_stack) = (false, false);
```
ed255, line 1683: In the EVM context there's many places where push_stack and pop_stack could be understood as pushing a word to the stack or popping a word to the stack.  I suggest renaming these variables to `push_call_stack` and `pop_call_stack` to make the meaning more clear.

ed255 (commented):
Thanks a lot for submitting this fix and the corresponding test cases :)
The fix looks good to me!
I've just added a suggestion to change two variable names for clarity.

bus-mapping/src/circuit_input_builder.rs
ripwu: The variables had been renamed.

ed255 (approved):
LGTM! Thanks a lot for your contribution!

CPerezz (approved):
LGTM!!

Thanks for this work and thanks @ed255 for the suggestions!!


Feel free to merge! :smile:

PR #413, opened 2022-03-30
add blockCtxOp gadgets

bchyl:
1.  only can merge 3 different size RLC length because const generic  `const N: usize` must been determined at compilation when config gadget.  
```
#[derive(Clone, Debug)]
pub(crate) struct BlockCtxGadget<F> {
    same_context: SameContextGadget<F>,
    block_ctx_value: RandomLinearCombinationN<F>,
}

#[derive(Clone, Debug)]
pub enum RandomLinearCombinationN<F> {
    RLCU64(RandomLinearCombination<F, N_BYTES_U64>),
    RLCU160(RandomLinearCombination<F, N_BYTES_ACCOUNT_ADDRESS>),
    Word(Word<F>),
}
```
can't match enum  type even if using type reflection:
```
use std::any::Any;
...
let value = cb.query_rlc();
let value_type = &value as &dyn Any;
match value_type.downcast_ref::<RandomLinearCombination<F, N_BYTES_U64>>() {
            Some(bcv) => (Some(RandomLinearCombinationN::RLCU64(bcv.clone())), Some(from_bytes::expr(&value.cells))),
            None => (None,None),
}

match value_type.downcast_ref::<RandomLinearCombination<F, N_BYTES_ACCOUNT_ADDRESS>>() {...}
match value_type.downcast_ref::<Word<F>>() {...}
...
```
but imo it is reasonable because the const N  must been determined at compilation when starting to config gadget rather than runtime step into specific op.

2. `BlockContextFieldTag` Keep the sequence consistent with `OpcodeId` in instrcution sets  for scalar in order to blockctx lookup op.

3.  3 kinds ops for scalar:  `u64(TIMESTAMP/NUMBER/GASLIMIT) and u160(COINBASE)` and `u256(DIFFICULTY/BASEFEE)`, so 3 different `gadget` for them.  And the first two kinds are values, and the last one is RLC value.

4. vector block ctx op `BLOCKHASH` will soon be implemented after scalar.


DreamWuGit:
Do you try put byte number in BlockCtxGadget as :
```
pub(crate) struct BlockCtxGadget<F,  const N_BYTES: usize> {
    same_context: SameContextGadget<F>,
    value: RandomLinearCombination<F, N_BYTES>,
}
```
and then construct several instances like: 
  ```
    block_ctx_u64_gadget: BlockCtxU64Gadget<F, 64>,
    block_ctx_u160_gadget: BlockCtxU160Gadget<F, 20>,
    block_ctx_u256_gadget: BlockCtxU256Gadget<F,  32>,
```
if it is feasible, it can reuse one BlockCtxGadget .

**Comment by bchyl** [comment]
> Do you try putting byte number in BlockCtxGadget as :
> 
> ```
> pub(crate) struct BlockCtxGadget<F,  const N_BYTES: usize> {
>     same_context: SameContextGadget<F>,
>     value: RandomLinearCombination<F, N_BYTES>,
> }
> ```
> 
> and then construct several instances like:
> 
> ```
>   block_ctx_u64_gadget: BlockCtxU64Gadget<F, 64>,
>   block_ctx_u160_gadget: BlockCtxU160Gadget<F, 20>,
>   block_ctx_u256_gadget: BlockCtxU256Gadget<F,  32>,
> ```
> 
> if it is feasible, it can reuse one BlockCtxGadget .

yeah,  now  use `BlockCtxGadget` for construct instance, and const generic parameters instances at  the inner layer concrete `BlockCtxU64Gadget` structure.
thanks very much.

**Comment by han0110** [comment]
Sorry for late review, just saw I am assigned to be reviewer. After reviewing the PR, all make sense to me, but another easier way to reuse the gadget for all block-context-family opcodes just comes to my mind (related to appliedzkp/zkevm-specs#101).

The idea is to constrain the most significant byte to be `0` when it's expected to contain less than 32 bytes, because the `block_table` already contains valid range data, so as long as it has no chance to wrap around the field, the extra bytes must be `0`. As for how to enable this constraint dynamically, we can put a auxiliary bit `not_rlc` in the last column of `ResponsibleOpcode` fixed lookup, so the constraint for most significant byte would be `not_rlc * bytes[-1]`, and finally lookup `block_table` by `not_rlc * lc(bytes, 256) + (1 - not_rlc) * lc(bytes, r)`.

Do you think we could rework on this PR directly, or merge this first and implement the idea in further PR?

**Comment by bchyl** [comment]
> 



> The idea is to constrain the most significant byte to be `0` when it's expected to contain less than 32 bytes, because the `block_table` already contains valid range data, so as long as it has no chance to wrap around the field, the extra bytes must be `0`. As for how to enable this constraint dynamically, we can put a auxiliary bit `not_rlc` in the last column of `ResponsibleOpcode` fixed lookup, so the constraint for most significant byte would be `not_rlc * bytes[-1]`, and finally lookup `block_table` by `not_rlc * lc(bytes, 256) + (1 - not_rlc) * lc(bytes, r)`.

nice, got it!

**Comment by bchyl** [comment]
> Do you think we could rework on this PR directly, or merge this first and implement the idea in further PR?

Maybe we can merge this first and implement the optimized version of  the idea in next new PR?

**Comment by han0110** [comment]
> Maybe we can merge this first and implement the optimized version of the idea in next new PR?

Yes, make sense to me, we would also expect spec of the new idea first, let's try to merge this first.

File: zkevm-circuits/src/evm_circuit/execution/block_ctx.rs [hunk 1]
```diff
@@ -0,0 +1,238 @@
...
+    #[test]
+    fn blockcxt_u64_gadget_test() {
+        let bytecode = bytecode! {
+            TIMESTAMP
+            STOP
+            NUMBER
+            STOP
+            GASLIMIT
+            STOP
+        };
```
han0110, line 216: You seem to want `POP` instead of `STOP` right?

bchyl: yeah, nice catch!

bchyl: fix block gaslimit cases for trace

han0110 (approved):
LGTM! Thanks!

zkevm-circuits/src/evm_circuit/execution/block_ctx.rs
```diff
@@ -0,0 +1,238 @@
...
+pub(crate) struct BlockCtxGadget<F, const N_BYTES: usize> {
+    same_context: SameContextGadget<F>,
+    value: RandomLinearCombination<F, N_BYTES>,
+}
+
+impl<F: Field, const N_BYTES: usize> BlockCtxGadget<F, N_BYTES> {
+    fn construct(cb: &mut ConstraintBuilder<F>) -> Self {
+        let value = cb.query_rlc();
+
+        // Push the u64 value to the stack
```
DreamWuGit, line 32:
```suggestion
        // Push the u64 value to the stack
```
```suggestion
        // Push the N_BYTES value to the stack
```

DreamWuGit, line 32:
```suggestion
        // Push the u64 value to the stack
```
```suggestion
        // Push the N_BYTES value to the stack
```

DreamWuGit, line 32:
```suggestion
        // Push the N_BYTES value to the stack
```

DreamWuGit, line 32:
```suggestion
        // Push the N_BYTES value to the stack
```

bchyl: yeah, fixed.
thanks!

DreamWuGit (approved):
since will handle variable size RLC length in future , Good to me now, Thanks !

PR #417, opened 2022-04-01
Degree reduction

Brechtpd:
Step 2/3 of experimental optimizations done here: https://github.com/Brechtpd/zkevm-circuits/pull/2.
Code is based on https://github.com/appliedzkp/zkevm-circuits/pull/416 so ideally that one should be merged first.
Code depends on some small halo2 changes: https://github.com/appliedzkp/halo2/pull/54

Currently only tries to reduced the degree to a max of 9 (which is an extended degree of 8x, lowered from 16x), because more is currently not really possible until the lookups are done in a different way (which will be done in a subsequent PR).

The degree is lowered which makes the circuit faster to prove (around ~30% faster), though this comes at the cost of using more cells which can increase the number of rows that are required for an opcode.

```
ADD: 9 rows
MUL: 9 rows
BITWISE: 9 rows
BeginTx: 13 rows
BYTE: 11 rows
CALLDATACOPY: 7 rows
CALLDATALOAD: 13 rows
CALLDATASIZE: 6 rows
CALLER: 6 rows
CALLVALUE: 6 rows
CMP: 9 rows
DUP: 6 rows
EndBlock: 4 rows
EndTx: 15 rows
ErrorOutOfGasStaticMemoryExpansion: 8 rows
JUMP: 6 rows
JUMPDEST: 6 rows
JUMPI: 6 rows
GAS: 6 rows
MEMORY: 8 rows
COPYTOMEMORY: 19 rows
PC: 6 rows
POP: 6 rows
PUSH: 8 rows
SELFBALANCE: 6 rows
SCMP: 9 rows
SIGNEXTEND: 11 rows
STOP: 5 rows
SWAP: 6 rows
MSIZE: 6 rows
COINBASE: 6 rows
TIMESTAMP: 6 rows
NUMBER: 6 rows
SLOAD: 6 rows
SSTORE: 6 rows
EXTCODEHASH: 6 rows
ISZERO: 6 rows
```

These are not optimal, but I will be tweaking things a bit more after the lookup refactor.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
+++ b/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
@@ -243,28 +243,33 @@ impl<'a, F: FieldExt> ConstraintBuilder<'a, F> {
         Vec<(&'static str, Expression<F>)>,
         Vec<(&'static str, Lookup<F>)>,
         Vec<Preset<F>>,
+        usize,
     ) {
         let mut constraints = self.cb.constraints;
         let mut presets = Vec::new();
 
+        let mut height = 0;
         for (row, usage) in self.curr.rows.iter().zip(self.curr_row_usages.iter()) {
             if usage.is_byte_lookup_enabled {
                 constraints.push(("Enable byte lookup", row.qs_byte_lookup.expr() - 1.expr()));
             }
+            if usage.next_idx > 0 {
+                height += 1;
 
-            presets.extend(
-                row.cells[usage.next_idx..]
-                    .iter()
-                    .map(|cell| (cell.clone(), F::zero())),
-            );
-            presets.push((
-                row.qs_byte_lookup.clone(),
-                if usage.is_byte_lookup_enabled {
-                    F::one()
-                } else {
-                    F::zero()
-                },
-            ));
+                presets.extend(
+                    row.cells[usage.next_idx..]
+                        .iter()
+                        .map(|cell| (cell.clone(), F::zero())),
+                );
+                presets.push((
+                    row.qs_byte_lookup.clone(),
+                    if usage.is_byte_lookup_enabled {
+                        F::one()
+                    } else {
+                        F::zero()
+                    },
+                ));
+            }
         }
 
         let execution_state_selector = self.curr.execution_state_selector([self.execution_state]);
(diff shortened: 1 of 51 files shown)


Brechtpd:
One thing I had to do to get the EVM bench working is increase the stack size by using `RUST_MIN_STACK=4194304 ` because of an out of stack memory error. I added this to the makefile, but it looks like the ci bench doesn't use the makefile so the bench on github failed. I tried to update the ci script but not sure I did that correctly.

It looks like this stack error is related to the recursive nature of the evaluation using vanilla halo2: https://github.com/zcash/halo2/issues/536. If so it seems likely https://github.com/appliedzkp/halo2/pull/44 will fix this (no recursion is used anymore to do the largest evaluation)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1024,19 +1042,29 @@ impl<'a, F: FieldExt> ConstraintBuilder<'a, F> {
...
     }
 
     pub(crate) fn add_constraint(&mut self, name: &'static str, constraint: Expression<F>) {
-        self.cb.add_constraint(name, constraint);
+        let constraint = match &self.condition {
+            Some(condition) => condition.clone() * constraint,
+            None => constraint,
+        };
+
+        let constraint = self.split_expression(name, constraint, MAX_DEGREE - 2);
```
han0110, line 1056: It seems we subtract the implicit degree twice, just tested locally and the `extended_k` still remains `3` (`validate_degree` also needs to be updated).
```suggestion
        let constraint = self.split_expression(name, constraint, MAX_DEGREE);
```

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1024,19 +1042,29 @@ impl<'a, F: FieldExt> ConstraintBuilder<'a, F> {
...
+
+        self.validate_degree(constraint.degree(), name);
+        self.constraints.push((name, constraint));
     }
 
     pub(crate) fn add_constraint_first_step(
         &mut self,
         name: &'static str,
         constraint: Expression<F>,
     ) {
```
han0110, line 1066: It seems we can also remove the `+ 1`  due to `Add 1 more degree due to the selector` below, I think it was a mistake by me.

han0110 (commented):
LGTM! The degree reduction will be extremely helpful to developer, we can remove all the `ConstrainBuilder::copy` to cap the degree by ourselves in the future.

Brechtpd:
Closed in favor of #425.

PR #420, opened 2022-04-02
deny debug_assert_with_mut_call in clippy

lispc:
Side effect in debug_assert will cause bugs in external codes using "zkevm-circuits" as dependency if they use default release profile.


han0110 (approved):
LGTM. Thanks for fixing this.

CPerezz (approved):
LGTM! Thanks!!

ed255 (approved):
LGTM!

PR #441, opened 2022-04-08
prover: add http/json-rpc daemon command

pinkiebell:
* Moves the proof computation function from prover_cmd into a module.
* Add prover_rpcd that provides a task-queue-like JSON-RPC interface
  for dispatching proof generation work.


ChihChengLiang (approved):
LGTM. Sorry for the late review

pinkiebell:
Thanks for the review! 🚢

PR #451, opened 2022-04-10
[Fix] Constrain `code_source` to be 0 for root and create case

icemelon:
It'd be a bug that the constraint doesn't match the description.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
+++ b/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
@@ -543,7 +543,7 @@ impl<'a, F: FieldExt> ConstraintBuilder<'a, F> {
         let is_root_create = self.curr.state.is_root.expr() * self.curr.state.is_create.expr();
         self.add_constraint(
             "The opcode source when is_root and is_create (Root creation transaction) is not determined yet",
-            is_root_create.clone(),
+            is_root_create.clone() * self.curr.state.code_source.expr(),
         );
         self.add_lookup(
             "Opcode lookup",


DreamWuGit:
seems to me old design is  `code_source ` refer to `tx_id` when it is root call ,  changed anywhere ?

han0110:
This was left as a todo which makes the proving fail, but we should just remove this constraint and mark it as a todo. We can start to implement it after we decide how to handle opcode lookup of creation transaction (see appliedzkp/zkevm-specs#73).

ed255:
> seems to me old design is `code_source ` refer to `tx_id` when it is root call , changed anywhere ?

As far as I know it's still like this.

I think it's better to follow @han0110 comment: remove this constraint and leave a TODO there.  Furthermore, in the witness converter or the function that assigns the values for the step state, we can add a `log::warning` that prints a message when `is_root` and `is_create` are both true.  This way we can get a nice message when trying a circuit with a functionality that is not yet implemented.

DreamWuGit:
> > seems to me old design is `code_source ` refer to `tx_id` when it is root call , changed anywhere ?
> 
> As far as I know it's still like this.
> 
> I think it's better to follow @han0110 comment: remove this constraint and leave a TODO there. Furthermore, in the witness converter or the function that assigns the values for the step state, we can add a `log::warning` that prints a message when `is_root` and `is_create` are both true. This way we can get a nice message when trying a circuit with a functionality that is not yet implemented.

Agree !

icemelon:
@ed255 @DreamWuGit Sounds good. I will change it into a TODO here and link to @han0110's issue about this.

PR #462, opened 2022-04-14
Implement constraints in state circuit spec

z2trillion:
This implements the constraints found in https://github.com/appliedzkp/zkevm-specs/blob/master/src/zkevm_specs/state.py, except for the https://github.com/appliedzkp/zkevm-specs/blob/f33ae436011cde6cf76b92e316e00662cd4ef2e1/src/zkevm_specs/state.py#L178 and https://github.com/appliedzkp/zkevm-specs/blob/master/src/zkevm_specs/state.py#L209. Adding those will require modifying the bus mapping to add the Rw::Start rows, so I will do it in a separate PR.


zkevm-circuits/src/state_circuit.rs
```diff
@@ -1,4 +1,273 @@
...
+impl<F: Field> Circuit<F> for StateCircuit<F> {
+    type Config = StateConfig<F>;
+    type FloorPlanner = SimpleFloorPlanner;
+
+    fn without_witnesses(&self) -> Self {
+        Self::default()
+    }
+
+    fn configure(meta: &mut ConstraintSystem<F>) -> Self::Config {
+        let selector = meta.fixed_column();
```
lispc, line 98: what about change selector into advice column?   Since it is not "fixed" now. ( We assign "1" to it only when a row is  non empty, so this column is dynamic.

ntampakas:
Bench failed with the following error:

```
running 1 test
test state_circuit::tests::bench_state_circuit_prover has been running for over 60 seconds
End:     Setup generation with degree = 19 .........................................582.270s
[zkevm-circuits/src/state_circuit.rs:171] self.rows.clone() = []
[zkevm-circuits/src/state_circuit.rs:171] self.rows.clone() = []
Start:   State Proof generation with 19 rows
thread 'state_circuit::tests::bench_state_circuit_prover' panicked at 'proof generation should not fail: InvalidInstances', circuit-benchmarks/src/state_circu
it.rs:53:10
test state_circuit::tests::bench_state_circuit_prover ... FAILED

failures:

failures:
    state_circuit::tests::bench_state_circuit_prover

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 618.72s

error: test failed, to rerun pass '-p circuit-benchmarks --lib'

```

lispc:
> Bench failed with the following error:
> 
> ```
> running 1 test
> Start:   Setup generation with degree = 19
> test state_circuit::tests::bench_state_circuit_prover has been running for over 60 seconds
> End:     Setup generation with degree = 19 .........................................582.270s
> [zkevm-circuits/src/state_circuit.rs:171] self.rows.clone() = []
> [zkevm-circuits/src/state_circuit.rs:171] self.rows.clone() = []
> Start:   State Proof generation with 19 rows
> thread 'state_circuit::tests::bench_state_circuit_prover' panicked at 'proof generation should not fail: InvalidInstances', circuit-benchmarks/src/state_circu
> it.rs:53:10
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> test state_circuit::tests::bench_state_circuit_prover ... FAILED
> 
> failures:
> 
> failures:
>     state_circuit::tests::bench_state_circuit_prover
> 
> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2 filtered out; finished in 618.72s
> 
> error: test failed, to rerun pass '-p circuit-benchmarks --lib'
> ```

related to <https://github.com/appliedzkp/zkevm-circuits/issues/458>

zkevm-circuits/src/state_circuit.rs
z2trillion: There needs to be a fixed column or selector involved in each of the constraints, or the MockProver will return a `ConstraintPoisoned` error when it tries to verify. 

We can fill up all the otherwise empty rows with `Rw::Start`'s so that the `selector` column will not be dynamic. I haven't quite figured out how to do this without massively slowing down the tests though.

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -0,0 +1,326 @@
...
+        );
+        // // for every first access, we add an Account write to setup the value
+        // from the // previous block with rw_counter = 0
+        // self.condition(q.first_access(), |cb| {
+        //     // cb.require_zero("first access is a write", q.is_write());
+        //     cb.require_zero("first access rw_counter is 0",
+        // q.rw_counter.value.clone()); });
+    }
+
+    fn build_account_destructed_constraints(&mut self, q: &Queries<F>) {
```
miha-stopar, line 209: A similar method for when an account is created is probably needed?
Also, an MPT lookup is needed here too, right (whenever something changes in the state trie)?

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -0,0 +1,278 @@
...
+use std::ops::Mul;
+
+// 2 limbs for tag, field_tag and id.
+// 10 limbs for address,
+// 16 limbs for storage key
+// 2 limbs for rw_counter
+// 30 limbs in total -> can fit into two field elements
+#[derive(Clone)]
+pub struct Config<F: Field> {
+    diff_1: Column<Advice>,
```
miha-stopar, line 23: Would it be possible to add a short description of this chip and especially of `diff_1` and `diff_2` fields?

miha-stopar (commented):
I really like the new structure, it's really nice and easily readable! I just did a first pass though, will do some more reviewing in the next days.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
z2trillion: I've added a comment describing this chip. I would love to hear any suggestions for better names for `diff_1` and `diff_2`.

miha-stopar: Thanks, it makes reading the code much easier! For `diff_1` and `diff_2` naming I am not sure - perhaps something that would give some intuition, like `aggregated_limbs_1`, `aggregated_limbs_2`? It would be also good perhaps to add some rationale why limbs are split into two parts, it's because of field overflow, right?

zkevm-circuits/src/state_circuit/constraint_builder.rs
z2trillion: These are the constraints for rows that match `Rw::AccountDestructed`. I believe that up to three `Rw::Account` will exist in the rw table for account creations? The constraints for these rows are in the `build_account_constraints` function.

I agree MPT lookup(s) will be required here. I need to review the specs to see what our plans (if any) are for handling account desctruction rows.

zkevm-circuits/src/state_circuit.rs
lispc: verify_at_rows can be helpful for this problem.

https://github.com/scroll-tech/zkevm-circuits/pull/140/commits/34eb52fb2244ab602c04cab0b048ec8f8a9c9644

zkevm-circuits/src/state_circuit/random_linear_combination.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct Config<const N: usize> {
+    pub encoded: Column<Advice>,
+    // bytes are little endian
+    pub bytes: [Column<Advice>; N],
+}
+
+#[derive(Clone)]
+pub struct Queries<F: Field, const N: usize> {
+    pub encoded: Expression<F>,
+    pub encoded_prev: Expression<F>,
```
DreamWuGit, line 20: want to know usage of `encoded_prev`, don't see its constraint anywhere

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -0,0 +1,316 @@
...
+        let prev_be_limbs = rw_to_be_limbs(prev);
+
+        let find_result = cur_be_limbs
+            .iter()
+            .zip(&prev_be_limbs)
+            .enumerate()
+            .find(|(_, (a, b))| a != b);
+        let (index, (cur_limb, prev_limb)) = find_result.expect("repeated rw counter");
+
+        let mut diff_1 = F::from((cur_limb - prev_limb).into());
```
miha-stopar, line 201: Doesn't this `diff_1` computation imply `diff_1` can only be `Ai - Bi` and not one of the following (cumulative) values specified above in the comment (and also used in the constraint):
 // C0 = A0 - B0,
 // C1 = C0 << 16 + A1 - B1,
 // ...
 // C14 = C13 << 16 + A14 - B14,

The constraints don't fail because Ci are 0 by having `diff_1` computed this way.

The `diff_1` computation implies only one change can occur in the first 15 limbs. Could be some explanation added why this is so? Also, specs repo doesn't seem to have info how lexicographic ordering is to be implemented, could that be added?

zkevm-circuits/src/state_circuit/constraint_builder.rs
miha-stopar: Yes, for the account creation I think we would have three modular changes - changing only one field at a time: nonce, balance, codehash. And we would need to add constraints for the order of these three changes. Is that right @han0110 @ed255 ?

But I think we could leave this for the future PR.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
icemelon: Why not just encode 6 fields into two 240-bit values, `A_hi` and `A_lo`? Then you just need to compare two 240-bit values, and only the diff needs to be encoded in 16-bit limbs for range check.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
z2trillion: I think it's more clear to view it the other way around. The only way that `diff_1` can only be one of the 15 values `C0, ..., C14`, and that `diff_1` can fit into 16 bits is for `diff_1` to be the first `Ai - Bi` such that `Aj = Bj` for j < i. 

You're right that multiple difference can occur between the first 15 pairs of limbs, but (I think) that the constraints are such that `diff_1` has to be the value of the most significant non-zero limb difference.

The specs are for lexicographic ordering in the spec repo are in the code here: https://github.com/scroll-tech/zkevm-specs/blob/master/src/zkevm_specs/state.py#L355. I plan to update that section and the book once the implementation is finalized here. I had to make some changes to how the bit-packing of the keys worked so that everything could fit into 2 field elements.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
z2trillion: That would definately be easier to understand and explain, but it would require 30 advice columns for the limbs of the `diff_hi` and `diff_lo` instead of the 4 we use now.

lispc: agree.  Proving cost of adding new gates is much smaller than adding new columns.

zkevm-circuits/src/test_util.rs
```diff
@@ -84,16 +85,13 @@ pub fn test_circuits_using_witness_block(
...
+    // TODO: use randomness as one of the circuit public input, since randomness in
+    // state circuit and evm circuit must be same
     if config.enable_state_circuit_test {
-        let state_circuit =
-            StateCircuit::<Fr, true, 2000, 200, 1023, 2000>::new(block.randomness, &block.rws);
-        let prover = MockProver::<Fr>::run(12, &state_circuit, vec![]).unwrap();
-        prover.verify()?;
+        let state_circuit = StateCircuit::new(block.randomness, block.rws);
+        let power_of_randomness = state_circuit.instance();
+        let prover = MockProver::<Fr>::run(18, &state_circuit, power_of_randomness).unwrap();
```
lispc, line 93: For evm circuit, we dynamically decide the k. So many small tests will use only k = 12 (without bitwise fixed table).
If we use k = 18 for every state circuit here(since some big fixed table is needed), the CI time will be quite long.

@han0110 @ed255 Do you have any thoughts in this?

z2trillion: what do you think about setting `enable_state_circuit_test` to be false for evm circuit tests, once we add more state circuit tests?

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
miha-stopar: I think I got it now, thanks!

I think it would be helpful to add such details as above to the comments and docs. Perhaps write something like (if I understood correctly):

Assignment: `diff_i` is change in the 15 limbs that occurs first (if `prev = (2, 1, 2, ...), curr = (2, 1, 6, ...)`, then `diff_i = 4`.
Constraint: `diff_i` will be equal to the first non-zero `Ci`.

Perhaps you could rename `diff_1`, `diff_2` to `first_diff_high`, `first_diff_low`?

z2trillion: I've renamed them to `upper_limb_difference` and `lower_limb_difference`.

miha-stopar: Cool, thanks!

zkevm-circuits/src/state_circuit/constraint_builder.rs
ed255: Regarding the account destructed, I remember from a discussion with @han0110 that he mentioned that after a SELFDESTRUCT the account is not immediately deleted.  Looking at the yellowpaper I see:

> `SELFDESTRUCT`: Halt execution and register account for later deletion.

So an account destructed row is not directly connected to the MPT, but we need to figure out a way to keep track of all pending account deletes, and at the step EndTx process all those deletions.  Maybe we need to introduce a new tag in the state that signals an account deletion (which must match a previous non-reverted account destruction).  Currently that's open to be specified.

Regarding account creation, I guess there are several ways to create an account:
1. Send balance to an empty account.
2. Tx with `to = 0x0`, CREATE, CREATE2

1 and 2 set different account fields.
I would assume that a non-existing account has the fields to a default value, and on account creation some of those fields are set to a different value.  This makes me wonder if we need to explicitly set all the fields, or just the ones that get a non-default value; and I also wonder how the MPT handles non-existing accounts VS existing accounts with default values in their fields.

zkevm-circuits/src/test_util.rs
ed255: > what do you think about setting `enable_state_circuit_test` to be false for evm circuit tests, once we add more state circuit tests?

This sounds like a good tradeoff to me.

ed255:
I assume that currently this PR is assigned to be reviewed by @miha-stopar (from the EF side) and either icemelon OR lispc OR DreamWuGit (from scroll-tech side).  Let me know if you'd like me to review it as well at any point!

lispc (approved):

z2trillion:
@miha, can you take a look at this again, please?

zkevm-circuits/src/state_circuit/constraint_builder.rs
miha-stopar: Yes, the accounts are deleted when root is [recomputed](https://github.com/ethereum/go-ethereum/blob/master/core/state/statedb.go#L861).

Non-existing accounts are not stored in the trie (no leaf at the position specified by the account address), while for the account with default values there is a leaf which contains default values. However, in MPT circuit, when there is no account, a placeholder leaf with default values is set, so the lookup should contain default value as previous value. Because currently MPT circuit allows only one change at a time, there would be four lookups needed for the account creation. Just working on that actually. Let me prepare a hackmd note on that.

miha-stopar (approved):
LGTM

zkevm-circuits/src/state_circuit/constraint_builder.rs
lispc: I am not very familiar with the current mpt circuit. But just a remainder: it is very dangerous to allow two tries to have same semantic. I mean "However, in MPT circuit, when there is no account, a placeholder leaf with default values is set, so the lookup should contain default value as previous value.", maybe we should be very careful about the soundness if we allow both "non exist account(not exist in tree)" and "account with default value", where same semantic trees can have different roots.

ed255: I think this EIP can be very useful when reasoning about empty accounts and how they are represented in the MPT: https://eips.ethereum.org/EIPS/eip-161

miha-stopar: > I am not very familiar with the current mpt circuit. But just a remainder: "However, in MPT circuit, when there is no account, a placeholder leaf with default values is set, so the lookup should contain default value as previous value.", maybe we should be very careful about the soundness if we allow both "non exist account(not exist in tree)" and "account with default value", where same semantic trees can have different roots.

Agree, but I think I just didn't phrase it clearly. The two tries (with non-existent account and account with default values) are not equivalent in the MPT circuit, it's just that in case of a non-existent account a placeholder leaf is added not to break the MPT circuit layout. This placeholder leaf is then ignored in the circuit, but the circuit of course checks that this leaf is actually empty - this check happens in the branch (below this branch a placeholder leaf resides).

icemelon (approved):
LGTM

zkevm-circuits/src/state_circuit.rs
lispc: "fill up all the otherwise empty rows with Rw::Start" is also a bit counterintuitive..
Maybe we can merge this PR first(since it may be not solved very cleanly now?), then revisit this problem (make sure vk generated from one circuit instance can verify another circuit) later.

PR #466, opened 2022-04-19
add TxReceipt constraint in begin & end tx 

DreamWuGit:
Spec pr is https://github.com/appliedzkp/zkevm-specs/pull/181
- [x] add TxReceipt  RW tag &buss mapping Op
- [x] begin tx gadget constrain log_id is 0
- [x] end tx buss mapping generates TxReceipt ops 
- [x]  end tx gadget constrain TxReceipt fields


bus-mapping/src/circuit_input_builder.rs
```diff
@@ -1524,9 +1524,17 @@ impl<'a> CircuitInputBuilder {
         eth_block: &EthBlock,
         geth_traces: &[eth_types::GethExecTrace],
     ) -> Result<(), Error> {
+        // record tx gas cumulative used
+        let mut tx_cumulative_gas: HashMap<usize, u64> = HashMap::new();
```
roynalnaruto, line 1528: You don't need to mention the type here, since it's known from the `handle_tx` function call.
```suggestion
        let mut tx_cumulative_gas = HashMap::new();
```

bus-mapping/src/operation.rs
```diff
@@ -734,6 +736,63 @@ impl Op for CallContextOp {
...
+    }
+}
+
+impl Op for TxReceiptOp {
+    fn into_enum(self) -> OpEnum {
+        OpEnum::TxReceipt(self)
+    }
+
+    fn reverse(&self) -> Self {
+        unreachable!("TxReceiptOp can't be reverted")
```
roynalnaruto, line 792: Should be:
```suggestion
        unreachable!("TxReceiptOp can't be reversed")
```
Right?

bus-mapping/src/evm/opcodes.rs
```diff
@@ -464,6 +477,56 @@ pub fn gen_end_tx_ops(state: &mut CircuitInputStateRef) -> Result<ExecStep, Erro
...
+    );
+
+    let log_id = exec_step.log_id;
+    state.push_op(
+        &mut exec_step,
+        RW::READ,
+        TxReceiptOp {
+            tx_id: state.tx_ctx.id(),
+            field: TxReceiptField::LogLength,
+            value: log_id as u64,
```
roynalnaruto, line 498: Is `log_id` zero indexed? I mean, for the first log, is `log_id == 0`?

In that case, we should have `LogLength == log_id + 1` right?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -187,6 +187,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             code_source: To(code_hash.expr()),
             gas_left: To(gas_left),
             reversible_write_counter: To(2.expr()),
+            log_id: To(0.expr()),
```
roynalnaruto, line 190: From here it seems `log_id` is zero indexed.

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
```diff
@@ -85,14 +93,50 @@ impl<F: Field> ExecutionGadget<F> for EndTxGadget<F> {
...
+        // constrain tx receipt fields
+        cb.tx_receipt_lookup(
+            tx_id.expr(),
+            TxReceiptFieldTag::PostStateOrStatus,
+            is_persistent.expr(),
+        );
+        cb.tx_receipt_lookup(
+            tx_id.expr(),
+            TxReceiptFieldTag::LogLength,
+            cb.curr.state.log_id.expr(),
```
roynalnaruto, line 113: Ditto, may be this needs to be addressed?

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
```diff
@@ -203,6 +250,23 @@ impl<F: Field> ExecutionGadget<F> for EndTxGadget<F> {
             coinbase_balance,
         )?;
 
+        let current_cumulative_gas_used: u64 = if tx.id == 1 {
+            0
+        } else {
+            let rw = &block.rws[(RwTableTag::TxReceipt, (tx.id - 1) * 3 - 1)];
```
roynalnaruto, line 256: I think we should refactor this line. In case the `TxReceiptField` has more variants, this line would lead to bugs in assignment.

Something more like:
```rust
let rw = block.rws.tx_receipt(tx.id, TxReceiptField::CumulativeGasUsed);
```

What do you think?

roynalnaruto (commented):
I have some suggestions and questions :)

bus-mapping/src/circuit_input_builder.rs
ChihChengLiang, line 1528: On a different subject: It looks like `tx_cumulative_gas` accumulates gas across all txs in the block, should we rename `tx_cumulative_gas` to `block_cumulative_gas`?

ChihChengLiang (approved):
LGTM, modulo @roynalnaruto's comment and a naming question.

bus-mapping/src/circuit_input_builder.rs
DreamWuGit: seems correct, will have a try !

bus-mapping/src/operation.rs
DreamWuGit: hmm, I have no sense of this,  there are multi places use `can't be reverted` for `CallContextOp` `MemoryOp` `StackOp` in this file

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
DreamWuGit: this is by design, no logs in the begin tx, hence set it to zero

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
DreamWuGit: as the above !

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
DreamWuGit: I am not really got what is the concerned issue :),  Here is to get previous tx 's `CumulativeGasUsed`,  since each tx receipt has three rows corresponding to three tags (PostStateOrStatus,  CumulativeGasUsed,  LogLength).  Glad to  know if you can make it more clear !

bus-mapping/src/circuit_input_builder.rs
DreamWuGit: yes, it accumulates gas across all txs in the block !  since by default we handle txs in one block,  Can we omit the `block` prefix or   `block_tx_cumulative_gas` which sounds good to me , what do you think @ChihChengLiang ?

roynalnaruto: My opinion would be to simply use `cumulative_gas_used`. The `tx_` is definitely misleading.

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
roynalnaruto: I meant that this part: `(tx.id - 1) * 3 - 1` looks a bit messy and prone to errors in case the number of fields is edited in the future. Hence to refactor this would make sense.

bus-mapping/src/operation.rs
roynalnaruto: I think all of those must be typos? Since the function is `reverse`, so the error should also be `reversed` and not `reverted` ?

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
DreamWuGit: I see,  will try to use fields number here, like            ` let rw = &block.rws[(RwTableTag::TxReceipt, (tx.id - 1) * TxReceiptFieldTag::amount() - 1)];`.

bus-mapping/src/circuit_input_builder.rs
DreamWuGit: updated in https://github.com/appliedzkp/zkevm-circuits/pull/466/commits/db106a33e264ff347da007fa2028dfc9ee8d0063 thanks

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
DreamWuGit: updated in https://github.com/appliedzkp/zkevm-circuits/pull/466/commits/db106a33e264ff347da007fa2028dfc9ee8d0063 thanks

bus-mapping/src/circuit_input_builder.rs
DreamWuGit: use  `cumulative_gas_used`  in https://github.com/appliedzkp/zkevm-circuits/pull/466/commits/db106a33e264ff347da007fa2028dfc9ee8d0063 thanks

bus-mapping/src/operation.rs
DreamWuGit: from the  evm/ethereum source code,  `reverted` is much used !

zkevm-circuits/src/evm_circuit/execution/end_tx.rs
```diff
@@ -203,6 +250,26 @@ impl<F: Field> ExecutionGadget<F> for EndTxGadget<F> {
             coinbase_balance,
         )?;
 
+        let current_cumulative_gas_used: u64 = if tx.id == 1 {
+            0
+        } else {
+            let rw = &block.rws[(
```
ed255, line 256: I think this can be simplified if the `cumulative_gas_used` field is stored in the `tx: &Transaction`. See comment above.

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -1524,9 +1524,17 @@ impl<'a> CircuitInputBuilder {
         eth_block: &EthBlock,
         geth_traces: &[eth_types::GethExecTrace],
     ) -> Result<(), Error> {
+        // accumulates gas across all txs in the block
+        let mut cumulative_gas_used = HashMap::new();
```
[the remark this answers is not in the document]
ed255: Here's a better idea: Introduce the field `cumulative_gas_used` in `CircuitInputBuilder::Transaction` which is initialized at 0.  Then in `gen_end_tx_ops` you just need to read the previous value with `state.block.txs[state.tx_ctx.id() - 1].cumulative_gas_used` and assign the new value to `state.tx.cumulative_gas_used`.

Finally you also add `cumulative_gas_used` in `evm_circuit::witness::Transaction`, and then in `EndTxGadget::assign_exec_step` you can just read the value from `block.txs[i].cumulative_gas_used`.

With this change you don't need to read that value in the assign function from the `block.rws`, and thus you can get rid of `TxReceiptFieldTag::iterator` and `TxReceiptFieldTag::amount`.

What do you think?

ed255 (commented):
Everything looks correct!  Nevertheless, please take a look at my comment about the handling of the `cumulative_gas_used` value and let me know what you think about my proposal.

roynalnaruto (commented):
Looks good to me, barring the changes suggested by @ed255 and resolving conflicts!

bus-mapping/src/circuit_input_builder.rs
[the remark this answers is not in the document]
DreamWuGit: @ed255 thanks for the idea!!  I understand `CircuitInputBuilder::Transaction` represents the concept of transaction in ethereum code,  existing fields are its intrinsic property, like `nonce` `gas` `from` `to`  etc.  but the `cumulative_gas_used` is an statistic value not only relate to current tx but also relying on previous txs.  moving it into `Transaction` may violates the raw structure purpose.   
Any thinking ?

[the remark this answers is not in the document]
ed255: The idea of the `Block` and `Transaction` structs found in `CircuitInputBuilder` is to keep all the block and transaction values that the circuits will need as input, and that extends the intrinsic fields.
For example, in `Block` we have `history_hashes` and `container`, in `Transaction` we have `calls` and `steps`.
For structs that match the ethereum data structure we have `eth_types::{Block, Transaction}`.

Since there's a 1 to 1 match between transactions and transaction receipt, I think adding `cumulative_gas_used` to `CircuitInputBuilder::Transaction` (as a field corresponding to the receipt of that transaction) fits; and I believe this simplifies the handling of the `cumulative_gas_used` at `EndTx`.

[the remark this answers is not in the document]
DreamWuGit: agree that it simplifies the handling of the cumulative_gas_used at EndTx.  `call` and `steps` can be derived from single tx,  they can be seen implicitly properties  while `cumulative_gas_used` can not refer by single tx though it is 1:1 mapping.   it can be done in practice, but somewhat different of concept .  let's hear more ideas , i.e. @han0110 what do you think about this?

bus-mapping/src/operation.rs
```diff
@@ -760,6 +762,63 @@ impl CallContextOp {
     }
 }
 
+/// Represents a field parameter of the TxReceipt that can be accessed via EVM
+/// execution.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TxReceiptField {
+    /// flag indicates whether if a tx succeed or not
```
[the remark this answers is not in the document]
DreamWuGit: updated in https://github.com/appliedzkp/zkevm-circuits/pull/466/commits/f130ebd9e8d8f456dd229cd33ea8c00c36606090

bus-mapping/src/circuit_input_builder.rs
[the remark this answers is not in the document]
DreamWuGit: I have a try locally, by adding `cumulative_gas_used` field into  `Transaction`,  we need do following:
1.  handle `CircuitInputBuilder::Transaction` initialization with new field
2.  set `cumulative_gas_used` value in `handle_tx` 
3.  add `cumulative_gas_used`  field as well in `evm_circuit::witness::Transaction ` 
4.  update `evm_circuit::witness::tx_convert` to propagate new field value   

finally , we reach here  :  
```
 // let current_cumulative_gas_used: u64 = if tx.id == 1 {
        //     0
        // } else {
        //     let rw = &block.rws[(
        //         RwTableTag::TxReceipt,
        //         (tx.id - 1) * TxReceiptFieldTag::amount() - 1,
        //     )];
        //     rw.receipt_value()
        // };
// below is using cumulative_gas_used as transaction field way.
let current_cumulative_gas_used: u64 = if tx.id == 1 {
                0
            } else {
                block.txs[tx.id - 2].cumulative_gas_used
        };
```.
it seems to me bring more complexity just for this little place.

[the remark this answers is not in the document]
ed255: > agree that it simplifies the handling of the cumulative_gas_used at EndTx. `calls` and `steps` can be derived from single tx, they can be seen implicitly properties of tx while `cumulative_gas_used` can not refer by single tx though it is 1:1 mapping. it can be done in practice, but somewhat different of concept . let's hear more ideas , i.e. @han0110 what do you think about this?

I understand your concern about `cumulative_gas_used` not being a value derived from a single tx.  In such case, I would say that another option (more code but more correct) is to define a struct `Receipt { cumulative_gas_used }` and add a new field to `CircuitInputBuilder::Block` `receipts: Vec<Receipt>`.  Then in `handle_tx`, an empty `Receipt` is appended to `block.receipts` and it is later set in the `gen_associated_ops` for `EndTx`.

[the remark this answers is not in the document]
ed255: > I have a try locally, by adding `cumulative_gas_used` field into `Transaction`, we need do following:
> 
>     1. handle `CircuitInputBuilder::Transaction` initialization with new field
> 
>     2. set `cumulative_gas_used` value in `handle_tx`
> 
>     3. add `cumulative_gas_used`  field as well in `evm_circuit::witness::Transaction `
> 
>     4. update `evm_circuit::witness::tx_convert` to propagate new field value
> 
> 
> finally , we reach here in `end_tx` :
> 
> ```
>  // let current_cumulative_gas_used: u64 = if tx.id == 1 {
>         //     0
>         // } else {
>         //     let rw = &block.rws[(
>         //         RwTableTag::TxReceipt,
>         //         (tx.id - 1) * TxReceiptFieldTag::amount() - 1,
>         //     )];
>         //     rw.receipt_value()
>         // };
> // below is using cumulative_gas_used as transaction field way.
> let current_cumulative_gas_used: u64 = if tx.id == 1 {
>                 0
>             } else {
>                 block.txs[tx.id - 2].cumulative_gas_used
>         };
> ```
> 
> it seems to me bring more complexity just for this little place.

Bear in mind that by following this approach you can remove the implementation of `TxReceiptFieldTag::iterator` and `TxReceiptFieldTag::amount`; and with that I don't see it being more complex than the current implementation in the PR (I find it more easy to follow).

Nevertheless your current implementation is correct :)  My comment was a proposal: it's OK if after discussing it we don't reach an agreement; and in that case we can proceed with your original implementation

ed255 (approved):
LGTM!

bus-mapping/src/circuit_input_builder.rs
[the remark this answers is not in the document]
ed255: I'm marking this conversation as resolved as this was a proposal and we didn't reach consensus about it :P

PR #469, opened 2022-04-19
fix Execution::assign_block_exact to allow gen vk from empty circuit

lispc:
or else, empty circuit cannot be synthesized

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -607,9 +607,9 @@ impl<F: Field> ExecutionConfig<F> {
                         offset += STEP_HEIGHT;
                     }
                 }
-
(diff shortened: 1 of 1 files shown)


DreamWuGit:
currently `q_step` `q_step_first` `q_step_last` are selectors columns , Can empty circuit assigning the same value with real circuit ? seems not to me ,  correct me if I am wrong.

lispc:
> currently `q_step` `q_step_first` `q_step_last` are selectors columns , Can empty circuit assigning introduce the same value with real circuit ? seems not to me , correct me if I am wrong.

Yes you are correct.. So it seems the correct way for this it to pre-set q_last at a big offset (maybe row_num - blinding_row_num? ) , and fill all rows between last tx to q_last with EndBlock? @han0110 what is your opinion?

PR #475, opened 2022-04-26
Fix bytecode table width

Brechtpd:
The bytecode table width had been increased to `5` but the actual lookup table width was still at `4` making the bytecode lookups ignore the last column in the lookup.

Changed to `zip_eq` while setting the table values so hopefully this mismatch doesn't happen again without throwing an error.


roynalnaruto (approved):
Thanks for noticing and resolving this. Looks good to me!

han0110 (approved):
LGTM!

CPerezz (approved):
LGTM!

Merge when you want @Brechtpd

Brechtpd:
I don't have write access so even if I want I can't merge myself. :)

PR #483, opened 2022-05-02
opcode CODESIZE

roynalnaruto:
Specs: https://github.com/appliedzkp/zkevm-specs/pull/191


bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,37 @@
...
+        let codesize = code.len();
+
+        assert_eq!(codesize, geth_steps[1].stack.last()?.as_usize());
+
+        state.push_stack_op(
+            &mut exec_step,
+            RW::WRITE,
+            geth_step.stack.last_filled().map(|a| a - 1),
+            codesize.into(),
+        )?;
```
lispc, line 33: the new `stack_write` method can be used here.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/ad66311c57680189cf9504012711078d31050f77/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L224

bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        state.stack_write(
+            &mut exec_step,
+            geth_step.stack.last_filled().map(|a| a - 1),
+            codesize.into(),
+        )?;
+
+        Ok(vec![exec_step])
+    }
+}
```
DreamWuGit, line 35: would you to add test for buss mapping code side?

roynalnaruto: @DreamWuGit added

bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,92 @@
...
+        geth_steps: &[GethExecStep],
+    ) -> Result<Vec<ExecStep>, Error> {
+        let geth_step = &geth_steps[0];
+        let mut exec_step = state.new_step(geth_step)?;
+
+        let code_source = state.call()?.code_hash;
+        let code = state.code(code_source)?;
+        let codesize = code.len();
+
+        assert_eq!(codesize, geth_steps[1].stack.last()?.as_usize());
```
lispc, line 25: great!

zkevm-circuits/src/evm_circuit/execution/codesize.rs
```diff
@@ -0,0 +1,110 @@
...
+        let codesize_rlc = RandomLinearCombination::random_linear_combine_expr(
+            [codesize.expr()],
+            cb.power_of_randomness(),
+        );
+
+        cb.require_equal(
+            "Constraint: code size (bytecode table) == stack top",
+            codesize_rlc.expr(),
+            codesize.expr(),
+        );
```
lispc, line 45: since the first argument of `random_linear_combine_expr` is a len-1 array, `random_linear_combine_expr` will become "codesize.expr() * 1".    So codesize_rlc is exact codesize. The "cb.require_equal" can never be false no matter what witness is assigned.

BTW i am a little afraid of the RLC, better to construct a test case with codesize larger than 256?

lispc:
i guess current impl will fail with codesize > 256.

In `bytecode_length` in cb, should we change `query_cell` to `query_word`? So we assign the real bytecode size in codesize::assign_exec_step, in "configure", we can "cb.stack_push(codesize_word.exp()); // word.exp() is RLC underhood".

zkevm-circuits/src/evm_circuit/execution/codesize.rs
roynalnaruto: Yes, you're correct. I missed this point. The PR was open for a while and I can see the mistake after reviewing the code myself now! Will fix this part

bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    use crate::{
+        circuit_input_builder::ExecState,
+        mock::BlockData,
+        operation::{StackOp, RW},
+    };
+
+    #[test]
+    fn codesize_opcode_impl() {
+        let code = bytecode! {
```
icemelon, line 58: add a test case for large code size in the busmapping test as well?

icemelon (approved):
lgtm

zkevm-circuits/src/evm_circuit/execution/codesize.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl<F: Field> ExecutionGadget<F> for CodesizeGadget<F> {
+    const NAME: &'static str = "CODESIZE";
+
+    const EXECUTION_STATE: ExecutionState = ExecutionState::CODESIZE;
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let opcode = cb.query_cell();
+
+        let codesize_bytes = array_init(|_| cb.query_cell());
```
lispc, line 36: why not query_byte? 
We should constrain them really being byte

roynalnaruto: Yes that's right, we do wish to constrain each one to be within a byte.

icemelon:
@ed255 @CPerezz @miha-stopar anyone has the bandwidth to take a look? 😄

bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,107 @@
...
+            op.op(),
+            &StackOp::new(1, StackAddress::from(st_addr), Word::from(codesize))
+        );
+    }
+
+    #[test]
+    fn codesize_opcode_impl() {
+        test_ok(false);
+        test_ok(true);
+    }
```
CPerezz, line 106: I'd probably add one more test which tries to go over the maximum codesize permitted. 
We need to see the behaviour when this happens and, if it's not propperly handled, add support for it.

CPerezz (changes requested):
LGTM overall.
Just want to check all the edge cases. Like going over 24kB for the code.

ed255 (commented):
Overall looks good to me!  But please see my comment on the `Codesize::gen_associated_ops`.

bus-mapping/src/evm/opcodes/codesize.rs
```diff
@@ -0,0 +1,107 @@
...
+
+use eth_types::GethExecStep;
+
+use super::Opcode;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Codesize;
+
+impl Opcode for Codesize {
+    fn gen_associated_ops(
```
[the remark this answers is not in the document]
CPerezz: If we need to implement error handling for `code >24kB` then we cannot reuse `StackOnlyOpcode` no?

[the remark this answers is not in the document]
roynalnaruto: I have handled this in the latest commit :)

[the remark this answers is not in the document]
ed255: I am a bit confused by this error handling.  As far as I understand, the limit on code size is enforced at contract creation (see https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md#specification), so all deployed code must be valid, and thus a `CODESIZE` should never fail because of the actual code size.
So I think this error handling can't be triggered in a normal setup.

Did I miss something?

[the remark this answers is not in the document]
roynalnaruto: @ed255 yes makes sense. After looking more carefully, yes it should be handled at the time of CREATE or CREATE2, and codesize opcode can simply use the stack only opcode method. I'll make the change.

I'll also remove the err and constant I added since it is not appropriate to add along with this PR.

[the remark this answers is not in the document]
roynalnaruto: I have reverted the changes I made for max code size handling. What do you think we can do to keep the assertion? In that case we can't use the `StackOnlyOpcode` for `CODESIZE`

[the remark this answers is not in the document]
ed255: > Although, I would like to have the assertion that the code size from state matches the one from the next step's stack output. In this case, I won't be able to use the `StackOnlyOpcode` right?

I like having assertions like this one!  They can be extremely useful to figure out bugs that we may encounter in the future.  I don't think using the `StackOnlyOpcode` is compatible with having that assertion; and for that I think it's OK to leave the `Codesize::gen_associated_ops` implementation.  So from my side this resolves the conversation :)

ed255 (approved):
LGTM!

bus-mapping/src/evm/opcodes/codesize.rs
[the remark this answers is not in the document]
CPerezz: My bad.. you're right!!!!

CPerezz (approved):
LGTM

PR #500, opened 2022-05-09
Multi-AZ provers

ntampakas:
This PR implements a redundant model with prover instances in multiple Availability Zones.

The idea for this implementation was born when AWS didn` t have enough available capacity to run the prover instance in a specific Availability Zone. With this approach, if a prover cannot be launched for any reason, a prover with similar specs will be launched in a different Availability Zone.


ChihChengLiang (approved):
Untested ack.

CPerezz (approved):
utACK.

Let's merge this and check that no issues happen in the first run! :)

PR #507, opened 2022-05-12
fix log gas by adding static cost

DreamWuGit:
when handling #476 , found that `SameContextGadget::construct` won't add op code constant gas into gas cost now,  this PR adds log static gas in op code for now.! 
while in spec `instruction::step_state_transition_in_same_context ` still sum constant and dynamic gas.  https://github.com/appliedzkp/zkevm-specs/blob/master/src/zkevm_specs/evm/instruction.py#L245


ChihChengLiang (approved):
LGTM

ed255 (approved):
LGTM!

PR #510, opened 2022-05-16
fix multi copy step data index

DreamWuGit:
this PR  can be part of #476 when doing multiple  copy step in buss mapping ,  for `TxLogFieldTag::Data` index should start from zero and consecutively increasing in multi copy log steps.  i.e.  total copies size is 66 bytes, data index should be monotone increasing by 1 from 0 across all three copy steps.  

| step      | data_start_index |copy range |
| ----------- | ----------- |------- |
| CopyStep1    | 0         | 0 --> 31    |
| CopyStep2   | 32       |32 --> 63  |
| CopyStep2   | 64       |64 -> 65    |.   

while old is as below though `Data` value is correct !  
| step     | data_start_index |copy range |
| ----------- | ----------- |------- |
| CopyStep1    | 0       | 0 --> 31    |
| CopyStep2   | 0       | 0 --> 31   |
| CopyStep2   | 0       | 0 --> 31    |.     

found in issue #511

### DIFF
--- a/bus-mapping/src/circuit_input_builder/execution.rs
+++ b/bus-mapping/src/circuit_input_builder/execution.rs
@@ -144,6 +144,7 @@ pub enum CopyDetails {
     Code(U256),
     /// The bytes are being copied to a Log.
     /// Call's state change's persistance and tx_id are provided.
+    /// the data start index when enter this copy step
     Log((bool, usize, usize)),
 }
 
--- a/bus-mapping/src/circuit_input_builder/execution.rs
+++ b/bus-mapping/src/circuit_input_builder/execution.rs
@@ -144,7 +144,7 @@ pub enum CopyDetails {
     Code(U256),
     /// The bytes are being copied to a Log.
     /// Call's state change's persistance and tx_id are provided.
-    Log((bool, usize)),
+    Log((bool, usize, usize)),
 }
 
 /// Auxiliary data of Execution step
--- a/zkevm-circuits/src/evm_circuit/execution/copy_to_log.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/copy_to_log.rs
@@ -337,7 +337,6 @@ pub mod test {
             .zip(buffer.iter().copied())
             .collect();
 
-        //let mut data_start_index = 0;
         let mut copied = 0;
         while copied < length {
             let (step, rw_offset) = make_log_copy_step(
--- a/zkevm-circuits/src/evm_circuit/execution/copy_to_log.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/copy_to_log.rs
@@ -31,6 +31,8 @@ pub(crate) struct CopyToLogGadget<F> {
     is_persistent: Cell<F>,
     // The tx id
     tx_id: Cell<F>,
+    // the log data start index fetched from last step
+    data_start_index: Cell<F>,
     // Buffer reader gadget
     buffer_reader: BufferReaderGadget<F, MAX_COPY_BYTES, N_BYTES_MEMORY_ADDRESS>,
     // The comparison gadget between num bytes copied and bytes_left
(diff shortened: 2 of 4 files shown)


zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -509,9 +509,9 @@ pub enum Rw {
         log_id: u64, // pack this can index together into address?
         field_tag: TxLogFieldTag,
         // topic index (0..4) if field_tag is TxLogFieldTag:Topic
-        // byte index (0..32) if field_tag is TxLogFieldTag:Data
+        // byte index if field_tag is TxLogFieldTag:Data
         // 0 for other field tags
-        index: u8,
+        index: usize,
```
z2trillion, line 514: On ~line 759, can you change 
```
Self::TxLog { log_id, index, .. } => {
     Some((U256::from(*index as u64) + (U256::from(*log_id) << 8)).to_address())
}
``` 
to
```
Self::TxLog { log_id, index, .. } => {
     Some((U256([*index as u64, *log_id, 0, 0]).to_address())
}
```

icemelon (commented):
should add a test case that could cover the new fix code

zkevm-circuits/src/evm_circuit/witness.rs
DreamWuGit: updated  in https://github.com/appliedzkp/zkevm-circuits/pull/510/commits/245e0688f9166fe88c01d03edcc1b6b48fc5fd17 ,however, I think I will refactor this line soon in coming PR as I think field tag should also place here if we consider sorting.

DreamWuGit:
> 



> should add a test case that could cover the new fix code

tests of `copy_to_log_multi_step` cover this,  the incoming log test with multiple steps can indirect check this !

zkevm-circuits/src/evm_circuit/witness.rs
DreamWuGit: realize that change to  `Some(U256([*index as u64, *log_id, 0, 0]).to_address())` will cause lookup error, I prefer to remain it for now as it is not closely related to this PR's target .

.

bus-mapping/src/circuit_input_builder/execution.rs
```diff
@@ -144,7 +144,7 @@ pub enum CopyDetails {
     Code(U256),
     /// The bytes are being copied to a Log.
     /// Call's state change's persistance and tx_id are provided.
-    Log((bool, usize)),
+    Log((bool, usize, usize)),
```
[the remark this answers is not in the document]
ed255: It seems this third parameter is `data_start_index`.  Why not use `StepAuxiliaryData::dst_addr` for that?

zkevm-circuits/src/evm_circuit/execution/copy_to_log.rs
```diff
@@ -68,7 +71,7 @@ impl<F: Field> ExecutionGadget<F> for CopyToLogGadget<F> {
                 cb.tx_log_lookup(
                     tx_id.expr(),
                     TxLogFieldTag::Data,
-                    i.expr(),
+                    data_start_index.expr() + i.expr(),
```
ed255, line 74: I see that in line 90 you do the following constraints for the next step when the copy is not yet finished:
- next src_addr
- next bytes_left
- next src_addr_end

Shouldn't it also constraint that this parameter is the same?:
- next tx_id

And also the next `data_start_index` is?:
- next data_start_index = data_start_index +  MAX_COPY_BYTES

Please, let me know if I missed something.

DreamWuGit: Cool,  added in https://github.com/appliedzkp/zkevm-circuits/pull/510/commits/e592c5e920eae2677e69dae001b66b1f14cee716 and also constrain data start index start with zero in first copy step, thanks for catching !

ed255 (approved):
LGTM! Thanks for the fix :)

icemelon (approved):
LGTM

PR #527, opened 2022-05-23
[keccak] rm state conversion

ChihChengLiang:
After the cleanup in #501, the state conversion now introduces no new gates. 

We remove StateBaseConversion and consolidate the assign_region part to the BaseConversionConfig, so that the codebase can be more concise.


CPerezz (approved):
LGTM!

Thanks for this work!

PR #530, opened 2022-05-24
Give state_circuit create_proof the correct instance to avoid Invalid Instance Error

smtmfft:
state_circuit does not need instance several days ago, now it is failed to pass the check in create_proof().
```rust
    for instance in instances.iter() {
        if instance.len() != pk.vk.cs.num_instance_columns {
            return Err(Error::InvalidInstances);
        }
    }
```
So just give the correct instances to pass the test.
Not sure if it's a good fix. 

A question here is: Now it takes more than 32Gb mem to run a benchmark test of a degree-18 state circuit, not finished as my machine has 32Gb mem only. And ~24Gb and ~6600s to finish a degree-17 test. Are these the correct perf data now? Maybe I should upgrade my machine to 16core/64Gb ? 

Signed-off-by: smtmfft <smtmfft@taikocha.in>

### DIFF
--- a/circuit-benchmarks/src/state_circuit.rs
+++ b/circuit-benchmarks/src/state_circuit.rs
@@ -40,6 +40,8 @@ mod tests {
         // Create a proof
         let mut transcript = Blake2bWrite::<_, _, Challenge255<_>>::init(vec![]);
 
+        let instance = empty_circuit.instance();
+        let instances: Vec<&[Fr]> = instance.iter().map(|v| v.as_slice()).collect();
         // Bench proof generation time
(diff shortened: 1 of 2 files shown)


smtmfft:
Oh, That ridiculous bad perf data is because I used wrong profile, now it's back to reasonable level.

han0110 (approved):
LGTM! Thanks for fixing this!

lispc (approved):

pinkiebell:
@lispc Can this be merged or does it break the circuit benchmarks?

PR #537, opened 2022-05-27
[Keccak] Optimization round: Reduce columns

ChihChengLiang:
By @therealyingtong's advice: We should create columns outside of the configurations instead of inside.

Before optimization:
```
Num advice columns: 102
Num fixed columns: 24
Num instance columns: 2
Num lookups: 6
Num gates: 19
Max advice rows 11087 - lane rotate conversion
num polys: 110
max_gate_degree_poly: 6 - 
```

After optimization:

```
Num advice columns: 31
Num fixed columns: 23
Num instance columns: 2
Num lookups: 6
Num gates: 18
Max advice rows 19042 - Constraint out_state and out_mixing
num polys: 109
max_gate_degree_poly: 6 - 
```
Most column creations happen in mixing (50 columns created) and base conversion (10 columns created). 

Comes with a cost of row increment. Since now rows are stacked together.

Related issue: #489

### DIFF
--- a/keccak256/src/permutation/base_conversion.rs
+++ b/keccak256/src/permutation/base_conversion.rs
@@ -6,7 +6,6 @@ use halo2_proofs::{
 
 use super::tables::BaseInfo;
 use eth_types::Field;
-use itertools::Itertools;
 use std::convert::TryInto;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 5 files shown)


CPerezz (approved):
LGTM!!!!

That's huge news!!! 🎉🎉
Nice job @ChihChengLiang 👏

PR #541, opened 2022-05-29
[Presentation][WIP] Radical keccak

ChihChengLiang:
During working on the idea of simplifying iota. I realized this is a good primitive gate that can replace almost all the non-lookup gates. 

~~`adv__input(next) = adv__input + adv__x * fixed__value`~~
update:
```
adv__input(next) = adv__input + adv__left * adv_right
```
We can do simple addition or subtraction on advice value or fixed value. We can do a simple running sum or even a linear combination of advice columns and fixed columns. All with this gate.

In this PR we replaced the iota, rho, theta, and xi with this gate. Compared to the iota PR, this PR reduced the number of gates from 14 to 9 and num polys from 105 to 51. The cost is that it increases max advice rows from 19042 to 28993.

Together with tag trick in lookup, the value of this might be figuring out what's the thinnest and longest possible shape we can have. Widen it is always easy by just cloning the primitive gate and applying it to different columns. Then we have more room for short-thin tradeoffs to optimize.

In the extreme thin case, 
- this gates takes 2 advice cols and 1 fixed
- base conversions: we want to use as many rows as possible, so they are not tag trick stackable. There are "from 2", "from 9", "from 13", each needs 3 table columns. So 9 fixed columns in total for them.
- tag trick stackable tables are range checks and special chunk lookup. With tag we need 3 fixed columns.
- Need at least 3 advice columns to work with the lookups above.
- 1 fixed column is required by "constrain_constant" function.

so 3 advice columns and 14 fixed columns.

### DIFF
--- a/keccak256/src/permutation.rs
+++ b/keccak256/src/permutation.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::type_complexity)]
 #![allow(clippy::too_many_arguments)]
 pub(crate) mod absorb;
+pub(crate) mod add;
 pub(crate) mod base_conversion;
 pub mod circuit;
 pub(crate) mod iota;
--- a/keccak256/src/permutation.rs
+++ b/keccak256/src/permutation.rs
@@ -4,6 +4,7 @@ pub(crate) mod absorb;
 pub(crate) mod add;
 pub(crate) mod base_conversion;
 pub mod circuit;
+pub(crate) mod flag;
 pub(crate) mod iota;
 pub(crate) mod mixing;
 pub(crate) mod pi;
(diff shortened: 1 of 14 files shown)


ChihChengLiang:
So this is what the advice column minimized looks like. We still have a base conversion missing in the absorbing part, which might create more rows. But log2(64282) is already 15.97..., so a 2^17 circuit can only process 1 keccak. 

```
Num advice columns: 3
Num fixed columns: 17
Num instance columns: 0
Num lookups: 5
Num gates: 2
Max advice rows 64282 - linear combine
num polys: 3
max_gate_degree_poly: 4 - 
```

The picture is long. I'll leave the link here: https://user-images.githubusercontent.com/3391420/171745479-8d0ce41b-4b27-48f0-881e-0e076e68f5ef.png

ChihChengLiang:
superseded by #596

PR #543, opened 2022-05-30
add value_prev to state circuit; make rw table of evm state & state circuit identical

lispc:
background: https://github.com/privacy-scaling-explorations/zkevm-specs/issues/202
spec draft PR https://github.com/privacy-scaling-explorations/zkevm-specs/pull/210

Constrain value_prev when needed: (1) value_prev.cur() == value.prev() when keys not changed, (2) constrain default prev value when keys changed
Change rw value tuple to struct in evm_circuit::constraint_builder. It makes reading codes easier in IDE. Using meaning field names is less likely to introduce bugs. (eg: fix things likevalues.swap(4, 5))
Assignment of rw table of evm circuit and state circuit shard much more codes. In fact rw table of evm circuit and state circuit are identical now.
Change FieldExt to Field in some places. We should always use eth_types::Field.
Use a constant bool generic to control state circuit to be "full" or "light"(disable lexicographic_ordering u16 lookup to reduce k. Since this contraint is almost impossible to be broken if assignment codes of state circuit are called). Use 'light state circuit' in opcodes tests. Reduce CI time 35min -> 17min.

### DIFF
--- a/zkevm-circuits/src/state_circuit.rs
+++ b/zkevm-circuits/src/state_circuit.rs
@@ -13,17 +13,18 @@ use crate::{
         param::N_BYTES_WORD,
         table::RwTableTag,
         util::RandomLinearCombination,
-        witness::{Rw, RwMap, RwRow},
+        witness::{Rw, RwMap},
     },
-    rw_table::RwTable,
 };
 use binary_number::{Chip as BinaryNumberChip, Config as BinaryNumberConfig};
 use constraint_builder::{ConstraintBuilder, Queries};
 use eth_types::{Address, Field, ToLittleEndian};
 use gadgets::is_zero::{IsZeroChip, IsZeroConfig, IsZeroInstruction};
 use halo2_proofs::{
-    circuit::{Layouter, Region, SimpleFloorPlanner},
-    plonk::{Circuit, Column, ConstraintSystem, Error, Expression, Fixed, Instance, VirtualCells},
+    circuit::{Layouter, SimpleFloorPlanner},
+    plonk::{
+        Advice, Circuit, Column, ConstraintSystem, Error, Expression, Fixed, Instance, VirtualCells,
+    },
     poly::Rotation,
 };
 use lexicographic_ordering::{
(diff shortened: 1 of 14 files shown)


zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -323,19 +318,19 @@ impl<F: Field> Queries<F> {
     }
 }
 
-fn rw_to_be_limbs(row: &Rw) -> Vec<u16> {
-    let mut id = row.id().unwrap_or_default() as u32;
+fn rw_to_be_limbs<F: Field>(row: &RwRow<F>) -> Vec<u16> {
```
ed255, line 321: Could you add a TODO here stating that the order from the spec is: `tag, id, address, field_tag, storage_key, rw_counter`?
https://github.com/privacy-scaling-explorations/zkevm-specs/blob/cba1e81ce323a6be8fb0ad512269c5e54549547d/src/zkevm_specs/state.py#L487

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -100,6 +102,40 @@ impl<F: Field> ConstraintBuilder<F> {
...
+        // and they should be connected to MPT cricuit later to check the
+        // `value_prev`.
+        // (2)`TxAccessListAccount` and
+        // `TxAccessListAccountStorage`:  Default values of them should be `false`
+        // indicating "not accessed yet".
+        // (3) `AccountDestructed`: Since we probably
+        // will not support this feature, it is skipped now.
+        // (4) `TxRefund`: Default values should be '0'. BTW it may be moved out of rw table in the future. See https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/395
+        // for more details.
+        self.condition(
```
ed255, line 122: See https://github.com/privacy-scaling-explorations/zkevm-specs/pull/210#discussion_r885479989

ed255 (commented):
I'm pending review of `zkevm-circuits/src/state_circuit.rs`.  Nevertheless I've already left some comments!

zkevm-circuits/src/state_circuit.rs
```diff
@@ -30,75 +31,139 @@ use multiple_precision_integer::{Chip as MpiChip, Config as MpiConfig, Queries a
...
+            );
+        }
+
+        Ok(())
+    }
 }
 
-impl<F: Field> Circuit<F> for StateCircuit<F> {
-    type Config = StateConfig<F>;
+impl<F: Field, const QUICK_CHECK: bool> Circuit<F> for StateCircuitBase<F, QUICK_CHECK> {
```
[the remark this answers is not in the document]
ed255: This comment is actually related to https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/543#discussion_r885486293
I think that if you manage to make `CircuitBase` `pub(crate)` (and other internal structures become `pub(crate)` as well, then `QUICK_CHECK` will not be exposed in any `pub` struct.

ed255:
I've completed an initial review.  In general it looks good and I think everything is correct.  Nevertheless I would like to discuss this:
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/543#discussion_r885628592

My main points are:
1. Can we reduce the constraint complexity (and degree) by applying the constraint to all tags?
2. It seems this constraint may be applied twice for some tags; it would be better if we can avoid this redundancy.

PR #544, opened 2022-05-30
Add 1 more instance row for starting row

z2trillion:
Previously we were short one instance row, which meant that the powers of randomness for the final row in the rw table were all 0.

The `storage_key_rlc` test fails without the fix in the `instance` function.


ed255 (approved):
LGTM!

PR #550, opened 2022-06-02
Add BinaryNumberChip for selecting tags

z2trillion:
Previously the expression for selecting constraints based on the row's tag value was degree 12 (the value of `RwTableTag::COUNT`). This change lowers it to 4, at the cost of 3 more advice columns. This lowers the degree of the constraint system from 18 ->16. These highest degree constraint is lexicographic ordering gadget after this.


zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -291,10 +292,17 @@ impl<F: Field> Queries<F> {
     }
 
     fn tag_matches(&self, tag: RwTableTag) -> Expression<F> {
-        generate_lagrange_base_polynomial(
-            self.tag.clone(),
-            tag as usize,
-            RwTableTag::iter().map(|x| x as usize),
+        and::expr(
```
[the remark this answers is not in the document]
z2trillion: I combined this with a similar method inside `BinaryNumberConfig`.

zkevm-circuits/src/state_circuit/binary_number.rs
```diff
@@ -0,0 +1,172 @@
...
+        let bits = value.as_bits();
+        move |meta: &mut VirtualCells<'_, F>| {
+            Self::value_equals_expr(
+                &bits,
+                &self.bits.map(|bit| meta.query_advice(bit, rotation)),
+            )
+        }
+    }
+
+    pub fn value_equals_expr<F: Field>(
```
ed255, line 85: Could you add these comments to make it easier to follow?
- `expressions` must be binary
- `value` and `expressions` must be the same length
- Returns an expression that evaluates to 1 if `expressions` are equal to `value`.  The expression is degree `n`, where `n` is the bit size of `value`.

zkevm-circuits/src/state_circuit/binary_number.rs
```diff
@@ -0,0 +1,172 @@
...
+        bits
+    }
+}
+
+#[derive(Clone, Copy)]
+pub struct Config<T, const N: usize>
+where
+    T: AsBits<N>,
+{
+    pub bits: [Column<Advice>; N],
```
ed255, line 52: Could you add this comment?
- Must be constrained to be binary for correctness.

zkevm-circuits/src/state_circuit/binary_number.rs
```diff
@@ -0,0 +1,172 @@
...
+            if bit {
+                expression.clone()
+            } else {
+                not::expr(expression.clone())
+            }
+        }))
+    }
+}
+
+pub struct Chip<F: Field, T, const N: usize>
```
ed255, line 99: Could you add some documentation saying what `T` and `N` are?

zkevm-circuits/src/state_circuit/binary_number.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        config
+    }
+
+    pub fn assign(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        value: RwTableTag,
```
ed255, line 155: maybe use `value: &S`, where `S: AsBits<N>` so that this chip can be reused for other tag types?

zkevm-circuits/src/state_circuit/binary_number.rs
```diff
@@ -0,0 +1,172 @@
...
+        move |meta: &mut VirtualCells<'_, F>| {
+            let bits = self.bits.map(|bit| meta.query_advice(bit, rotation));
+            bits.iter()
+                .fold(0.expr(), |result, bit| bit.clone() + result * 2.expr())
+        }
+    }
+
+    pub fn value_equals<F: Field, S: AsBits<N>>(
+        &self,
+        value: &S,
```
ed255, line 73: Is this `S` and `T` from the struct different?

zkevm-circuits/src/state_circuit/binary_number.rs
ed255: Could this function be rewritten so that it takes `value: &T`, so that instead of calling it like:
```rust
BinaryNumberConfig::<RwTableTag, 4>::value_equals_expr(&tag.as_bits(), &self.tag_bits)
```
It can be called
```rust
BinaryNumberConfig::<RwTableTag, 4>::value_equals_expr(&tag, &self.tag_bits)
```

zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -77,7 +87,7 @@ fn test_state_circuit_ok(
 fn degree() {
     let mut meta = ConstraintSystem::<Fr>::default();
     StateCircuit::configure(&mut meta);
-    assert_eq!(meta.degree(), 18);
+    assert_eq!(meta.degree(), 16);
```
ed255, line 90: :partying_face:

zkevm-circuits/src/state_circuit/binary_number.rs
ed255: Suggestion for documentation:
This chip helps working with binary encoding of integers of length `N` bits by:
- Enforcing that the binary representation is in the valid range defined by `T`.
- Creating expressions (via the `Config`) that evaluate to `1` when the bits match a specific value.

ed255 (commented):
Everything looks good!
I've added some comments requesting a bit of documentation, and also discussing some API details.

zkevm-circuits/src/state_circuit/binary_number.rs
z2trillion: This was an oversight. It should just have been `value: &T`.

zkevm-circuits/src/state_circuit/binary_number.rs
z2trillion: Yes, because we need to disallow N-bit patterns that correspond to "invalid" values of T. This constrait is above with the name "binary number value in range".

zkevm-circuits/src/state_circuit/binary_number.rs
z2trillion: After learned some more about the borrow checker, I was able to update the types to do this.

zkevm-circuits/src/state_circuit/binary_number.rs
z2trillion: thanks. I've added these.

zkevm-circuits/src/state_circuit/test.rs
z2trillion: it gets better: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/566/files#diff-854df13b82f490882b640353edbb314afff39d5d02b34c0f0d2ddeea2aab5712

ed255 (approved):
LGTM! Very nice improvement :)

icemelon (approved):
LGTM

PR #558, opened 2022-06-07
Pad StateCircuit to a fixed length

z2trillion:
The number of rows in the state circuit needs to be constant so that the fixed columns that are used as selectors are fixed across all proofs so that the proving and verifying keys won't need to be to be regenerated for each block.

### DIFF
--- a/zkevm-circuits/src/state_circuit.rs
+++ b/zkevm-circuits/src/state_circuit.rs
@@ -35,9 +35,6 @@ use random_linear_combination::{Chip as RlcChip, Config as RlcConfig, Queries as
 use std::collections::HashMap;
 use std::iter::once;
 
-// TODO: use a better value for production.
-const N_ROWS: usize = 1 << 16;
-
 const N_LIMBS_RW_COUNTER: usize = 2;
 const N_LIMBS_ACCOUNT_ADDRESS: usize = 10;
 const N_LIMBS_ID: usize = 2;
@@ -66,14 +63,14 @@ type Lookup<F> = (&'static str, Expression<F>, Expression<F>);
 
 /// State Circuit for proving RwTable is valid
 #[derive(Default)]
-pub struct StateCircuit<F: Field> {
+pub struct StateCircuit<F: Field, const N_ROWS: usize> {
     pub(crate) randomness: F,
     pub(crate) rows: Vec<Rw>,
     #[cfg(test)]
     overrides: HashMap<(test::AdviceColumn, isize), F>,
 }
 
-impl<F: Field> StateCircuit<F> {
+impl<F: Field, const N_ROWS: usize> StateCircuit<F, N_ROWS> {
     /// make a new state circuit from an RwMap
     pub fn new(randomness: F, rw_map: RwMap) -> Self {
         let mut rows: Vec<_> = rw_map.0.into_values().flatten().collect();
@@ -103,7 +100,7 @@ impl<F: Field> StateCircuit<F> {
     }
 }
 
-impl<F: Field> Circuit<F> for StateCircuit<F> {
+impl<F: Field, const N_ROWS: usize> Circuit<F> for StateCircuit<F, N_ROWS> {
     type Config = StateConfig<F>;
     type FloorPlanner = SimpleFloorPlanner;
 
--- a/zkevm-circuits/src/state_circuit.rs
+++ b/zkevm-circuits/src/state_circuit.rs
@@ -35,6 +35,7 @@ use random_linear_combination::{Chip as RlcChip, Config as RlcConfig, Queries as
 use std::collections::HashMap;
 use std::iter::once;
 
+const N_ROWS: usize = 1 << 16;
 const N_LIMBS_RW_COUNTER: usize = 2;
 const N_LIMBS_ACCOUNT_ADDRESS: usize = 10;
 const N_LIMBS_ID: usize = 2;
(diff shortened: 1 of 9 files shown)


z2trillion:
@han0110 can you take a look at this? I think this will be compatible with the plan to check the rw_counter consistency in the EVM circuit. I think we will need to add one additional lookup at the end of the block for a row that equals `Rw::Padding {rw_counter: rw_rows.len()}`.

han0110:
Isn't `Start` already there for padding?

For checking `state.rw_rows.len() == evm.rw_counter`, although we haven't decided which approach we will take (see [here](https://hackmd.io/VCGteyZEQFKUjHA_A72f1Q#Uniqueness-of-rw_counter) for explanation), but I think using the same lookup to do it seems great, because then we don't need to feed a shared public input to them.

If we take this approach, perhaps we can embed the `rw_rows.len()` just in `Start`, and counting down to `0`?

zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -77,7 +78,34 @@ fn test_state_circuit_ok(
 fn degree() {
     let mut meta = ConstraintSystem::<Fr>::default();
     StateCircuit::configure(&mut meta);
-    assert_eq!(meta.degree(), 18);
+    assert_eq!(meta.degree(), 19);
```
lispc, line 81: curious: which constraint increased the degree? or just increase it now since the future rw_counter constaint of Pad  will increase it by one?

z2trillion:
> Isn't `Start` already there for padding?
Good point. I switched over to adding an rw_counter to Rw::Start and using those for padding at the beginning of the table.
> For checking `state.rw_rows.len() == evm.rw_counter`, although we haven't decided which approach we will take (see [here](https://hackmd.io/VCGteyZEQFKUjHA_A72f1Q#Uniqueness-of-rw_counter) for explanation), but I think using the same lookup (or shuffle but need some adjustment) to do it seems great, because then we don't need to feed a shared public input to them.
> 
> If we take this approach, perhaps we can embed the `rw_rows.len()` just in `Start`, and counting down to `0`?
Can you take a look at what I did and see if it's compatible with what you have in mind? Instead of counting down, we count up to the maximum rw_counter (2^32 - 1), so verifying that the first row in the state circuit matches Rw::Start {rw_counter: 2^32 - N_ROWS - self.rows.len()} should be enough to enforce that there are at most `N_ROWS - self.rows.len()` padding rows. Together with the `self.rows.len()` rw lookups in the EVM circuit proving that there are at least `self.rows.len()` non-padding rows, gives us what we need?

han0110:
> Instead of counting down, we count up to the maximum rw_counter (2^32 - 1), so verifying that the first row in the state circuit matches Rw::Start {rw_counter: 2^32 - N_ROWS - self.rows.len()} should be enough to enforce that there are at most N_ROWS - self.rows.len() padding rows. Together with the self.rows.len() rw lookups in the EVM circuit proving that there are at least self.rows.len() non-padding rows, gives us what we need?

You mean `Rw::Start { rw_counter: 2^32 - (N_ROWS - self.rows.len()) }` right? It makes sense to me, but like you said, it only enforces that there are "at most" `N_ROWS - self.rows.len()` padding rows, we can still load more unexpected rws.

Perhaps we need to also check there exists `Rw::Start { rw_counter: 2^32 - 1 }`?

z2trillion:
You're right: as written, the constraints don't prevent there from being fewer Rw::Start rows (which would allow us to maliciously insert unexpected other rows). I don'

I don't think that checking that `Rw::Start { rw_counter: 2^32 - 1 }` is in the table is enough, because we only enforese that the rows are increasing, so we could still skip `rw_counter` values to make space for unexpected rows. If we added a constraint that on `Rw::Start` rows where offset !=0, the `rw_counter` increased by exactly 1, would that prevent this though?

If we added that, would we need the shuffle arguement mentioned in https://hackmd.io/VCGteyZEQFKUjHA_A72f1Q#Uniqueness-of-rw_counter at all?

han0110:
> If we added a constraint that on `Rw::Start` rows where `offset !=0`, the `rw_counter` increased by exactly `1`, would that prevent this though?

Yeah this is also what I thought.

> If we added that, would we need the shuffle arguement mentioned in https://hackmd.io/VCGteyZEQFKUjHA_A72f1Q#Uniqueness-of-rw_counter at all?

It's alwasy required I think, with shuffle we can do rw at any row in EVM circuit. Enforcing these constraint to check no malicious skip/insert in State circuit is just a way to make EVM circuit lazy, we can also do this in EVM circuit (if it's cheaper or easier to realize).

z2trillion:
I've added the rw_counter only changes by 1 constraint. I'm not entirely sure that we'll still need the shuffle argument anymore, but we can talk about this in the future once we are adding to the EVM circuit.

z2trillion:
@ed255 can you take a look at this, please?

zkevm-circuits/src/state_circuit.rs
```diff
@@ -32,6 +32,9 @@ use random_linear_combination::{Chip as RlcChip, Config as RlcConfig, Queries as
 use std::collections::HashMap;
 use std::iter::once;
 
+// TODO: use a better value for production.
+const N_ROWS: usize = 1 << 16;
```
ed255, line 36: Can this constant be moved to the `StateCircuit`?: `pub struct StateCircuit<F: Field, const N_ROWS: usize>`.  So that we can keep tests small and use higher value for production circuits.

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -103,7 +104,14 @@ impl<F: Field> ConstraintBuilder<F> {
     }
 
     fn build_start_constraints(&mut self, q: &Queries<F>) {
-        self.require_zero("rw_counter is 0 for Start", q.rw_counter.value.clone());
+        self.require_zero("field_tag is 0 for Stack", q.field_tag());
+        self.require_zero("address is 0 for Stack", q.address.value.clone());
+        self.require_zero("id is 0 for Stack", q.id());
+        self.require_zero("storage_key is 0 for Stack", q.storage_key.encoded.clone());
```
ed255, line 110: Do you mean `Start` instead of `Stack` in the strings?

ed255 (approved):
LGTM!

ed255:
The approach to constraint that the number of non-Start entries matches with the EVM circuit is very nice!
In the current implementation it would go like this, correct me if I'm wrong:

Prove that there are `n` non-Start entries in the `rw_table`:
1- `lookup(Start, rwc=1)`
2- `lookup(Start, rwc=N_ROWS-n)`
And that's it!  This proves that there are only `N_ROWS-n` Start rows (acting as padding), by looking up the two extremes of the range made with Start where the rwcs are constrained to increment by one on each row.  This means there are exactly `n` non-Start rows.

z2trillion:
Techincally, I think that the two lookups only prove that there are at least `N_ROWS-n` Start rows, since there could be more Start rows after `Start, rwc=N_ROWS-n`. However this isn't actually possible, because the evm circuit also proves that there are at least `n` non-Start entries (and hence at most `N_ROWS-n` Start rows) in the table by doing `n` non-padding lookups with distinct rw_counter's.

Issue #561, opened 2022-06-08
fix state circuit order and packing issue


DreamWuGit:
current state circuit sorts RW as following order :
```
 let mut rows: Vec<_> = rw_map.0.into_values().flatten().collect();
        rows.sort_by_key(|row| {
            (
                row.tag() as u64,
                row.field_tag().unwrap_or_default(),
                row.id().unwrap_or_default(),
                row.address().unwrap_or_default(),
                row.storage_key().unwrap_or_default(),
                row.rw_counter(),
            )
        });
```

correct order should put `field_tag` after `id` & `address` & `storage_key` . current order will encounter errors when RW records consist of multiple `id` or `address` , for example:  
TxLog:  
| tag         | field_tag    |      tx_id |   value | 
| -----      | ---------   |   -----   | -----|
| TxLog    | Address    |    1          |      ..    | 
| TxLog    | Address    |    2         |       ..   | 
| TxLog    | Topic        |    1        |      ..       | 
| TxLog    | Topic        |    1        |      ..       | 
| TxLog    | Topic        |    2        |       ..       | 

while correct layout looks like:  
| tag         | tx_id    |      field_tag | value | 
| -----      | ---------   |   -----   | -----|
| TxLog    |     1               |  Address    |   ..    | 
| TxLog    |     1            |  Topic          |   ..    | 
| TxLog    |     1             |    Topic       |   ..    | 
| TxLog    |     2             |    Address   |   ..    | 
| TxLog    |     2             |    Topic       |   ..    |   

it is the same effect for Ops which contains field like   `id`, `address`  (`TxAccessListAccount`, `TxAccessListAccountStorage`, `Account`, `CallContext` ). 
also need to adjust packing order in https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/state_circuit/lexicographic_ordering.rs#L326  accordingly I think .

PR #552, opened 2022-06-03
implement log state circuit 

DreamWuGit:
change list: 
1. add is_field_tag_unchanged_column in state circuit for check field_tag diff  
2.  add tx log constraint for state part 
3. add part of positive and negative tests 
4. fix  address expression in #581

PR #566, opened 2022-06-13
Use BinaryNumberChip in LexicographicOrderingConfig to lower degree

z2trillion:
fixes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/561


zkevm-circuits/src/state_circuit.rs
```diff
@@ -37,23 +37,27 @@ const N_LIMBS_ACCOUNT_ADDRESS: usize = 10;
...
+}
+
+/// Keys for sorting the rows of the state circuit
+#[derive(Clone, Copy)]
+pub struct SortKeysConfig {
+    tag: BinaryNumberConfig<RwTableTag, 4>,
     id: MpiConfig<u32, N_LIMBS_ID>,
-    is_id_unchanged: IsZeroConfig<F>,
     address: MpiConfig<Address, N_LIMBS_ACCOUNT_ADDRESS>,
     field_tag: Column<Advice>,
```
DreamWuGit, line 58: put `field_tag` behind `storage_key`

z2trillion: why? do you want me to change the order in which the table is sorted too?

DreamWuGit: sorry, I remember wrong after check the latest spec :).

DreamWuGit: sorry, I remember wrong after check the latest spec :).

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -19,17 +20,16 @@ pub struct Queries<F: Field> {
     pub rw_counter: MpiQueries<F, N_LIMBS_RW_COUNTER>,
     pub is_write: Expression<F>,
     pub tag: Expression<F>,
-    pub prev_tag: Expression<F>,
+    pub tag_bits: [Expression<F>; 4],
     pub id: MpiQueries<F, N_LIMBS_ID>,
-    pub is_id_unchanged: Expression<F>,
+    pub is_tag_and_id_unchanged: Expression<F>,
```
DreamWuGit, line 25: we also need `is_tag_unchanged`  and `is_id_unchanged`  `is_field_tag_unchanged`,  I have use them in  PR #552

z2trillion: we can add them later when they're needed.

z2trillion:
@miha-stopar and @ed255 can you take a look at this please?

miha-stopar (commented):
Nice! I like the idea! I would just prefer to have a couple of comments in the code to quickly see what is about. Especially, you need to update the description in `lexicographic_ordering.rs` as now we use RLC and there is still an old description using shifts.

z2trillion:
I left the shift description in the top of the file because I think it's much easier to understand. In my mind, the gadget is showing "one of C0, ..., C31 is non-zero and fits into 16 bits" by using an RLC to avoid overflowing.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -154,13 +154,16 @@ impl Config {
         });
 
         meta.create_gate(
-            "limb_difference is equal to the difference at claimed limb",
+            "limb differences before first_different_limb are all 0",
```
miha-stopar, line 157: I feel the previous description was the right one?

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -154,13 +154,16 @@ impl Config {
...
             |meta| {
                 let selector = meta.query_fixed(selector, Rotation::cur());
                 let cur = Queries::new(meta, keys, Rotation::cur());
                 let prev = Queries::new(meta, keys, Rotation::prev());
                 let limb_difference = meta.query_advice(limb_difference, Rotation::cur());
 
+                // E.g. if first_different_limb = 5, four limb differences before need to be 0.
+                // Using RLC, we check that (cur_1 - prev_1) + r(cur_2 - prev_2) + r^2(cur_3 -
+                // prev_3) + r^3(cur_4 - prev_4) = 0.
```
miha-stopar, line 166: Should be at the previous gate?

miha-stopar:
> I left the shift description in the top of the file because I think it's much easier to understand. In my mind, the gadget is showing "one of C0, ..., C31 is non-zero and fits into 16 bits" by using an RLC to avoid overflowing.

Just thought it might be a bit confusing if docs are different from the implementation. Perhaps leave the shifts in the comment, but mention that RLC is used to lower the degree?

BTW, a naive question: wouldn't be possible to implement the approach with RLC without BinaryNumberChip?

z2trillion:
No it wouldn't be possible because we need to use the BinaryNumberChip to represent `first_different_limb`. Without the BinaryNumberChip, we can't have constraints that are conditional on the value of `first_different_limb`.

miha-stopar:
> No it wouldn't be possible because we need to use the BinaryNumberChip to represent `first_different_limb`. Without the BinaryNumberChip, we can't have low-degree constraints that are conditional on the value of `first_different_limb`.

Of course, got it now. Thanks! Perhaps add a comment that there are two optimisations to reduce the degree: RLC and BinaryNumberChip. The PR looks cool to me! Just one consideration: seems like BinaryNumberChip could be useful elsewhere too, perhaps it could be put in evm_circuit/util?

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -1,220 +1,187 @@
...
             vec![(
-                upper_limb_difference,
+                meta.query_advice(limb_difference, Rotation::cur()),
                 meta.query_fixed(u16_range, Rotation::cur()),
             )]
         });
-        meta.lookup_any(
-            "upper_limb_difference is zero or lower_limb_difference fits into u16",
+
+        meta.create_gate("limb_difference is not zero", |meta| {
```
ed255, line 128: It would be nice to have an IsZeroGadget that doesn't depend on the evm_circuit ConstraintBuilder!

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -348,39 +289,33 @@ fn rw_to_be_limbs(row: &Rw) -> Vec<u16> {
...
-        result.push(partial_sum.clone())
+    let powers_of_randomness = once(1.expr()).chain(powers_of_randomness.into_iter());
+    for ((cur_limb, prev_limb), powers_of_randomness) in cur
+        .be_limbs()
+        .iter()
+        .zip(&prev.be_limbs())
+        .zip(powers_of_randomness)
+    {
+        result.push(partial_sum.clone());
+        partial_sum = partial_sum + powers_of_randomness * (cur_limb.clone() - prev_limb.clone());
```
ed255, line 307:
```suggestion
    for ((cur_limb, prev_limb), power_of_randomness) in cur
        .be_limbs()
        .iter()
        .zip(&prev.be_limbs())
        .zip(powers_of_randomness)
    {
        result.push(partial_sum.clone());
        partial_sum = partial_sum + power_of_randomness * (cur_limb.clone() - prev_limb.clone());
```

Suggestion because inside the iteration you have a particular power of randomness (in singular).

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -348,39 +289,33 @@ fn rw_to_be_limbs(row: &Rw) -> Vec<u16> {
...
-    let mut partial_sum = 0u64.expr();
-    for (cur_limb, prev_limb) in cur.be_limbs()[..15].iter().zip(&prev.be_limbs()[..15]) {
-        partial_sum = partial_sum * (1u64 << 16).expr() + cur_limb.clone() - prev_limb.clone();
-        result.push(partial_sum.clone())
-    }
-    result
-}
-
-fn lower_limb_difference_possible_values<F: Field>(
+fn rlc_limb_differences<F: Field>(
```
ed255, line 292: Note: at index `i`, we have the rlc of limb_differences from `0` to `i-1`.
At index `0` we have `0`.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
ed255: I've unresolved this conversation because I still see the comment, which should go to the previous gate.

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
ed255: Agree, I think the description of this gate and the previous one are swapped.  This one should be:
`limb_difference equals the difference of the limbs at first_different_limb.`

ed255 (commented):
Overal looks good! Very interesting use of the RLC to verify the equality between a large window of values!

I've added some notes and comments with suggestions; all are small details.  Please take a look!

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
z2trillion: i forgot to push :|

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
z2trillion: `IsZeroGadget` doesn't depend on the evm `ConstraintBuilder`, but not using it saves one constraint here.

z2trillion:
> Of course, got it now. Thanks! Perhaps add a comment that there are two optimisations to reduce the degree: RLC and BinaryNumberChip. The PR looks cool to me! Just one consideration: seems like BinaryNumberChip could be useful elsewhere too, perhaps it could be put in evm_circuit/util?

I don't think it makes sense to move it into evm_circuit/util right now, since it's only used in by the state circuit at the moment.

miha-stopar (approved):
LGTM!

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
ed255: Oh I was thinking about this [IsZeroGadget](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/d567bb23fe16eedfb15ff8a67063de201724ae49/zkevm-circuits/src/evm_circuit/util/math_gadget.rs#L20) but I just saw that we also have an `IsZeroChip` under `gadgets`.  I now see the one saved constraint, thanks!

ed255 (approved):
LGTM!

PR #577, opened 2022-06-16
small mixing fix

ChihChengLiang:
`assign_out_mixing_states` already created the merged state. Returned it directly is the correct implementation.


CPerezz (approved):
As said the other day, really good catch! At some point I think git re-introduced this or something as it was fixed for some time already.

Thanks a lot @ChihChengLiang

PR #580, opened 2022-06-20
add boolean flag lookup

ChihChengLiang:
We reuse the stackable table for the boolean flag lookup.
Cost: 1. Adds two rows to the stackable. 2. Need a tag advice cell when doing the lookup.
Benefit: don't need the [x(x-1), x+y=1] gate.

TODO: replace the whole mixing config with this.


CPerezz (approved):
I really like it! Thanks @ChihChengLiang for the work!!!

Issue #598, opened 2022-07-04
evm_circuit/RestoreContextGadget: index out of bounds


pinkiebell:
After running cargo test w/ prover enabled on latest zkevm-chain I encounter:
`thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 3 but the index is 3', /zkevm-circuits/zkevm-circuits/src/evm_circuit/util/common_gadget.rs:196:21`

Here is a standard block trace for that block: https://bafybeidmdglhpbv7fkvslh7sh4mpjqtwl2k4lagkkogqaesuiwacgvoeri.ipfs.infura-ipfs.io/TRACE

pinkiebell:
this a side effect of using STOP for unimplemented opcodes. Sorry for the noise!

pinkiebell:
> > this is a side effect of using STOP for unimplemented opcodes.
> 
> Ah sorry I didn't realize that `STOP` is used for unimplemented opcodes somewhere else, I thought I already replaced all of that with `RETURN`, would you like to share which part causes this panic? Or it's just modification locally?

A local modification here: https://github.com/pinkiebell/zkevm-circuits/commit/d535c52c4d2a8f7289ac38c6db821c161777531f

han0110:
Ah I see, replacing it with `RETURN` should fix this.

pinkiebell:
Does it make sense to implement a DummyGadget for them instead of using RETURN?

han0110:
Yeah it makes totally sense, then we don't need to update that again if `RETURN` is implemented someday.

`DummyGadget` is actually implemented [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/dummy.rs), but seems not "dummy" enough for all unimplemented cases. @lispc Do you think it's okay to remove those stack operation checks and make `DummyGadget` for all unimplemented states usage? Or there is any thing depending on this check so it's necessary?

pinkiebell:
I'm adding the missing opcodes at the moment :)

pinkiebell:
I'm getting an error for `EXTCODESIZE`: `index out of bounds: the len is 0 but the index is 0`
where `extcodesize_gadget: DummyGadget<F, 1, 1, { ExecutionState::EXTCODESIZE }>,` .

Could this error originate from somewhere else higher up the callchain?

han0110:
[...]

PR #599, opened 2022-07-04
evm: add dummy gadgets for remaining opcodes

pinkiebell:
ref: #598


zkevm-circuits/src/evm_circuit/execution/dummy.rs
```diff
@@ -54,6 +54,14 @@ impl<F: Field, const N_POP: usize, const N_PUSH: usize, const S: ExecutionState>
         _: &Call,
         step: &ExecStep,
     ) -> Result<(), Error> {
+        // this happens if some opcodes are in the
+        // process of being implemented but are still
+        // using DummyGadget.
+        // See `bus-mapping/src/evm/opcodes.rs`
+        if step.rw_indices.len() != N_POP + N_PUSH {
+            return Ok(());
+        }
```
han0110, line 63: Perhaps we could just panic here instead of silent?

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -1251,6 +1251,25 @@ impl From<&circuit_input_builder::ExecStep> for ExecutionState {
...
+                    OpcodeId::EXTCODESIZE => ExecutionState::EXTCODESIZE,
+                    OpcodeId::EXTCODECOPY => ExecutionState::EXTCODECOPY,
+                    OpcodeId::RETURNDATASIZE => ExecutionState::RETURNDATASIZE,
+                    OpcodeId::RETURNDATACOPY => ExecutionState::RETURNDATACOPY,
+                    OpcodeId::CREATE => ExecutionState::CREATE,
+                    OpcodeId::CALLCODE => ExecutionState::CALLCODE,
+                    OpcodeId::DELEGATECALL => ExecutionState::DELEGATECALL,
+                    OpcodeId::CREATE2 => ExecutionState::CREATE2,
+                    OpcodeId::STATICCALL => ExecutionState::STATICCALL,
+                    OpcodeId::SELFDESTRUCT => ExecutionState::SELFDESTRUCT,
```
han0110, line 1272: Seems `REVERT` is missing.

zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -1251,6 +1251,25 @@ impl From<&circuit_input_builder::ExecStep> for ExecutionState {
                     OpcodeId::CALLDATALOAD => ExecutionState::CALLDATALOAD,
                     OpcodeId::CODESIZE => ExecutionState::CODESIZE,
                     OpcodeId::RETURN | OpcodeId::REVERT => ExecutionState::RETURN,
+                    // dummy ops
+                    OpcodeId::ADDRESS => ExecutionState::ADDRESS,
```
ed255, line 1255: I would ask to add a `warn` for each of those indicating the particular ExecutionState and mentioning that it's using a dummy gadget.
Like this:
```rust
                    OpcodeId::SHA3 => {
                        log::warn!("ExecutionState::SHA3 is implemented with DummyGadget");
                        ExecutionState::SHA3
                    }
```
And please do it as well for SHA3, which is currently using the dummy gadget.

I'm aware that there's a warning in the configure of the dummy gadget, but unfortunately this warning shows up even if the execution state doesn't appear in the trace used for the proof.  On the other hand, having a warning here will only appear if the trace uses that opcode; so I think that's preferable.

zkevm-circuits/src/evm_circuit/execution/dummy.rs
pinkiebell: this one sidesteps the dummy implementations for *CALL and other ones. Otherwise we reach `unreachable` code. We can add a log warning if that happens

zkevm-circuits/src/evm_circuit/witness.rs
pinkiebell: it hides here 🙂 
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/599/files/c907e35cf49f4836c45faea58bff62826920648b#diff-09bf944a4f9216c1053a271fea0ff2f7a985380789b990f8e4ac530159214df2R1253

lispc (approved):
great!

zkevm-circuits/src/evm_circuit/witness.rs
han0110: I see, thanks!

zkevm-circuits/src/evm_circuit/execution/dummy.rs
han0110: I see, sounds good to me.

han0110 (approved):

PR #606, opened 2022-07-06
bus mapping:  refactor begin&end tx stop with read/write ops

DreamWuGit:
Just found and update `begin_tx` `end_tx` `stop` with read/write ops style


ed255 (approved):
LGTM! Thanks for the update :)

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -260,7 +261,7 @@ impl<'a> CircuitInputStateRef<'a> {
 
     /// Push a read type [`AccountOp`] into the
     /// [`OperationContainer`](crate::operation::OperationContainer) with the
-    /// next [`RWCounter`](crate::operation::RWCounter)  and `call_id`, and then
```
z2trillion, line 263: good catch!

z2trillion (approved):

PR #620, opened 2022-07-14
Set initial TxReceipt accesses to be writes instead of reads

z2trillion:
Also adds cumulative gas used to the block context instead of storing it as a map.


bus-mapping/src/evm/opcodes.rs
```diff
@@ -438,43 +434,39 @@ pub fn gen_end_tx_ops(
     )?;
 
     // handle tx receipt tag
-    state.tx_receipt_read(
+    state.tx_receipt_write(
```
DreamWuGit, line 437: if change to write type, pls. update spec accordingly.

lispc (approved):

bus-mapping/src/evm/opcodes.rs
z2trillion: can you take a look at https://github.com/privacy-scaling-explorations/zkevm-specs/pull/234 please? It won't let me add you as a reviewer.

z2trillion:
Waiting to merge https://github.com/privacy-scaling-explorations/zkevm-specs/pull/234 before this.

bus-mapping/src/evm/opcodes.rs
DreamWuGit: will have a look !

DreamWuGit: approved the spec change.

DreamWuGit (approved):
LGTM!

Issue #607, opened 2022-07-07
State circuit - MPT circuit integration


miha-stopar:
Besides lookups that are enabled in the MPT, I think there are two other things need to be considered.

1. Address // Key

Address or key gets hashed before it’s inserted into the trie. So the position of the node in the trie is determined by the hashed value.

For example, let the address be `B0 … B31`. Let’s say:
`hash(B0 … B31) = b0 … b31`

If the node is stored in the second level, we split `b0` into nibbles: `b0 = n1 * 16 + n2`, the position
of the node in the first branch is `n1`, the position of the node in the second level is `n2`.

MPT circuit checks the node is stored at the proper position, so it needs the values `b0 … b31` (and not `B0 … B31`).

The address RLC that is used for MPT lookup thus needs to be `b0 + b1*r + … + b31*r^31`. That means
state circuit needs to verify `hash(B0 … B31) = b0 … b31` using Keccak lookup and then use `b0 + b1*r + … + b31*r^31` for MPT lookup.

2. Randomness for RLC

When the randomness `r` is computed, the commitments to the MPT advice columns need to be in the transcript because MPT advice columns contain data that we didn’t commit to in other circuits. Implicitly we committed to all the data that determines the getProof data (determined by address/key and value), but it feels like if the randomness `r` does not depend directly on getProof data it’s possible to do the following for example: we have a branch `b0 … b140` and we should check `keccak_lookup(b0 + … b140*r^140, hash(branch))`, but we can try to find another stream of bytes `a0 … a140` such that `b0 + … b140*r^140 = a0 + … a140*r^140`, because we didn’t commit explicitly to `b0 … b140` and pushed this commitment into the transcript. But I haven’t followed the discussion about randomness too closely so maybe these kind of things have already been discussed.

Any thoughts on these two points?

z2trillion:
What lookups are enabled in the MPT circuit at the moment?

Regarding 1:
The MPT lookups the state circuit does from https://github.com/privacy-scaling-explorations/zkevm-specs/issues/217#issuecomment-1170338113 look like
```
MptAccountStorageLookup {address, key, commited_value, value,  old_root, new_root}
```
and 
```
MptAccountLookup {address, field_tag, commited_value, value,  old_root, new_root}
```
Are you proposing that instead they should look
```
MptAccountStorageLookup {keccak(address), keccak(key), commited_value, value,  old_root, new_root}
```
and 
```
[...]

PR #621, opened 2022-07-14
Add mpt lookups to state circuit

z2trillion:
Fixes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/607


ed255 (commented):
I've taken an initial look and everything looks correct.  I would like to review the spec part once it's ready and come back for a second look before approving.

On a second note, I'm a bit worried about debuggability in the lookups, since now we have the lookup expression being a single `Expression<F>`, and the lookup destination being a single column, where both side are RLC encoded.  So on failure, it's hard to figure out the original values of each column out of the RLC encoded value.  Does anyone share this worry?

On a high level I wonder if it would be possible to modify halo2 such that the lookups are kept as before (with all their original columns), and the RLC is performed by some halo2 functionality automatically (maybe only when working with the real prover).

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -129,6 +133,16 @@ impl<F: Field> ConstraintBuilder<F> {
                 q.initial_value.clone() - q.initial_value_prev(),
             );
         });
+
+        self.add_lookup(
+            "if state_root changes, mpt_update exists in mpt circuit",
+            (
+                q.lexicographic_ordering_selector.clone()
+                    * (q.state_root() - q.state_root_prev())
```
noel2004, line 141: In my opinion, the lookup to mpt table should not be triggered by change of state_root, they should be triggered by some intrinsic rules (`Tag` is `AccountStorage` and it is the last row for a specified account / key, etc.). Following is the reasons:

1. the `state_root` col would be assigned arbitrarily in state circuit and there is no other way to provde them. For example we can assign identical value in `state_root` and all lookups are disabled.

2. Some entries are required in mpt table even it do not change state root. Consider a sole `SLOAD` to specified key or a `SELFBALANCE` op. We must provide it in mpt circuit but the state root after this op would never change.

z2trillion: Thanks for catching this. I have changed the constraints to be what you suggested.

z2trillion:
spec pr: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/236

zkevm-circuits/src/state_circuit/constraint_builder.rs
noel2004: Great. I think now it is ok.

ed255:
@z2trillion will this PR resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/607 ?

z2trillion:
yeah

Issue #600, opened 2022-07-04
Define a circuit that contains all circuits


ed255:
For the [first milestone](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/277) we will try to put all the circuits together and verify them under a single aggregation circuit (in order to make verification viable in a smart contract).

Currently we have all the circuits defined as independent circuits, where each circuit has copies of the tables used between circuits.  Such tables may have small differences and the assignments between circuits may not be 100% equal.

By having a super circuit that contains all circuits and contains each table only once, we can start finding these small bugs and prepare for a full zkEVM circuit for the first milestone.

I believe scroll-tech has started working on this already via this branch https://github.com/privacy-scaling-explorations/zkevm-circuits/compare/main...scroll-tech:zkevm-circuits:scroll-dev-0607  My impression is that their focus is on matching the rw_table used in EVM Circuit and State Circuit.

ed255:
I have started working on integrating the EVM Circuit with the Tx Circuit (tx table), and Bytecode Circuit (bytecode table).  My progress is in this branch: https://github.com/privacy-scaling-explorations/zkevm-circuits/tree/feature/super-circuit

An initial test with a Super Circuit containing the EVM Circuit and Tx Circuit leads to long testing times.  I used the MockProver with verify all rows and a test took 22 minutes on my laptop.  To accelerate tests I'm using this branch of halo2: https://github.com/ed255/halo2/tree/feature/dev-tooling which is built on top of https://github.com/privacy-scaling-explorations/halo2/compare/main...scroll-tech:halo2:scroll-dev-0607 from scroll-tech.  The main benefit of this branch is using the parallel verification in the MockProver from [spherel](https://github.com/spherel)

PR #622, opened 2022-07-15
SuperCircuit

ed255:
The SuperCircuit is a new circuit that will contain all the circuits of the zkEVM in order to achieve two things:
1. Check the correct integration between circuits via the shared lookup tables
2. Allow having a single circuit setup for which a proof can be generated that would be verified under a single aggregation circuit for the [first milestone](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/277)

The name "SuperCircuit" is open for discussion, other name proposals are welcome!

Adding the State Circuit into the SuperCircuit will be done after the work being done by scroll-tech to integrate the EVM circuit with the State circuit is merged (See https://github.com/privacy-scaling-explorations/zkevm-circuits/compare/main...scroll-tech:zkevm-circuits:scroll-dev-0607)

NOTE: The super circuit test is disabled from the CI because it requires more memory than is available in the github workers.

Summary of changes
- general:
  - Addapt the circuits so that the shared tables can be assigned outside of
    the circuit's assign functions.

- evm_circuit.rs
  - load table functions moved to `table.rs`

- bytecode_circuit/bytecode_unroller.rs
  - Rename `hash_length` to `code_length` in `Config`
  - Rename `r` to `randomness`
  - Add `is_enabled` expression to the keccak lookup
  - Move `load_keccaks` to `table.rs`
  - Fix endianness in keccak hash output
  - Add function to obtain all the keccak inputs required by the `SignVerifyChip`

- table.rs
  - Move table definitions from `evm_circuit/table.rs` and `rw_table.rs` to here
  - Move all shared table load functions to here
  - Add BytecodeTable to help reusing the table config with other
  - Rename `hash` to `code_hash` in `BytecodeTable`
  - Use `RLC(reversed(input))` in keccak table assignment for `input_rlc`.

- tx_circuit.rs
  - Add function to obtain all the keccak inputs required by the `TxCircuit`
  - Fix `Gas` encoding: remove RLC
  - Add missing field `TxFieldTag::CallDataGasCost`

- tx_circuit/sign_verify.rs
  - Move `load_keccak` to `table.rs`
  - Add function to obtain all the keccak inputs required by the `SignVerifyChip`
  - Reverse the `pk_hash` used in the lookup so that the RLC is calculated as if
    it was a hash in an Ethereum Word.

The SuperCircuit contains the following circuits:

- [x] EVM Circuit
- [ ] State Circuit
- [x] Tx Circuit
- [x] Bytecode Circuit
- [ ] Copy Circuit
- [ ] Keccak Circuit
- [ ] MPT Circuit
- [ ] PublicInputs Circuit

And the following shared tables, with the circuits that use them:

- [x] Copy Table
  - [ ] Copy Circuit
  - [x] EVM Circuit
- [ ] Rw Table
  - [ ] State Circuit
  - [ ] EVM Circuit
  - [ ] Copy Circuit
- [x] Tx Table
  - [x] Tx Circuit
  - [x] EVM Circuit
  - [ ] Copy Circuit
  - [ ] PublicInputs Circuit
- [x] Bytecode Table
  - [x] Bytecode Circuit
  - [x] EVM Circuit
  - [ ] Copy Circuit
- [x] Block Table
  - [x] EVM Circuit
  - [ ] PublicInputs Circuit
- [ ] MPT Table
  - [ ] MPT Circuit
  - [ ] State Circuit
- [x] Keccak Table
  - [ ] Keccak Circuit
  - [ ] EVM Circuit
  - [x] Bytecode Circuit
  - [x] Tx Circuit
  - [ ] MPT Circuit

Diagram of the Super Circuit (in green are the included components for this PR):
image

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/600


zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -716,27 +680,45 @@ mod tests {
...
-    fn verify<F: Field>(k: u32, bytecodes: Vec<UnrolledBytecode<F>>, success: bool) {
+    fn verify<F: Field>(k: u32, bytecodes: Vec<UnrolledBytecode<F>>, randomness: F, success: bool) {
         let circuit = MyCircuit::<F> {
             bytecodes,
             size: 2usize.pow(k),
+            randomness,
         };
 
-        let prover = MockProver::<F>::run(k, &circuit, vec![]).unwrap();
+        let instance = vec![vec![randomness; (1 << k) - 7 + 1]];
```
lispc, line 720: 7 is the degree? (Maybe either add some comment or use cs.degree() will  be better)

lispc:
the general idea looks good to me. 

Some questions:

(1) in [scroll-dev-0714](https://github.com/scroll-tech/zkevm-circuits/pull/176), i have already make rw table of evm circuit and state circuit identical. (implicitly, not a explicit "RwTable" struct). Will it be good to add it also into this branch? But i may submit no earlier than next Tuesday. 
(2) this pr may bring a lot of conflict, eg: [state circuit mpt](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/621) pr. Any idea on the merging order?
(3) `power_of_randomness_from_instance`: (a) i am not sure whether it should be instance cols now, anyway finally it should be cs.challege().     (Another thought is that i find it be useful to allow a fixed Fr to be fake "randomness" for debugging, better for reproducibility rather than real Fr::rand().  In my own branch, I use power of 0x1000 as randomness [Expr; 31] in daily development.  )

zkevm-circuits/src/table.rs
```diff
@@ -0,0 +1,874 @@
...
+use keccak256::plain::Keccak;
+use strum_macros::{EnumCount, EnumIter};
+
+/// Trait used for dynamic tables.  Used to get an automatic implementation of
+/// the LookupTable trait where each `table_expr` is a query to each column at
+/// `Rotation::cur`.
+pub trait TableColumns {
+    /// Returns the list of advice columns following the table order.
+    fn columns(&self) -> Vec<Column<Advice>>;
+}
```
han0110, line 29: Perhaps we can give it a more specific name like `DynamicTableColumns`? Otherwise it looks similar to builtin `TableColumn`.

zkevm-circuits/src/table.rs
```diff
@@ -0,0 +1,874 @@
...
+    /// Decrementing counter denoting reverse read-write counter.
+    pub rwc_inc_left: Column<Advice>,
+    /// Binary chip to constrain the copy table conditionally depending on the
+    /// current row's tag, whether it is Bytecode, Memory, TxCalldata or
+    /// TxLog.
+    pub tag: BinaryNumberConfig<CopyDataType, 3>,
+}
+
+impl CopyTable {
+    /// Construct a new KeccakTable
```
han0110, line 741:
```suggestion
    /// Construct a new CopyTable
```

zkevm-circuits/src/table.rs
```diff
@@ -0,0 +1,874 @@
...
+            tag: BinaryNumberChip::configure(meta, q_enable),
+            addr: meta.advice_column(),
+            src_addr_end: meta.advice_column(),
+            bytes_left: meta.advice_column(),
+            rw_counter: meta.advice_column(),
+            rwc_inc_left: meta.advice_column(),
+        }
+    }
+
+    /// Generate the keccak table assignments from a byte array input.
```
han0110, line 755:
```suggestion
    /// Generate the copy table assignments from a copy event.
```

han0110 (commented):
Everything looks great! With structured tables It makes things much easier to follow! Only left some small questions and suggestions.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -212,17 +202,20 @@ impl<F: Field> Config<F> {
...
+            // FIXME: Since randomness is only known at synthesis time, the RLC of empty
+            // code_hash is not constant.  Consider doing a lookup to the empty code_hash
+            // value? cb.condition(length_is_zero.clone().is_zero_expression,
+            // |cb| {     cb.require_equal(
+            //         "if length == 0: code_hash == RLC(EMPTY_HASH, randomness)",
+            //         meta.query_advice(bytecode_table.code_hash, Rotation::cur()),
+            //         Expression::Constant(keccak(&[], randomness)),
+            //     );
+            // });
+
```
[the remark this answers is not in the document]
ed255: Ah, but it's not that easy!  This gate needs to be redesigned.  Currently this gate requires the randomness to be constant, as it expects the `RLC(keccak([]))` to be a constant expression, but this won't be the case, because the randomness is different for each proof.

So instead we need to hardcode the bytes of `keccak([])` and then build the RLC expression.  Actually, after explaining the solution it sounds easy to do.  I think I can update the PR with this.

zkevm-circuits/src/super_circuit.rs
```diff
@@ -0,0 +1,411 @@
...
+        instance.push(vec![]);
+
+        let chain_id = block.context.chain_id;
+        let tx_circuit = TxCircuit::new(aux_generator, block.randomness, chain_id.as_u64(), txs);
+        let circuit = SuperCircuit::<F, MAX_TXS, MAX_CALLDATA> {
+            block,
+            fixed_table_tags,
+            tx_circuit,
+            bytecode_size: bytecodes_len + 64,
+        };
```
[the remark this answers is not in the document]
ed255: Right, originally I had `1 << k` here, so that the bytecode region occupies the entire circuit; but the other circuits require a big k, which ends up being `k = 19`.  And with `2^19` enabled rows in the bytecode circuit, the mock prover takes 6h for a test run!
So that's why I decided to use enough rows to hold the bytecode in the test, plus some extra rows so that we also check the constraints for padding rows in the circuit.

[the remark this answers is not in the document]
han0110: I see! It makes sense now.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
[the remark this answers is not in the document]
han0110: Ah right, it's not that simple to just uncomment, the approach you describe sounds good to me, we can also leave it until challenge API is ready.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
ed255: I've added a variable and a constant that show what those values are in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -716,27 +680,45 @@ mod tests {
...
 
         fn configure(meta: &mut ConstraintSystem<F>) -> Self::Config {
-            Config::configure(meta, MyCircuit::r())
+            let bytecode_table = BytecodeTable {
+                code_hash: meta.advice_column(),
+                tag: meta.advice_column(),
+                index: meta.advice_column(),
+                is_code: meta.advice_column(),
+                value: meta.advice_column(),
+            };
```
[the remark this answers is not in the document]
ed255: Done in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -426,15 +253,23 @@ pub mod test {
...
+            config
+                .tx_table
+                .load(&mut layouter, &self.block.txs, self.block.randomness)?;
+            config
+                .rw_table
+                .load(&mut layouter, &self.block.rws, self.block.randomness)?;
+            config.bytecode_table.load(
                 &mut layouter,
-                self.block.bytecodes.values(),
+                self.block.bytecodes.iter().map(|(_, b)| b),
```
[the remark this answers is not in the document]
ed255: you're right! Done in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

zkevm-circuits/src/table.rs
ed255: Thanks for the suggestion! Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

zkevm-circuits/src/table.rs
ed255: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

zkevm-circuits/src/table.rs
ed255: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/622/commits/7955480c55c5be39fc7efb1710ca2dbf58e34ec8

lispc:
ok. for the rlc, i can discuss it later ( we currently use AggregationCircuit to verify multi small circuits, and check some advice column commitments are equal between small circuits, so using rlc can reduce num of cols, reducing the cost of AggregationCircuit. I may adjust the  "RLC"-ness of shared table between circuits configurable, or at least, merge friendly later)

lispc (approved):

han0110 (approved):

PR #627, opened 2022-07-19
fix integration test 

lispc:
vec\<u8\> should always be valid bytecode. So it will also be more consistent with BytecodeCircuit and geth


roynalnaruto (approved):
Looks good!

PR #614, opened 2022-07-11
reconstruct memory when set EnableMemory to false in tracing

lightsing:
This pull request implements issue described in https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/502.

It implements memory reconstruction when external tracer traced with `EnableMemory: false`.

PR #635, opened 2022-07-26
SHA3 opcode

### LINKED ISSUE #574: `SHA3` opcode
Implement the SHA3 execution state for the SHA3 opcode.

This opcode uses dynamic length memory copy so it may depend on the Copy Circuit.

Depends on spec: https://github.com/privacy-scaling-explorations/zkevm-specs/issues/223

roynalnaruto:
This PR closes #574 

- [x] Specs: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/235
- [x] Bus-Mapping
    - [x] Implementation
    - [x] Testing
- [x] Copy Table Lookup (Update)
- [x] Copy Circuit (Update)
- [x] Keccak Lookup
- [x] SHA3 Gadget

### DIFF
--- a/bus-mapping/src/evm/opcodes/sha3.rs
+++ b/bus-mapping/src/evm/opcodes/sha3.rs
@@ -54,8 +54,7 @@ impl Opcode for Sha3 {
         // Memory read operations
         let mut steps = Vec::with_capacity(2 * size.as_usize());
         for (i, byte) in memory.iter().enumerate() {
-            let rwc = state.block_ctx.rwc;
-            state.memory_read(&mut exec_step, (offset.as_usize() + i).into(), *byte)?;
+            // Read step
             steps.push(CopyStep {
                 addr: offset.as_u64() + (i as u64),
                 tag: CopyDataType::Memory,
@@ -63,9 +62,11 @@ impl Opcode for Sha3 {
                 value: *byte,
                 is_code: None,
                 is_pad: false,
-                rwc,
+                rwc: state.block_ctx.rwc,
                 rwc_inc_left: 0,
             });
+            state.memory_read(&mut exec_step, (offset.as_usize() + i).into(), *byte)?;
+            // Write step
             steps.push(CopyStep {
                 addr: i as u64,
                 tag: CopyDataType::RlcAcc,
@@ -73,7 +74,7 @@ impl Opcode for Sha3 {
                 value: *byte,
                 is_code: None,
                 is_pad: false,
-                rwc,
+                rwc: state.block_ctx.rwc,
                 rwc_inc_left: 0,
             })
         }
@@ -113,9 +114,9 @@ mod sha3_tests {
     use rand::{random, Rng};
 
     use crate::{
-        circuit_input_builder::ExecState,
+        circuit_input_builder::{CopyDataType, CopyStep, ExecState},
         mock::BlockData,
-        operation::{StackOp, RW},
+        operation::{MemoryOp, RWCounter, StackOp, RW},
     };
 
     enum MemoryKind {
@@ -156,6 +157,7 @@ mod sha3_tests {
             code.push(32, (32 * i).into());
             code.write_op(OpcodeId::MSTORE);
         }
+        let memory_len = memory.len();
 
         // append SHA3 related opcodes at the tail end.
         let code_tail = bytecode! {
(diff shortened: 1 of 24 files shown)


bus-mapping/src/evm/opcodes/sha3.rs
```diff
@@ -1,33 +1,294 @@
...
             state
                 .call_ctx_mut()?
                 .memory
-                .extend_at_least(offset + length);
+                .extend_at_least(offset.as_usize() + size.as_usize());
+        }
+
+        let memory = geth_step
+            .memory
+            .read_chunk(offset.as_usize().into(), size.as_usize().into());
```
lispc, line 43: we may disable memory trace later so step.memory will be empty( https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/614 ) .   https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/a86aa1d32d58dc73fc5d7d5670636f1cc0edfc92/bus-mapping/src/evm/opcodes/logs.rs#L143 memory can be read from call_ctx like this.

zkevm-circuits/src/evm_circuit/execution/sha3.rs
```diff
@@ -0,0 +1,201 @@
...
+    const EXECUTION_STATE: ExecutionState = ExecutionState::SHA3;
+
+    const NAME: &'static str = "SHA3";
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let opcode = cb.query_cell();
+
+        let offset = cb.query_cell();
+        let size = cb.query_rlc();
+        let sha3_rlc = cb.query_word();
```
lispc, line 41: just curious: query_word is same as query_rlc. So line 40 and line 41 is there any indication by using `query_rlc` and `query_word`?

roynalnaruto: There is no such difference. I used `query_word` for SHA3's RLC value because it is a 32-bytes value pushed to the stack, and I used `query_rlc` for the size because even though its coming from the stack, the `MemoryAddressGadget` later constraints it to be within 5-bytes. So the differentiation is simply for me to note that one is 32-bytes while the other is not. I could use `query_rlc` for `sha3_rlc` and it shouldn't be a problem :)

lispc:
general good.

Two questions:

(1) i mentioned "let key = (tx.id, call.id, step.program_counter as usize);" may be not uniq enough. 
(2) in future, will keccak table only be from copy table ? If not, we can add a comment above keccak_table.load" indicating in future we may also need to fill keccak table from other sources besides copy table

lispc (approved):

roynalnaruto:
@lispc 

> (1) i mentioned "let key = (tx.id, call.id, step.program_counter as usize);" may be not uniq enough.

Yes, you're right. As per your commit, using `gas_left` instead of `program_counter` should be the correct approach. Since `gas_left` is monotonically decreasing, while `program_counter` could be repeated within a loop.

> (2) in future, will keccak table only be from copy table ? If not, we can add a comment above "keccak_table.load" indicating in future we may also need to fill keccak table from other sources besides copy table

Keccak table is not tied only to the copy table. Any other circuit can also use it, like for eg. the bytecode circuit uses keccak table as well.

zkevm-circuits/src/evm_circuit/execution/sha3.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        let values = block
+            .copy_events
+            .get(&key)
+            .unwrap()
+            .steps
+            .iter()
+            .filter(|s| s.rw.is_write())
+            .map(|s| s.value)
+            .collect::<Vec<u8>>();
```
z2trillion, line 139:
```suggestion
use eth_types::ToScalar // move this to the top of the file.
self.copy_rwc_inc.assign(region, offset, size.to_scalar())?;

 let values: Vec<u8> = (3..3 + (size.low_u64() as usize))
      .map(|i| block.rws[step.rw_indices[i]].memory_value())
      .collect();
```

z2trillion: >(1) i mentioned "let key = (tx.id, call.id, step.program_counter as usize);" may be not uniq enough.
We can get rid of the key altogether here (and probably in all the other copy table lookups as well), but just directly calculating `copy_rwc_inc` and `values`.

ChihChengLiang (approved):
LGTM. The implementation matches the spec. Thanks for the great work!

PR #638, opened 2022-07-27
Ci self hosted

ntampakas:
Modified "CI checks" workflow to run specific heavy job(s), on a new self-hosted runner.
Forks, will still be executed on GH runners.
Concurrency was introduced to avoid parallel runs against the self-hosted runner. 

Note:
All self hosted runners have by default 3 tags that cannot be removed.
Since we introduced a new runner in this repo, we changed the tags in two workflows that already use a self-hosted runner to avoid conflicts.

### DIFF
--- a/.github/ciChecksScripts/wakeUpRunner.sh
+++ b/.github/ciChecksScripts/wakeUpRunner.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+profile="cirunner"
+runner_vpc_id="vpc-08ac0db4d8a7cce9e"
+
+# Get runner status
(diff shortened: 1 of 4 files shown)


ntampakas:
Even though concurrency works fine for one pending workflow, when it comes to more than that, the first workflow in the queue is cancelled.
Let me commit a change that fixes it. I` ll also try to isolate the concurrency to the self-hosted runner job(s).

ed255 (approved):
LGTM!  This is really nice! We went from 50 minutes to 10 minutes for the unit tests workflow! :tada:

Issue #636, opened 2022-07-26
Failing `CALL` when sending value


adria0:
`test_with_value` fails 

```
---- testme::test::test_with_value stdout ----
thread 'testme::test::test_with_value' panicked at 'should work: [ConstraintNotSatisfied { constraint: Constraint { gate: Gate { index: 44, name: "CALL" }, index: 114, name: "numerator - remainder == quotient ⋅ denominator" }, location: InRegion { region: Region { index: 6, name: "Execution step" }, offset: 73 }, cell_values: [(VirtualCell { name: "", column: Column { column_type: Advice, index: 23 }, rotation: 0 }, "1")] }, ConstraintNotSatisfied { constraint: Constraint { gate: Gate { index: 44, name: "CALL" }, index: 116, name: "lhs - rhs == diff - (lt ⋅ range)" }, location: InRegion { region: Region { index: 6, name: "Execution step" }, offset: 73 }, cell_values: [(VirtualCell { name: "", column: Column { column_type: Advice, index: 23 }, rotation: 0 }, "1")] }]', evm-testvectors/src/testme.rs:65:56
```
when `test_without_value` works for the following code:

```
    use eth_types::geth_types::GethData;
    use eth_types::{address, bytecode, Word};
    use mock::TestContext;
    use zkevm_circuits::evm_circuit::witness::block_convert;
    use zkevm_circuits::evm_circuit::test::run_test_circuit_incomplete_fixed_table;

    #[test]
    fn test_with_value() {
        test(Word::one());
    }

    #[test]
    fn test_without_value() {
        test(Word::zero());
    }

    fn test(value: Word) {
        let addr0 = address!("0x00000000000000000000000000000000000cafe0");
        let addr1 = address!("0x00000000000000000000000000000000000cafe1");
        let addr2 = address!("0x00000000000000000000000000000000000cafe2");

        let code1 = bytecode! {
            PUSH32(0)
            PUSH32(0)
            PUSH32(0)
            PUSH32(0)
            PUSH32(value)
            PUSH20(0xcafe2)
            PUSH32(100000)
            CALL
            STOP
        };
        let code2 = bytecode! {
            STOP
        };

        let block : GethData = TestContext::<3, 1>::new(
            None,
            |accs| {
                accs[0]
                    .address(addr0)
                    .balance(Word::from(1u64 << 30));
                accs[1].address(addr1).code(code1);
                accs[2].address(addr2).code(code2);
            },
[...]

PR #639, opened 2022-07-27
Fix `CALL` to calculate `is_empty_account` correctly

han0110:
Resovles #636.

It turns out to be a hidden implementation bug which can only be triggered by specific setting (account with code but with zero nonce), which should never happen after [EIP-161](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-161.md)).

### DIFF
--- a/bus-mapping/src/evm/opcodes/call.rs
+++ b/bus-mapping/src/evm/opcodes/call.rs
@@ -101,7 +101,7 @@ impl Opcode for Call {
         )?;
 
         let (_, callee_account) = state.sdb.get_account(&call.address);
-        let is_account_empty = callee_account.is_empty();
+        let is_empty_account = callee_account.is_empty();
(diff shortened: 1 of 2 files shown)


lispc:
oh...if remember correctly this bug can occur in current mainnet settings... Recently i mock proved some real mainnet blocks and fixed some bugs.. (still in my personal branch). It is a bug found by testing mainnet txs.. 

https://github.com/scroll-tech/zkevm-circuits/blob/7da623332b1b8391331f50194d35aaf0b74b9896/zkevm-circuits/src/evm_circuit/execution/call.rs#L470

han0110:
Yes tho a fresh new chain start with EIP-161 enabled won't have such issue, but it could happen on current mainnet, so it's clear now we should check in such conservative way to avoid such edge cases.

lispc (approved):

PR #640, opened 2022-07-27
fix memory expand gas

lispc:


han0110 (approved):
Thanks for testing and fixing this!

PR #643, opened 2022-07-28
Ci self hosted runner (fix)

ntampakas:
This is a fix for the script that wakes up the runner.
Instance state should always be updated and that` s why it was inserted into the loop.


ed255 (approved):
LGTM!

PR #648, opened 2022-08-01
reconstruct access list in bus-mapping

lightsing:
## For reviewers

Apply "Hide writespace" can get a better diff view.

## Reason

Currently statedb calculated incorrectly due to incomplete implementation of [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929). OpCode like `EXTCODESIZE` can easily break the system.

We need correct access list to calculate gas (like `CALL`, `CREATE`, `SSTORE`, etc.)

### Notice

Some of the Opcode need further implementation of busmapping, `TODO` is added to comments.


lispc (approved):

roynalnaruto (approved):
Looks good to me!

PR #655, opened 2022-08-04
Fix RLC values in Copy/Keccak table lookup (SHA3)

roynalnaruto:
Fixes the `integration-tests` warning after merging #635 for `SHA3` using a dummy gadget.

Still investigating failure because of Keccak lookup.

### DIFF
--- a/zkevm-circuits/Cargo.toml
+++ b/zkevm-circuits/Cargo.toml
@@ -16,7 +16,7 @@ sha3 = "0.10"
 digest = "0.7.6"
 array-init = "2.0.0"
 paste = "1.0"
-bus-mapping = { path = "../bus-mapping" }
+bus-mapping = { path = "../bus-mapping", features = ["test"] }
 eth-types = { path = "../eth-types" }
 gadgets = { path = "../gadgets" }
 ethers-core = "0.6"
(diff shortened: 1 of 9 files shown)


icemelon (approved):
LGTM

lispc:
this pr looks good to me. But what about also fixing integration test in this PR?

roynalnaruto:
> But what about also fixing integration test in this PR?

Adding a commit now to fix that!

roynalnaruto:
Because of the Keccak table assignment update, the bytecode tests are failing. Will fix the bytecode config assignment as well.

zkevm-circuits/Cargo.toml
```diff
@@ -16,7 +16,7 @@ sha3 = "0.10"
 digest = "0.7.6"
 array-init = "2.0.0"
 paste = "1.0"
-bus-mapping = { path = "../bus-mapping" }
+bus-mapping = { path = "../bus-mapping", features = ["test"] }
```
lispc, line 19: It is possible to make zkevm circuits only need busmapping test feature when zkevm circuit is also in test mode？

roynalnaruto: Yes I will make the change

lispc (approved):

PR #656, opened 2022-08-04
Remove keys from CopyEvent

z2trillion:
They were being used to look up `copy_rwc_inc`, which can be directly computed.

### DIFF
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -373,7 +373,7 @@ impl<F: Field> CopyCircuit<F> {
             || "assign copy table",
             |mut region| {
                 let mut offset = 0;
-                for copy_event in &block.copy_events {
+                for copy_event in block.copy_events.iter() {
                     let rlc_acc = if copy_event.dst_type == CopyDataType::RlcAcc {
                         let values = copy_event
                             .steps
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -334,7 +334,7 @@ impl<F: Field> CopyCircuit<F> {
             || "assign copy table",
             |mut region| {
                 let mut offset = 0;
-                for copy_event in block.copy_events.values() {
+                for copy_event in &block.copy_events {
                     for (step_idx, copy_step) in copy_event.steps.iter().enumerate() {
                         self.assign_step(
                             &mut region,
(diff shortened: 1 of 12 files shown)


zkevm-circuits/src/copy_circuit.rs
```diff
@@ -373,7 +373,7 @@ impl<F: Field> CopyCircuit<F> {
             || "assign copy table",
             |mut region| {
                 let mut offset = 0;
-                for copy_event in block.copy_events.values() {
+                for copy_event in &block.copy_events {
```
roynalnaruto, line 376:
```suggestion
                for copy_event in block.copy_events.iter() {
```

This adds consistency to the rest of the changes. Nitpick

zkevm-circuits/src/evm_circuit/execution/calldatacopy.rs
```diff
@@ -191,8 +190,11 @@ impl<F: Field> ExecutionGadget<F> for CallDataCopyGadget<F> {
             ),
         )?;
         let src_id = if call.is_root { tx.id } else { call.caller_id };
-        self.src_id
-            .assign(region, offset, Some(F::from(src_id as u64)))?;
+        self.src_id.assign(
+            region,
+            offset,
+            Some(F::from(u64::try_from(src_id).unwrap())),
```
roynalnaruto, line 196: Why not simply `F::from(src_id as u64)` ?

roynalnaruto (approved):
Overall a great PR, removes redundant code, thank you :)

zkevm-circuits/src/evm_circuit/execution/calldatacopy.rs
z2trillion: I'm a big fan of not using `as` to convert values because it doesn't trigger some clippy warning. E.g.
```
let x: u64 = 9;
let _y: u64 = x as u64; // clippy is ok with this.
let _z: u64 = x.into(); // error: useless conversion to the same type: `u64`
let _a: u64 = x.try_into().unwrap(); // error: useless conversion to the same type: `u64`
```

z2trillion: When I have the time, I want to implement `ToScalar` for `usize` and `u64` so we can all type less.

zkevm-circuits/src/evm_circuit/execution/calldatacopy.rs
```diff
@@ -201,20 +203,28 @@ impl<F: Field> ExecutionGadget<F> for CallDataCopyGadget<F> {
             (call.call_data_length, call.call_data_offset)
         };
         self.call_data_length
-            .assign(region, offset, Some(F::from(call_data_length as u64)))?;
```
z2trillion, line 204: This is an example where a cast from u64 to u64 didn't raise any clippy errors.

lispc (approved):

Issue #650, opened 2022-08-02
backlog: unify circuits naming


lispc:
Sometimes we use "XXCircuitConfig" as the struct containing columns and "XXCircuit" as the struct to containing witness(block), while sometime we use "XXCircuit" as the struct containing columns.

There are more inconsistency. Eg: EvmCircuit is the "proper" circuit without shared tables, and "TestCircuit" in evm_circuit.rs is the  "full(runnable)" circuit.   While for StateCircuit there is no "proper" circuit without shared tables. "StateCircuit" is the "full" circuit. 

I suggest XXCircuitConfig will be columns while XXCircuit be witness(block).   And we should  unify names like "MyCircuit"/"TestCircuit" to "BytecodeTestCircuit"/"EvmTestCircuit" later.

aguzmant103:
Besides the above, also unify the name of the functions for how a circuit is configured and assigned. This is helpful to simplify the testnet.

ed255:
Here's a summary of the current status:

| circuit  | columns struct     | witness struct      | config constructor            | assign fn                                                      | instance build        |
| ---      | ---                | ---                 | ---                           | ---                                                            | ---                   |
| EVM      | EvmCircuit         | (only test)         | EvmCircuit::configure         | EvmCircuit.assign_block(layouter, block)                       | -                     |
| State    | StateCircuitConfig | StateCircuit        | StateCircuitConfig::configure | StateCircuitConfig.assign(layouter, rws, ...)                  | -                     |
| Tx       | TxCircuitConfig    | TxCircuit           | TxCircuitConfig::new          | TxCircuit.assign(tx_circuit_cfg, layouter, ...)                | -                     |
| Bytecode | Config             | (only test)         | BytecodeConfig::configure     | BytecodeConfig.assign(layouter, ..., bytecodes)                | -                     |
| Copy     | CopyCircuit        | (only test)         | CopyCircuit::configure        | CopyCircuit.assign_block(layouter, block)                      | -                     |
| Exp      | ExpCircuitConfig   | ExpCircuit          | ExpCircuitConfig::configure   | ExpCircuitConfig.assign_block(layouter, block)                 | -                     |
| Keccak   | KeccakPackedConfig | KeccakPackedCircuit | KeccakConfig::configure       | KeccakConfig.assign_from_witness(layouter, keccak_inputs, ...) | -                     |
[...]

PR #670, opened 2022-08-08
add state circuit to super circuit

lispc:
most of the changes are for extracting the "rw table" out of state circuit so it can be shared between evm circuit and state circuit. Most of circuit config and assignment codes are moved from `StateCircuit` to `StateCircuitConfig` similar to EvmCircuit, so `StateCircuit` now becomes a "tester" circuit similar to TestCircuit in evm_circuit.rs, or "MyCircuit" in bytecode circuit / copy circuit / tx circuit.

My concern is that moving codes from StateCircuit to StateCircuitConfig results large diffs and makes it a bit difficult to review... Any suggestions..? 

I manually tested this PR using super circuit with cmd `cargo test --release -p zkevm-circuits skip_test_super_circuit -- --ignored` and it passed.

TODO:

the newly added "aux2" and "value_prev" cols in rw table are not constrained yet. After https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/621 is merged, we can revisit how to constrain the relationship between initial_value / committed_value / prev_value.


zkevm-circuits/src/evm_circuit/witness.rs
```diff
@@ -377,7 +379,62 @@ impl std::ops::Index<(RwTableTag, usize)> for RwMap {
         &self.0.get(&tag).unwrap()[idx]
     }
 }
-
+impl RwMap {
+    /// Check rw_counter is continous and starting from 1
+    pub fn check_rw_counter_sanity(&self) {
```
DreamWuGit, line 384: i like this check style !!

zkevm-circuits/src/state_circuit.rs
```diff
@@ -69,8 +213,8 @@ pub struct SortKeysConfig {
 type Lookup<F> = (&'static str, Expression<F>, Expression<F>);
 
 /// State Circuit for proving RwTable is valid
-#[derive(Default)]
-pub struct StateCircuit<F: Field> {
+#[derive(Default, Clone)]
+pub struct StateCircuit<F> {
     pub(crate) randomness: F,
     pub(crate) rows: Vec<Rw>,
     pub(crate) n_rows: usize,
```
DreamWuGit, line 220: one question:  Does `n_rows` need to be equal to evm circui rows number in real proof ?

zkevm-circuits/src/super_circuit.rs
```diff
@@ -176,9 +184,13 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize> Circuit<F>
...
             .load_fixed_table(&mut layouter, self.fixed_table_tags.clone())?;
         config.evm_circuit.load_byte_table(&mut layouter)?;
-        config
-            .rw_table
-            .load(&mut layouter, &self.block.rws, self.block.randomness)?;
+        config.rw_table.load(
+            &mut layouter,
+            &self.block.rws.table_assignments(),
+            self.block.randomness,
+            self.block.state_circuit_pad_to,
```
DreamWuGit, line 191: where is `state_circuit_pad_to` assigned ?

zkevm-circuits/src/super_circuit.rs
```diff
@@ -74,6 +75,7 @@ pub struct SuperCircuitConfig<F: Field, const MAX_TXS: usize, const MAX_CALLDATA
     keccak_table: KeccakTable,
     copy_table: CopyTable,
     evm_circuit: EvmCircuit<F>,
+    state_circuit: StateCircuitConfig<F>,
```
DreamWuGit, line 78: not sure if need add `state_circuit` into L86 like `tx_circuit` I see it will be used for unit tests

DreamWuGit (approved):
overall LGTM!! leave minor comments

z2trillion:
Are we still planning to move to RLC based lookups in the future? 

I think the `value_prev` column in the RwTable can just be constrained to be equal to the `value_prev` in the state circuit constraint builder.

z2trillion (changes requested):

zkevm-circuits/src/evm_circuit/witness.rs
roynalnaruto, line 384: In case the RWs do not pass the sanity check, do we want to panic (the `debug_assert_eq` will panic)? Or handle the error before loading the RW table?

zkevm-circuits/src/state_circuit/lexicographic_ordering.rs
```diff
@@ -224,6 +219,7 @@ impl Config {
         BinaryNumberChip::construct(self.first_different_limb).assign(region, offset, &index)?;
 
         let limb_difference = F::from(*cur_limb as u64) - F::from(*prev_limb as u64);
+        //assert!(limb_difference < F::from(1 << 16));
```
roynalnaruto, line 222: We can remove this line?

zkevm-circuits/src/super_circuit.rs
```diff
@@ -188,6 +200,12 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize> Circuit<F>
         config
             .evm_circuit
             .assign_block(&mut layouter, &self.block)?;
+        config.state_circuit.assign(
+            &mut layouter,
+            &self.block.rws.table_assignments(),
+            self.block.state_circuit_pad_to,
+            self.block.randomness,
```
roynalnaruto, line 207: We can make the order of these arguments the same as the ones for RW table.. not a big issue
```suggestion
            self.block.randomness,
            self.block.state_circuit_pad_to,
```

roynalnaruto (approved):
Just a couple of questions, looks good overall. Also the `copy_circuit.rs` needs resolving conflict from the last merge :)

zkevm-circuits/src/state_circuit.rs
lispc: no. n_rows is the rows of rw table. rw table in evm circuit and state circuit should be same.   Not related to the total rows (max of fixed_table, rw_table, tx_table, other parts..) of circuit.

zkevm-circuits/src/super_circuit.rs
lispc: it is left to do in the final circuit (aggregation circuit  / super circuit / etc) in the future. Currently we can either set it to (1<<k - 64), or in tests set it to 0 to indicate no padding needed.

zkevm-circuits/src/super_circuit.rs
lispc: nonono `TxCircuit` is witness, while `StateCircuitConfig` is columns....  See for more details: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/650

zkevm-circuits/src/evm_circuit/witness.rs
lispc: there are two scenarios that cannot satisfy `check_rw_counter_sanity`. (1)in evm circuit witness is wrong, it is bug so we debug_assert (2) we manually construct test cases for state circuit where rw_counter may not be continuous.

So i only `check_rw_counter_sanity` in evm circuit assignment, and let it fail if error happens

zkevm-circuits/src/super_circuit.rs
lispc: yea, changed the order in state circuit. seems more intuitive

z2trillion (approved):
looks good

lispc:
> Are we still planning to move to RLC based lookups in the future?
> 

 i am not sure. Even in scroll-dev branch i no longer use RLC to reduce the long term fork maintenance burden at the cost of 1-2% more rows in aggregation circuit.
 
> I think the `value_prev` column in the RwTable can just be constrained to be equal to the `value_prev` in the state circuit constraint builder.


there are still something missing to constrain the relationship between value.prev() <=> value_prev.cur().

At least Rw::Account prev value cannot satisfy the equality.   In another branch i constrain value_prev like this:  https://github.com/scroll-tech/zkevm-circuits/blob/bdf8fd3361d49fc6d8110489d744ce202cf022ee/zkevm-circuits/src/state_circuit/constraint_builder.rs#L154

Issue #673, opened 2022-08-10
Fix oog error in call handling 


DreamWuGit:
@lispc  found if oog(out of gas) error occurs in `Call`,  circuit will fail unexpectedly.   after investigation, for  oog and stack related error cases(stack overflow& underflow) don't need push call in the buss mapping.

DreamWuGit:
close this as #674 merged!

PR #674, opened 2022-08-10
fix oog in call handling #673

DreamWuGit:
issue #673 fixes and add dedicated unit test for it.


bus-mapping/src/evm/opcodes.rs
```diff
@@ -283,7 +283,7 @@ pub fn gen_associated_ops(
         );
 
         exec_step.error = Some(exec_error);
-        if geth_step.op.is_call_or_create() {
+        if geth_step.op.is_call_or_create() && !exec_step.oog_or_stack_error() {
```
lispc, line 286: it would be better to add several lines of comments here to explain why oog and stack don't need the push_call

lispc (approved):
lgtm.    only a small suggestion

bus-mapping/src/circuit_input_builder/execution.rs
```diff
@@ -71,6 +71,18 @@ impl ExecStep {
...
+    /// Returns `true` if `error` is oog and stack related..
+    pub fn oog_or_stack_error(&self) -> bool {
+        if self.error.is_none() {
+            return false;
+        }
+
+        matches!(
+            self.error.clone().unwrap(),
+            ExecError::OutOfGas(_) | ExecError::StackOverflow | ExecError::StackUnderflow
+        )
```
ChihChengLiang, line 84: Looks like it can be simplified
```suggestion
        matches!(self.error, Some(ExecError::OutOfGas(_) | ExecError::StackOverflow | ExecError::StackUnderflow))
```

ChihChengLiang (commented):
Added a nitpick

bus-mapping/src/circuit_input_builder/execution.rs
DreamWuGit: updated in [aa107c0](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/674/commits/aa107c0271feb0aa70118b0d596dcb34304d9afd) thanks !!

bus-mapping/src/evm/opcodes.rs
DreamWuGit: added in [aa107c0](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/674/commits/aa107c0271feb0aa70118b0d596dcb34304d9afd) thanks!!

ChihChengLiang (approved):
LGTM

PR #678, opened 2022-08-12
test: simplify fixed table tags

lispc:
Detect needed fixed table tags automatically from witness block. Remove some useless codes.


han0110 (approved):
LGTM! Nice refactor.

lispc:
I think this pr is trivial so i will just merge it

PR #686, opened 2022-08-16
Constrain rw_counter increase for copy table lookups

z2trillion:
The value of the `copy_rwc_inc` was previously not contrained, so it was possible copy additional bytes into or out of the memory.


icemelon (commented):
I think this could be a redundant check because the copy circuit will constrain the `rwc_inc` via the copy length. The `length` in the copy circuit will decrease by 1 each step, and `rwc_inc_left` and `rw_counter` in the copy circuit will dev/inc by 1 depending on if it's a lookup to the rw table and not padded.

z2trillion:
I made a repro where all the tests pass, even though the callcopygadget only copies padding bytes: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/687. I think all of the tests are passing because we're missing test coverage, because there's a constraint that seems to disallow this: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/687/files#diff-faa8cd2ba1550c16574b2560fa1b4fec0055de23164ada2bcf5dda20845761d7R255.

@roynalnaruto, is my interpretation correct?

If so, we can remove `rwc_inc_left` and `is_pad` fields from `CopyStep`, since the copy circuit can already compute the correct assignment from the other information in the `CopyEvent` and `CopyStep`.

roynalnaruto:
Closing as #691 clarifies the confusion.

PR #687, opened 2022-08-16
Repro - copy_rwc_inc

z2trillion:
It's valid to only write padding bytes for Calldatacopy.


roynalnaruto:
Hi @z2trillion which tests did you try to run with this change? Could you try running:
```
$ cargo test --release copy_circuit
```

PR #718, opened 2022-08-29
[WIP] Reduce min EVM Step height to 1

### LINKED ISSUE #711: Make the minimum state height = 1
Currently the minimum height is = 2. To facilitate the padding with the end block will require the minimum height = 1. This can be achieved by reducing the cells used for the selector of the state.

ed255:
WIP for now to get CI results.

Resolve #711


zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -322,10 +323,96 @@ impl ExecutionState {
...
+        let bool_checks = iter::once(&self.target_even)
+            .chain(&self.target_pairs)
+            .map(|cell| {
+                (
+                    "Representation for target_pairs and target_even should be bool",
+                    cell.expr() * (1u64.expr() - cell.expr()),
+                )
+            });
+        iter::once(sum_to_one).chain(bool_checks).collect()
+    }
```
han0110, line 366: It seems we need to disallow the invalid case just like `BinaryNumberChip`, to avoid invalid selector be enabled all the way down to escape constraints?

If so in this case we only need to add an extra constraint `target_even * target_pairs.last() == 0` when `count` is odd.

han0110 (commented):
LGTM! Also the documents look great and they are really helpful. Just left one question here.

zkevm-circuits/src/evm_circuit/step.rs
ed255: Well spotted!  Thanks!  I've fixed this in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/718/commits/e2bcc55aea01340382f818e3c6d7c88a73aac98e

I've also replaced the usage of `even` flag by `odd` flag so that it matches with the less significant bit of the value.  I think this way it's easier to reason about.

han0110: Makes sense!

Brechtpd:
> Reduce the cells used for the execution state selector in order to achieve a minimum step height of 1. This can be achieved by using slightly more complex selector expressions (that have degree 2) that only require about half of the cells.

The circuit width isn't just increased because of aggregation circuit considerations? If we can go wider then extended k = 1 is likely achievable, which especially for the EVM circuit could be worth it (custom gates in the EVM circuit are a big chunk of the prover time).

lispc:
> > Reduce the cells used for the execution state selector in order to achieve a minimum step height of 1. This can be achieved by using slightly more complex selector expressions (that have degree 2) that only require about half of the cells.
> 
> The circuit width isn't just increased because of aggregation circuit considerations? If we can go wider then extended k = 1 is likely achievable, which especially for the EVM circuit could be worth it (custom gates in the EVM circuit are a big chunk of the prover time).

In our experience the wider the evm circuit the more point operation in aggregation circuit the more aggregation proving time. In fact we even  make extend k to 3 so evm circuit becomes thiner to reduce aggregation circuit proving time. Han also did some experiments on this? @h

lispc (approved):

han0110:
Rethink again I think Brecht's point makes sense, since EVM circuit has much more custom gates than Aggregation circuit, so decrease EVM circuit's extended domain (by making it wider) and increase Aggregation circuit size seems reasonable (if the capacity allows). Although not sure the performance without benchmark, I believe this approach will have a faster proving time.

In the past we thought we can only do ~800 MSM, but since right now we can do ~2250 (or ~4500 if changing halo2's subset argument), we might want to adjust our strategy if larger Aggregation circuit has less cost than higher degree / larger circuit of application circuits.

@ed255 Do you think we need a benchmark (EVM circuit with different extended k + Aggregation circuit with different size) to see if this is the case?

lispc:
share some data from my side:

1. typical evm circuit k: Usually a 30M gas mainnet block needs ~7M rows, or k = 23. (While for small blocks, the minimalist k can be reduce to about 17 or 18.) 
2.typical aggregation circuit size: at least it needs k = 25. Each point operation(ecmul) needs 32k rows.

han0110:
@lispc Thanks for real world data, this is really helpful! So now I'm wondering 2 things:

1. If we make EVM circuit wide enough such that `EndBlock` is single row (so add extra 29 = 106 - 77 columns), and keep extended k as 2. Will this extra 29 * 32k rows increase Aggregation circuit size in your case?
2. If 1. is true, how much extra proving resource (time and memory) costs? Is that more than making EVM circuit extended k as 3?

If 2. is also true, then we should continue the approach in this PR.

lispc:
> Will this extra 29 * 32k rows increase Aggregation circuit size in your case?

Not. In fact the total "columns" of evm circuit is not a bit above 100, but 300-400 i remember. Each lookup needs 3(or 4?) implicit columns. Add another 30 columns, is not to increase 1/4, but increase 1/10.

For k = 25 for aggregation circuit, we needs 250-300G memory, and about 1 h for a 96 core machine.  Memory and time will double for k = 26. We have GPU halo2, the proving time will be reduced to about 5-10 minutes for k = 25/26.

> Is that more than making EVM circuit extended k as 3?

Since we use k = 18 for evm circuit currently, aggregation with k = 25 needs **much more** computing resource than evm circuit. (But later if we choose bigger k for evm circuit, it is another story).

han0110:
> Since we use k = 18 for evm circuit currently, aggregation with k = 25 needs much more computing resource than evm circuit.

I see. So it also depends on the size of EVM circuit.

I also think this approach is good since in Super circuit we already have extended k = 3, so it actually doesn't cost anything more for us. Perhaps we can go for this approach first, and update the strategy in the future when EVM circuit gets larger (for both Scroll and PSE's case), or when we no longer need Super circuit and each circuit can have its own extended k (for PSE's case)?

PR #719, opened 2022-08-29
RETURN gadget for non-create calls

### LINKED ISSUE #573: `RETURN` opcode
Implement the RETURN execution state for the RETURN opcode.

This opcode uses dynamic length memory copy so it may depend on the Copy Circuit.

Depends on spec: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/229

z2trillion:
This gadget implements the specs in https://github.com/scroll-tech/zkevm-specs/blob/master/specs/opcode/F3RETURN.md, except for "A. Returns the specified memory chunk as deployment code." I plan to add the copy lookup for code deployments in a follow up PR.

It also does handles the REVERT opcode, which is generally similar to RETURN.

TODO: add REVERT spec.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203088778995027


bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -824,6 +824,84 @@ impl<'a> CircuitInputStateRef<'a> {
...
+            ),
+            (
+                CallContextField::ReversibleWriteCounter,
+                self.caller_ctx()?.reversible_write_counter.into(),
+            ),
+        ] {
+            self.call_context_read(exec_step, caller.call_id, field, value);
+        }
+
+        let [last_callee_return_data_offset, last_callee_return_data_length] = match geth_step.op {
```
ed255, line 872: Why is the length `2` for STOP and RETURN/REVERT(offset, 0) ?
Shouldn't it be `0`?

bus-mapping/src/evm/opcodes/return.rs
```diff
@@ -1,175 +1,199 @@
...
+
+        let call = state.call()?.clone();
+        let is_root = call.is_root;
+        for (field, value) in [
+            (CallContextField::IsRoot, is_root.to_word()),
+            (CallContextField::IsCreate, call.is_create().to_word()),
+            (CallContextField::IsSuccess, call.is_success.to_word()),
+            (CallContextField::CallerId, call.caller_id.into()),
+            (
+                CallContextField::ReturnDataOffset,
```
ed255, line 45: I think this implementation is not quite following the spec.  For example, the lookup for `ReturnDataOffset` and `ReturnDataLength` is done conditionally in the case `not instruction.curr.is_root and not instruction.curr.is_create`.

Also the cases implemented here don't follow exactly the ones in the spec, so the logic is a bit different.

Could you annotate the code to make it clear where the different paths `A, B, C, D` are implemented?  The same applies for the circuit code.

And if you made the decision to deviate from the spec, could you explain why?

ed255 (commented):
I've taken an initial look and saw some differences from the spec, and I also miss the annotations `A, B, C, D` which help following the implementation with the spec.  Maybe the differences come from the fact that you've also added the `REVERT` case?  If that's the case, I would like to review the updated spec (with REVERT case) before continuing the review of this PR.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
z2trillion: It is 0? `[Word::zero(); 2]` means `[Word::zero(), Word::zero()]` and not `[Word::zero(), Word::from(2)]`.

ed255: Oh, sorry for the confusion! :flushed:

bus-mapping/src/evm/opcodes/return.rs
z2trillion: I have added references to the specs in the comments here and updated the specs in https://github.com/privacy-scaling-explorations/zkevm-specs/pull/283 so it matches the implemenation here.

zkevm-circuits/src/evm_circuit/execution/return.rs
```diff
@@ -24,25 +49,342 @@ impl<F: Field> ExecutionGadget<F> for ReturnGadget<F> {
...
+        );
+        cb.condition(
+            not::expr(is_create.expr())
+                * not::expr(is_root.expr())
+                * not::expr(copy_rw_increase_is_zero.expr()),
+            |cb| {
+                cb.copy_table_lookup(
+                    cb.curr.state.call_id.expr(),
+                    CopyDataType::Memory.expr(),
+                    caller_id.expr(),
```
ed255, line 147: Couldn't you skip a lookup and use `cb.next.state.call_id.expr()` here, like in the spec?
https://github.com/privacy-scaling-explorations/zkevm-specs/blob/677305ecb845cfec8b91cd52195af576feb29564/src/zkevm_specs/evm/execution/return_.py#L68

z2trillion: I wasn't aware that was available on the constraint builder :)

zkevm-circuits/src/evm_circuit/execution/return.rs
```diff
@@ -24,25 +46,313 @@ impl<F: Field> ExecutionGadget<F> for ReturnGadget<F> {
...
+        });
+
+        // Case C in the specs.
+        // TODO: have copy_table_lookup update rw_counter expression so that this can go
+        // at the end of the constraints.
+        let restore_context = cb.condition(not::expr(is_root.expr()), |cb| {
+            RestoreContextGadget::construct(
+                cb,
+                is_success.expr(),
+                2.expr() + copy_rw_increase.expr(),
```
ed255, line 107: I wonder where this `2` comes from.  I'm counting 3 + (2) fixed lookups:
- 2 Stack lookup
- CallContextFieldTag::Success lookup
- CallContextFieldTag::{ReturnDataOffset, ReturnDataLength} if not(is_create)

Could you help me clarify this?  Considering that the tests are passing, maybe I missed something :thinking:

ed255 (commented):
In general looks good!  I noticed 2 things that may be missing (the memory expansion gas cost, and state transition when root), and I also added a few questions.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -99,12 +99,14 @@ pub(crate) struct RestoreContextGadget<F> {
...
         cb: &mut ConstraintBuilder<F>,
-        rw_counter_delta: Expression<F>,
+        is_success: Expression<F>,                      // just remove this?
+        copy_lookup_rw_counter_increase: Expression<F>, // try to set it up so this is always 0?
         return_data_offset: Expression<F>,
         return_data_length: Expression<F>,
     ) -> Self {
         // Read caller's context for restore
         let caller_id = cb.call_context(None, CallContextFieldTag::CallerId);
+        // TODO: lookup is_success here instead of taking in as argument.
```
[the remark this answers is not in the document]
z2trillion: You're right. I removed this TODO.

zkevm-circuits/src/evm_circuit/execution/return.rs
```diff
@@ -24,25 +46,313 @@ impl<F: Field> ExecutionGadget<F> for ReturnGadget<F> {
...
-        Self { opcode }
+        // Case B in the specs.
+        cb.condition(is_root.expr(), |cb| {
+            cb.require_next_state(ExecutionState::EndTx);
+            cb.call_context_lookup(
+                false.expr(),
+                None,
+                CallContextFieldTag::IsPersistent,
+                is_success.expr(),
+            );
```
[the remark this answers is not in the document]
z2trillion: I added the `StepStateTransition` here. 

I think that we should never be using `StepStateTransition::any()` /  `Transition::Any`?

zkevm-circuits/src/evm_circuit/execution/return.rs
z2trillion: I renamed this argument to be `subsequent_rw_lookups` because it's the expression for the increase in the rw counter after this gadget is constructed. The two come from the lookups for `ReturnDataOffset` and `ReturnDataLength` in case D (if it's not a create, which was not accounted for).

zkevm-circuits/src/evm_circuit/execution/return.rs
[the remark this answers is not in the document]
ed255: > I added the `StepStateTransition` here.
> 
> I think that we should never be using `StepStateTransition::any()` / `Transition::Any`?

When using `any` we just don't enforce any constraint for the next step on that field, meaning that the prover is free to set it to any value.  When `any` is used, less constraints are defined, which can be good? (To be sincere, I'm not really sure how adding or removing these constraints affects the performance, maybe it's negligible?)

The argument in favor of using `any` is that in some execution states, we constraint which execution state the next one will be; and that next step may already constrain some of its own fields.  After a root return, we can only encounter `EndTx`. I believe `EndTx` only uses `log_id`, `gas_left` and `rw_counter`.  After `EndTx` we can encounter `BeginTx` or `EndBlock`.
`EndBlock` only uses `rw_counter` and `call_id`.
`BeginTx` is going to constraint the following step fields to fixed values (not deltas): https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2235254416aefd1bc9b5592e4e06cd58f8516e2a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs#L200

So in conclusion, I think here we can leave some fields to `any`, as they will either be unused or constrained to a fixed value later.  But constraining them all is also correct.  And in general, it seems a good idea to avoid leaving unconstrained witnesses as much as possible, because it gives code with less chances of having bugs.  For example in the snippet I copied I realize I missed constraining `log_id` to be `Same`, which would be a bug.  But I'm unable to reason about possible performance costs.

zkevm-circuits/src/evm_circuit/execution/return.rs
ed255: Ah got it, thanks!

zkevm-circuits/src/evm_circuit/execution/return.rs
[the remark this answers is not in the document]
z2trillion: In general, I agree that we shouldn't add redundant constraints. However, I think it's better for clarity to constrain every field of the `StepStateTransition`. This way, we don't need to reason about what fields of it EndTx, EndBlock, or BeingTx use.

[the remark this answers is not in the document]
ed255: Sounds good to me!

ed255 (approved):
LGTM!  Very nice work! :smiley:

ed255:
@roynalnaruto could you help with a second review to this PR?  Thanks!

z2trillion:
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/817 broke the test here, but I don't know why.

ed255:
> #817 broke the test here, but I don't know why.

Ah, the test that fails is `evm_circuit::execution::r#return::test::test_return_nonroot`.  Maybe there's a bug in the memory reconstruction logic somewhere?

z2trillion:
> I think I found the bug, we were reading from geth_step memory in the circuit input builder, but that's not populated when `enable_memory = false`, so we should always avoid it and rebuild the memory / memory.length. This should fix it:
> 
> ```diff
> diff --git a/bus-mapping/src/circuit_input_builder/input_state_ref.rs b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
> index f243f0cb..3fe96207 100644
> --- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
> +++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
> @@ -889,10 +889,7 @@ impl<'a> CircuitInputStateRef<'a> {
>                  geth_step_next.stack.stack_pointer().0.into(),
>              ),
>              (CallContextField::GasLeft, caller_gas_left.into()),
> -            (
> -                CallContextField::MemorySize,
> -                geth_step_next.memory.word_size().into(),
> -            ),
> +            (CallContextField::MemorySize, next_memory_word_size.into()),
>              (
>                  CallContextField::ReversibleWriteCounter,
>                  self.caller_ctx()?.reversible_write_counter.into(),
> ```
> 
> @z2trillion can you try this change, along with `enable_memory: false` in `LoggerConfig::default()`?

Thanks for finding the spot with the error. The correct new value is `self.caller_ctx()?.memory.word_size().into(),`

roynalnaruto (approved):
Everything looks good to me! I have a comment though, with regards to `is_success` not being constrained as boolean (in the `ReturnGadget`)

PR #720, opened 2022-08-29
Write extcodecopy circuit

DoHoonKim8:
I wrote the circuit for verifying `extcodecopy` opcode and added some tests.

+ I implemented bus-mapping for `extcodecopy` and added some tests too.

Spec link: [spec](https://github.com/privacy-scaling-explorations/zkevm-specs/pull/257)

Give me many comments, Thanks!


swjng (commented):
First of all, move [this line](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/289f92a490342183d982ce1cc4385fdd22cb5d9c/zkevm-circuits/src/evm_circuit/execution.rs#L940) into the 'opcodes' above for readability.
And I think you should implement copy event for bus-mapping in [bus-mapping/src/evm/opcodes/extcodecopy.rs](https://github.com/scroll-tech/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/extcodecopy.rs#L48).

DoHoonKim8:
@ChihChengLiang Would you please review this PR? :)🙏

ChihChengLiang:
Yeah sure, sorry for the delay. Let me also tag @ed255 and @roynalnaruto because they've reviewed the spec.

ChihChengLiang (approved):
Added a question. Other places LGTM. Great pull request! Thanks for the contribution.

zkevm-circuits/src/evm_circuit/execution/extcodecopy.rs
```diff
@@ -0,0 +1,391 @@
...
+            region,
+            offset,
+            call.rw_counter_end_of_reversion,
+            call.is_persistent,
+        )?;
+        let is_warm = match GasCost::from(step.gas_cost) {
+            GasCost::COLD_ACCOUNT_ACCESS => 0,
+            GasCost::WARM_ACCESS => 1,
+            _ => unreachable!(),
+        };
```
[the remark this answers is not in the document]
DoHoonKim8: I think your suggestion is better. I referred to `extcodehash.rs` for here, and maybe I can update the both side :)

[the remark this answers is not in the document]
DoHoonKim8: Fixed in [f3dd307](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720/commits/f3dd3073e721527f5b0f57eaf66d275438af0f58) :)

zkevm-circuits/src/evm_circuit/execution/extcodecopy.rs
```diff
@@ -0,0 +1,388 @@
...
+            [memory_address.address()],
+        );
+        let memory_copier_gas = MemoryCopierGasGadget::construct(
+            cb,
+            memory_address.length(),
+            memory_expansion.gas_cost(),
+        );
+        let gas_cost = memory_copier_gas.gas_cost()
+            + is_warm.expr() * GasCost::WARM_ACCESS.expr()
+            + (1.expr() - is_warm.expr()) * GasCost::COLD_ACCOUNT_ACCESS.expr();
```
Brechtpd, line 92:
```suggestion
        let gas_cost = memory_copier_gas.gas_cost()
            + select::expr(is_warm.expr(), GasCost::WARM_ACCESS.expr(), GasCost::COLD_ACCOUNT_ACCESS.expr());
```

DoHoonKim8: Fixed in 3347b40 :)

DoHoonKim8:
@roynalnaruto @ed255 Would you please review this PR? :) 🙏

bus-mapping/src/evm/opcodes/extcodecopy.rs
```diff
@@ -56,90 +60,190 @@ fn gen_extcodecopy_step(
...
+) -> Result<CopyEvent, Error> {
+    let rw_counter_start = state.block_ctx.rwc;
+    let external_address = geth_step.stack.nth_last(0)?.to_address();
+    let memory_offset = geth_step.stack.nth_last(1)?.as_u64();
+    let data_offset = geth_step.stack.nth_last(2)?.as_u64();
+    let length = geth_step.stack.nth_last(3)?.as_u64();
+
+    let mut exec_step = gen_extcodecopy_step(state, geth_step)?;
+    let code_hash = state.sdb.get_account(&external_address).1.code_hash;
+    let code = state.code(code_hash)?;
```
roynalnaruto, line 155: You can get the `eth_types::Bytecode` here:
```suggestion
    let code: Bytecode = state.code(code_hash)?.into();
```

bus-mapping/src/evm/opcodes/extcodecopy.rs
```diff
@@ -56,90 +60,190 @@ fn gen_extcodecopy_step(
...
+) -> Result<Vec<(u8, bool)>, Error> {
+    let mut copy_steps = Vec::with_capacity(bytes_left as usize);
+    for idx in 0..bytes_left {
+        let addr = src_addr + idx;
+        let value = if addr < src_addr_end {
+            code[addr as usize]
+        } else {
+            0
+        };
+        copy_steps.push((value, false));
```
roynalnaruto, line 136: You're forcing `is_code` to be `false` always, which is not correct.

Go through the following comment, where you can get a `eth_types::Bytecode` from the state DB. This will have both the `value` and `is_code` for each byte in the bytecode.

roynalnaruto (changes requested):
Went through a first pass, found some issues.

Also, could you add a test here for `extcodecopy`:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/copy_circuit.rs#L791

The reason why `is_code: false` was not an issue in your tests is because the copy circuit constraints were never checked (only lookup to copy table is done from EVM circuit).

You'll need to fix that part for the copy circuit to be satisfied.

roynalnaruto (changes requested):
Could you add the test case in `copy_circuit.rs` for `EXTCODECOPY`?

bus-mapping/src/evm/opcodes/extcodecopy.rs
DoHoonKim8: Fixed in [23aaa95](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720/commits/23aaa95f3d673b00942f76535a92177cda08a4ff).

bus-mapping/src/evm/opcodes/extcodecopy.rs
DoHoonKim8: Fixed in [7448dbe](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720/commits/7448dbe514f2f96f0232230c1b1a916594c690b7).

DoHoonKim8:
> Could you add the test case in `copy_circuit.rs` for `EXTCODECOPY`?

I think some tests are failing due to the changes in ea63957dab1c9613f17c961e50d9cc687383f5db. I'll resolve this and let you know again :)

DoHoonKim8:
@roynalnaruto 
Sorry for being quite late for this. I still cannot find out why `EXTCODECOPY` test case is failing in `copy_circuit.rs`. Can you help me for this issue?

ChihChengLiang:
@DoHoonKim8 I checked and reproduced the testing error. It looks like the `external_address` is not written into the CopyTable, so the lookup failed. I traced the `assign_block` function, and the related CopyEvent for  `external_address` was not there. 

I think ` accs[1]` in the test_ctx, which corresponds to the `external_address` was ignored. The block builder didn't generate the corresponding CopyEvent for it. We might need to add the CopyEvent somewhere.

roynalnaruto:
@ChihChengLiang @DoHoonKim8 Since the bytecode lookup is failing, I believe we can start debugging from [here](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/copy_circuit.rs#L672-L674):
- Only `block.bytecodes` is populated into the bytecode table.

The witness block is constructed [here](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/witness/block.rs#L173-L188):
- Which means it is only considering the `code_hash` of each `call` in the block's transactions. It does not take into consideration the bytecode at the `external_address`.

One possible solution is to:
* Add a field `ext_code_hashes: Vec<Hash>` to the [circuit_input_builder::block](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/block.rs#L78-L79) and:
* Populate it during the `bus-mapping` implementation of [`EXTCODECOPY`](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/extcodecopy.rs#L52)
```rust
    state.add_ext_code_hash(ext_code_hash)?;
```
which adds this to the block:
```rust
    /// Push code hash at an address accessed through `EXTCOPYCOPY` to the block.
    pub fn add_ext_code_hash(&mut self, code_hash: Hash) {
        self.block.add_ext_code_hash(code_hash);
    }
```
* We can then also populate the bytecode table during [witness generation](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/witness/block.rs#L173-L188):
```rust
        bytecodes: block
            .txs()
            .iter()
            .flat_map(|tx| {
                tx.calls()
                    .iter()
                    .map(|call| call.code_hash)
                    .unique()
                    .into_iter()
                    .map(|code_hash| {
                        let bytecode =
                            Bytecode::new(code_db.0.get(&code_hash).cloned().unwrap_or_default());
                        (bytecode.hash, bytecode)
                    })
            })
            .chain(block.ext_code_hashes.iter().unique().map(|code_hash| {
                let bytecode =
                    Bytecode::new(code_db.0.get(&code_hash).cloned().unwrap_or_default());
                (bytecode.hash, bytecode)
            }))
            .collect(),
```

There might be other ways to do this as well, let me know what you guys think :)

roynalnaruto:
Also, with regards to the code here: https://github.com/DoHoonKim8/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/extcodecopy.rs#L27-L29

EVM should allow `EXTCODECOPY` from an address that doesn't exist in the state or that doesn't have code. The copy size should be all `0` bytes. So those cases should be handled through bus-mapping.

ChihChengLiang:
@roynalnaruto, the approach of populating bytecode to the table looks great. Thank you for the proposal!

aguzmant103:
Hey @DoHoonKim8 just following up, is this PR still in progress?

DoHoonKim8:
> Hey @DoHoonKim8 just following up, is this PR still in progress?

Yup, I will gonna resolve all the comments in the next week. Sorry for being late!

DoHoonKim8:
> @ChihChengLiang @DoHoonKim8 Since the bytecode lookup is failing, I believe we can start debugging from [here](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/copy_circuit.rs#L672-L674):
> 
> * Only `block.bytecodes` is populated into the bytecode table.
> 
> The witness block is constructed [here](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/witness/block.rs#L173-L188):
> 
> * Which means it is only considering the `code_hash` of each `call` in the block's transactions. It does not take into consideration the bytecode at the `external_address`.
> 
> One possible solution is to:
> 
> * Add a field `ext_code_hashes: Vec<Hash>` to the [circuit_input_builder::block](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/block.rs#L78-L79) and:
> * Populate it during the `bus-mapping` implementation of [`EXTCODECOPY`](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/extcodecopy.rs#L52)
> 
> ```rust
>     state.add_ext_code_hash(ext_code_hash)?;
> ```
> 
> which adds this to the block:
> 
> ```rust
>     /// Push code hash at an address accessed through `EXTCOPYCOPY` to the block.
>     pub fn add_ext_code_hash(&mut self, code_hash: Hash) {
>         self.block.add_ext_code_hash(code_hash);
>     }
> ```
> 
> * We can then also populate the bytecode table during [witness generation](https://github.com/DoHoonKim8/zkevm-circuits/blob/main/zkevm-circuits/src/witness/block.rs#L173-L188):
> 
> ```rust
>         bytecodes: block
>             .txs()
>             .iter()
>             .flat_map(|tx| {
>                 tx.calls()
>                     .iter()
>                     .map(|call| call.code_hash)
>                     .unique()
>                     .into_iter()
>                     .map(|code_hash| {
>                         let bytecode =
>                             Bytecode::new(code_db.0.get(&code_hash).cloned().unwrap_or_default());
>                         (bytecode.hash, bytecode)
>                     })
>             })
>             .chain(block.ext_code_hashes.iter().unique().map(|code_hash| {
>                 let bytecode =
>                     Bytecode::new(code_db.0.get(&code_hash).cloned().unwrap_or_default());
>                 (bytecode.hash, bytecode)
>             }))
>             .collect(),
> ```
> 
> There might be other ways to do this as well, let me know what you guys think :)

Hey thanks for your suggestion! I updated `witness::block::block_convert` function to populate bytecode table in witness block with external code hashes. Would you check the patches?

DoHoonKim8:
As in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1014#discussion_r1058391676, the whole bytecode in `code_db` is populated in block now(also the external bytecode too). 

As Han commented, it `code_db` is used for caching bytecodes, then we should follow https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720#issuecomment-1325853263 in the future.

aguzmant103:
Hey @roynalnaruto  based on last comment I think this is ready for review whenever you can

roynalnaruto:
> Hey @roynalnaruto based on last comment I think this is ready for review whenever you can

@DoHoonKim8 I was on leave. I'm back to work this week and will review the PR by end of the week :)

roynalnaruto (approved):
Some nitpicks. Overall looks good! You need to also do some clippy fixes after upstream was updated!

DoHoonKim8:
Thanks for your kind review!! 👍 @roynalnaruto

lispc:
@DoHoonKim8 oh there is one thing missing.. After https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1020, we should no longer write `let code_hash = block.rws[step.rw_indices[8]]
            .table_assignment_aux(block.randomness)
            .value;`, it should be changed like `self.code_hash
            .assign(region, offset, region.word_rlc(code_hash))?;` . you can have a look at extcodesize gagdet.

lispc:
another problem is the non exist proof refactor like this https://github.com/privacy-scaling-explorations/zkevm-circuits/commit/aaaf13ee3020661e048484776bd88686d1659042

lispc:
oh i enabled RUST_LOG=trace for this gadget, can i believe ErrorOutOfGasEXTCODECOPY is triggered.. The extcodecopy gadget may not be tested at all.. @DoHoonKim8

PR #722, opened 2022-08-31
fix: also add feature guard `dev` to mod `test` of copy_circuit

silathdiir:
Mod `test` uses the function `use super::dev::test_copy_circuit`, and [above mod dev](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/copy_circuit.rs#L683) is protected by feature guard `dev`.

~~So it seems that this test module also needs this feature guard `dev` (failed to compile test as `cargo test` without specified features in zkevm-circuits).~~

Deleted feature `dev` and added `#[cfg(any(feature = "test", test))]` to mod `dev` in `copy_circuit.rs` as @roynalnaruto commented as below.


DreamWuGit (approved):
LGTM!

DreamWuGit:
@roynalnaruto  WDYT ?

roynalnaruto:
@silathdiir Instead of adding the `dev` gate to that module, I think we should completely remove the `dev` feature altogether.

So we can do:
1. `zkevm-circuits/src/copy_circuit.rs`
```
-#[cfg(feature = "dev")]
+#[cfg(test)]
```
2. `zkevm-circuits/Cargo.toml`
```
-test = [ "dev" ]
-dev = []
+test = []
```
3. `integration-tests/Cargo.toml`
```
-zkevm-circuits = { path = "../zkevm-circuits", features = ["test", "dev"] }
+zkevm-circuits = { path = "../zkevm-circuits", features = ["test"] }
```

roynalnaruto (approved):
Looks good, modulo the clippy fix needed

silathdiir:
> Looks good, modulo the clippy fix needed

@roynalnaruto It seems that mod `dev` (of `copy_circuit.rs`) is also used in `integration-tests` which is not guarded with `cfg(test)` as the cargo-clippy issue.

Should we delete `#[cfg(test)]` on mod `dev`. WDYT?

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -680,7 +680,6 @@ impl<F: Field> CopyCircuit<F> {
     }
 }
 
-#[cfg(feature = "dev")]
```
roynalnaruto, line 683: I think this should be:

```
#[cfg(any(feature = "test", test))]
```

roynalnaruto (changes requested):
Just a minor point. Instead of having the `dev` mod compile all the time, it's better to use this suggestion.

This is because, `integration-tests` is a different crate, and the `test` modules are by default not enabled from another crate. So we are already using the `test` feature to overcome this. Although `cfg(test)` is different from `cfg(feature = "test")`. Enabling this makes sure we are able to compile `integration-tests` as well :)

zkevm-circuits/src/copy_circuit.rs
silathdiir: Added `#[cfg(any(feature = "test", test))]` to mod `dev` in `copy_circuit`.

roynalnaruto (approved):
Looks good to me

PR #728, opened 2022-09-01
pad evm circuit to make vk/pk universal

lispc:
EndBlock is something only meaningful for evm circuit? So i moved it from witness.rs to execution.rs.    Doing so makes it easier for (1) padding (2) handling empty block.


han0110 (approved):
> EndBlock is something only meaningful for evm circuit? So i moved it from witness.rs to execution.rs

Yes I think so. Nice work!

(btw if this PR is not urgent, let's wait for #724 merged and process this one)

CPerezz (approved):
LGTM!

PR #731, opened 2022-09-02
Self hosted runner for Integration Tests GH Action

ntampakas:
With this PR a self hosted runner is implemented for Integration Tests GH Action.

Note: This action fails using GH runner, due to memory starvation.


CPerezz (approved):
LGTM!!!

Let's see how this works in the next PRs! :smile:

ed255 (approved):
LGTM!  Looking forward to seeing the results again for these tests :)

PR #736, opened 2022-09-07
Apply balance update to `StateDb` in `gen_end_tx_ops`

han0110:
This PR resolves bug found by @adria0 through testing file [`refundSSTOREFiller.yml`](https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/stRefundTest/refundSSTOREFiller.yml), which is due to the balances of caller and coinbase are not updated to `StateDb` in `gen_end_tx_ops`.

It also updates `ExecStep::new` to record `refund` from `GethExecStep` (not the reason of the bug)


ed255 (approved):
LGTM! Nice bug finding @adria0!

han0110:
@adria0 Thanks for checking!

Issue #163, opened 2021-11-11
Replace fork of bitvec by upstream


ed255:
In https://github.com/appliedzkp/zkevm-circuits/pull/162 we introduce a fork of bitvec in order to solve an external dependency build error.  We have a PR for upstream to solve that: https://github.com/bitvecto-rs/bitvec/pull/141.  Once that PR is resolved, we can use bitvec from upstream again.

PR #745, opened 2022-09-07
Remove forked bitvec

ed255:
The compliation error comming from the funty dependency of bitvec isn't happening anymore, so we remove the patched dependency to the forked bitvec.

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/163


CPerezz (approved):
LGTM!!!

That's really nice! Thanks for keeping an eye on this @ed255 :tada:

PR #748, opened 2022-09-08
Official ethereum/tests StateTests execution (final)

adria0:
this PR supersedes #382  

# Note for reviewers

You will find in this PR:
- A binary in `evm-testvectors/src/main.rs` that
  - scans all the tests from the `evm-testvectors/test` folder that is a git submodule
     - there are two types of files to scan: `json` and `.yaml`
  - compiles the code inside the tests if not found in the `code.cache` file
     - this is executed via docker (see `docker` folder) 
  - runs the test if not found in the `results.cache` file, but skips some tests
     - which ones contains unimplemented opcodes (defined in the `OPCODES_UNIMPLEMENTED` const) via scanning the geth trace
     - which ones uses too much gas or a maximum of steps (see `StateTestConfig`)
     - which ones the geth trace is too big ; I had to modify `geth-utils/lib/lib.go` to skip very big traces (geth crashed)
     - others ( see `FILE_IGNORE_LIST`)
  - writes the result in the `results.cache` file 

- Enrich `Bytecode` by adding a `disasm() -> String` method as also an `iter()` that generates an iterator over the bytecode Opcodes. From my previous experience running those tests there are is a lot of cases that you have to analyze and test the code, and I created a small command line tool for generating on-the-fly tests using assembly/bytecode  as also to parse bytecode (see `--raw` option in `main.rs`)

-  **note** that this is a first version, probally this is going to change, and eventually we have to move this into a specific repo, but for the moment I prefer to merge this to not having code far from origin/main
- **note2** this code is not critical at this point, so a _light_  review could be done IMO

### DIFF
--- a/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
+++ b/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
@@ -390,6 +390,17 @@ impl<F: Field> Config<F> {
         witness: &[UnrolledBytecode<F>],
         randomness: F,
     ) -> Result<(), Error> {
+        self.assign_internal(layouter, size, witness, randomness, true)
+    }
+
+    pub(crate) fn assign_internal(
+        &self,
+        layouter: &mut impl Layouter<F>,
+        size: usize,
+        witness: &[UnrolledBytecode<F>],
+        randomness: F,
+        fail_fast: bool,
+    ) -> Result<(), Error> {
         let push_rindex_is_zero_chip = IsZeroChip::construct(self.push_rindex_is_zero.clone());
         let length_is_zero_chip = IsZeroChip::construct(self.length_is_zero.clone());
 
@@ -408,6 +419,15 @@ impl<F: Field> Config<F> {
                     let mut hash_input_rlc = F::zero();
                     let code_length = F::from(bytecode.bytes.len() as u64);
                     for (idx, row) in bytecode.rows.iter().enumerate() {
+                        if fail_fast && offset > last_row_offset {
+                            log::error!(
+                                "Bytecode Circuit: offset={} > last_row_offset={}",
+                                offset,
+                                last_row_offset
+                            );
+                            return Err(Error::Synthesis);
+                        }
+
                         // Track which byte is an opcode and which is push
                         // data
                         let is_code = push_rindex == 0;
--- a/zkevm-circuits/src/bytecode_circuit/dev.rs
+++ b/zkevm-circuits/src/bytecode_circuit/dev.rs
@@ -48,7 +48,13 @@ impl<F: Field> Circuit<F> for BytecodeCircuitTester<F> {
             self.bytecodes.iter().map(|b| &b.bytes),
             self.randomness,
         )?;
-        config.assign(&mut layouter, self.size, &self.bytecodes, self.randomness)?;
+        config.assign_internal(
+            &mut layouter,
+            self.size,
+            &self.bytecodes,
+            self.randomness,
+            false,
+        )?;
         Ok(())
     }
 }


testool/src/statetest/json.rs
```diff
@@ -0,0 +1,599 @@
...
+        let mut builder = JsonStateTestBuilder::new(&mut compiler);
+        builder.from_json(
+            "",
+            include_str!(
+                "../tests/src/GeneralStateTestsFiller/stCodeCopyTest/ExtCodeCopyTestsFiller.json"
+            ),
+        )?;
+        Ok(())
+    }
+    */
```
ChihChengLiang, line 538: Do we intend to check in these?

testool/src/statetest/suite.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        std::panic::set_hook(Box::new(|_info| {}));
+
+        log::debug!("running test {}/{}...", tc.path, tc.id);
+        let result = std::panic::catch_unwind(|| tc.clone().run(config.clone()));
+
+        // handle panic
+        let result = match result {
+            Ok(res) => res,
+            Err(_) => {
```
ChihChengLiang, line 113: nitpick
```suggestion
        if let Err(_) = result {
```

testool/src/testme.rs
```diff
@@ -0,0 +1,58 @@
+#[cfg(test)]
+mod test {
+    /*
```
ChihChengLiang, line 3: Do we intend to check in these?

ChihChengLiang (commented):
I added some feedbacks

testool/code.cache
```diff
@@ -0,0 +1,3042 @@
+16fe53eeefc1ba3d12cf051c6e4fcb098369c6655c94361790a2a618dd245eba=7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0160005500
```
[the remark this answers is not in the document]
ChihChengLiang: Would the name "codehash.txt" suit it?

testool/docker/lllc/patch.diff
```diff
@@ -0,0 +1,15 @@
...
+--- a/liblll/Parser.cpp
++++ b/liblll/Parser.cpp
+@@ -67,8 +67,8 @@ void dev::lll::debugOutAST(ostream& _out, sp::utree const& _this)
+ 
+ 		break;
+ 	case sp::utree_type::int_type: _out << _this.get<int>(); break;
+-	case sp::utree_type::string_type: _out << "\"" << _this.get<sp::basic_string<boost::iterator_range<char const*>, sp::utree_type::string_type>>() << "\""; break;
+-	case sp::utree_type::symbol_type: _out << _this.get<sp::basic_string<boost::iterator_range<char const*>, sp::utree_type::symbol_type>>(); break;
++	// case sp::utree_type::string_type: _out << "\"" << _this.get<sp::basic_string<boost::iterator_range<char const*>, sp::utree_type::string_type>>() << "\""; break;
++	// case sp::utree_type::symbol_type: _out << _this.get<sp::basic_string<boost::iterator_range<char const*>, sp::utree_type::symbol_type>>(); break;
```
ChihChengLiang, line 12: nitpick
```suggestion
```

testool/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+        // filter non-csv files and files from the same commit
+        let mut files: Vec<_> = std::fs::read_dir(REPORT_FOLDER)
+            .unwrap()
+            .filter_map(|f| {
+                let filename = f.unwrap().file_name().to_str().unwrap().to_string();
+                if filename.ends_with(".csv") && !filename.contains(&format!(".{}.", git_hash)) {
+                    Some(filename)
+                } else {
+                    None
+                }
```
ChihChengLiang, line 188: We have something like `bool::then` or `bool::then_some` for these
```
assert_eq!(false.then_some(0), None);
assert_eq!(true.then_some(0), Some(0));
```
```
assert_eq!(false.then(|| 0), None);
assert_eq!(true.then(|| 0), Some(0));
```
```suggestion
                (filename.ends_with(".csv") && !filename.contains(&format!(".{}.", git_hash))).then_some(filename)
```

testool/docker/lllc/patch.diff
ChihChengLiang: The intention is to remove the commented-off code from the patch. Keeping only the deletion part in the patch should do the trick.

testool/src/statetest/executor.rs
```diff
@@ -0,0 +1,490 @@
...
+    NonceMismatch { expected: U256, found: U256 },
+    #[error("CodeMismatch(expected: {expected:?}, found:{found:?})")]
+    CodeMismatch { expected: Bytes, found: Bytes },
+    #[error("StorgeMismatch(slot:{slot:?} expected:{expected:?}, found: {found:?})")]
+    StorageMismatch {
+        slot: U256,
+        expected: U256,
+        found: U256,
+    },
+    #[error("SkipTesstMaxGasLimit({0})")]
```
ed255, line 33:
```suggestion
    #[error("SkipTestMaxGasLimit({0})")]
```

ed255 (commented):
Overall looks very good! There's a lot of new code but I think the structure is clear and everything is easy to follow.  Please take a look at my comments, but I want to emphasize that there's no need to address all of them for this PR, some of them could be addressed in future PRs.
In particular:
- The comment about the Bytecode type, I think replacing `Bytecode` by `Vec<u8>` in the mock functions can be done in a future PR as it may touch many things.  For this PR I would just ask to rename `Bytecode::from_raw_unsafe` to `Bytecode::from_raw_unchecked`.
- The comment about the json/yaml parser:  So the parsers currently work, and the code is easy to follow.  My comment on them is to reduce verbosity if possible.  I don't think this needs to be addressed in this PR, we can take a look at it in the future :)

CODEOWNERS
```diff
@@ -12,4 +12,7 @@ integration-tests/* @ed255
 
 # Prs done to `zkevm-circuits/state-circuit` will require an extra review from @miha-stopar
 zkevm-circuits/src/state_circuit/ @miha-stopar
-zkevm-circuits/src/state_circuit.rs @miha-stopar
\ No newline at end of file
+zkevm-circuits/src/state_circuit.rs @miha-stopar
+
+# Prs done in `testool` will requiere an extra review from @adria0
+testool/ @adria0
```
ChihChengLiang, line 18: I think we have to change the line 1 from "@appliedzkp/zkevm-reviewers" to "@privacy-scaling-explorations/zkevm-reviewers"

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -117,7 +117,6 @@ impl<F: Field> EvmCircuit<F> {
     }
 
     /// Assign exact steps in block without padding for unit test purpose
-    #[cfg(any(feature = "test", test))]
```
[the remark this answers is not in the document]
ed255: Unresolving this, as I still see the deleted `#[cfg(any(feature = "test", test))]`.  I also see it deleted in line 151 of this file.

eth-types/src/bytecode.rs
```diff
@@ -144,6 +164,147 @@ impl Bytecode {
         });
         self
     }
+
+    /// Generate the diassembly
+    pub fn disasm(&self) -> String {
+        let mut asm = String::new();
+        let mut code_iter = self.code.iter();
+        while let Some(byte) = code_iter.next() {
```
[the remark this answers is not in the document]
ed255: > and then using `for_each`?
> 
> Sometimes I'm applying `for_each`, but at this moment I think that functional style matches well for transformations, but it's unclear for me for other use cases (yes, I'm aware that `into_par_iter()` in rayon), when I find imperative way more clear. Anyway...

Oh no, I didn't mean to write it in functional style; what I meant is to reuse the `BytecodeIterator` and the `OpcodeWithData::to_string()` to simplify this function like this:
```rust
    pub fn disasm(&self) -> String {
        let mut asm = String::new();
        for op in self.iter() {
            asm.push_str(&op.to_string());
        }
        asm
    }
```

PR #763, opened 2022-09-13
bytecode circuit: throw synthesis error if witness is too large

pinkiebell:
This helps in catching errors in circuit setup early. This change is important in tooling that does configure & synthesize the circuit to measure the circuit height but is not using the MockProver


ed255 (approved):
LGTM!  I think these kind of checks are very useful, so that we get the failure as early as possible :)

pinkiebell:
> LGTM! I think these kind of checks are very useful, so that we get the failure as early as possible :)

Thanks. I recognized an error if `!fail_fast` and committed a fix.

pinkiebell:
I don't know what's happening today, but the if-else clause was completely useless 😆

PR #766, opened 2022-09-14
Add mpt updates to the state circuit

### LINKED ISSUE #607: State circuit - MPT circuit integration

z2trillion:
Same functionality as https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/621, but built from scratch because of changes to main.

closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/607
specs: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/236


zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -193,14 +221,38 @@ impl<F: Field> ConstraintBuilder<F> {
...
+                vec![
+                    (
+                        q.rw_table.address.clone(),
+                        q.mpt_update_table.address.clone(),
+                    ),
+                    (
+                        q.rw_table.storage_key.clone(),
+                        q.mpt_update_table.storage_key.clone(),
+                    ),
+                    (q.field_tag(), q.mpt_update_table.field_tag.clone()),
```
miha-stopar, line 248: There was a recent change in the MPT lookup [specs](https://github.com/privacy-scaling-explorations/zkevm-specs/commit/a1457dc02da26797b1f25aed0197122898761bfc) to use MPT proof type instead of field tag.

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -249,8 +319,26 @@ impl<F: Field> ConstraintBuilder<F> {
...
```
miha-stopar, line 334: Same as above.

zkevm-circuits/src/witness/mpt.rs
```diff
@@ -0,0 +1,190 @@
+use crate::evm_circuit::{util::RandomLinearCombination, witness::Rw};
+use crate::table::AccountFieldTag;
+use eth_types::{Address, Field, ToLittleEndian, ToScalar, Word};
+use itertools::Itertools;
+use std::collections::HashMap;
+
+/// asdfasdf
```
ed255, line 7: I think you missed the docs in this file :stuck_out_tongue_closed_eyes:

zkevm-circuits/src/witness/mpt.rs
```diff
@@ -0,0 +1,190 @@
...
+        match self {
+            Self::Account { address, .. } | Self::AccountStorage { address, .. } => {
+                address.to_scalar().unwrap()
+            }
+        }
+    }
+    fn field_tag<F: Field>(&self) -> F {
+        match self {
+            Self::Account { field_tag, .. } => F::from(*field_tag as u64),
+            Self::AccountStorage { .. } => F::zero(),
```
ed255, line 129: Probably this function should be renamed to `proof_type` (following the update Miha mentioned).
And for AccountStorage, the result would be `F::from(MptProofType::StorageMod as u64)`

zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -104,7 +104,7 @@ fn test_state_circuit_ok(
 fn degree() {
     let mut meta = ConstraintSystem::<Fr>::default();
     StateCircuit::<Fr>::configure(&mut meta);
-    assert_eq!(meta.degree(), 9);
+    assert_eq!(meta.degree(), 12);
```
ed255, line 107: If I'm not mistaken, going from 9 to 12 means we bump the extended k from 3 to 4, which increases the proving time.

This is an open discussion: are we OK with this degree?  If not, do we have ideas on how to bring it down in the future?

ed255 (commented):
Overall looks good!  I've left some comments, please take a look :)

zkevm-circuits/src/witness/mpt.rs
```diff
@@ -0,0 +1,190 @@
...
+                ])
+            })
+            .collect()
+    }
+}
+
+impl MptUpdate {
+    pub(crate) fn value_assignments<F: Field>(&self, word_randomness: F) -> (F, F) {
+        let assign = |x: Word| {
+            let mut result =
```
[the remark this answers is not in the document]
z2trillion: that worked. i learned something new about the match syntax.

zkevm-circuits/src/state_circuit/test.rs
z2trillion: You're right about `k`.

I think it should be possible to reduce the degree back down to 9, but I'd have to put some more work into (re)unifying the mpt lookups for Account and AccountStorage rows. I'd prefer to put this off for now :)

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -154,6 +169,13 @@ impl<F: Field> ConstraintBuilder<F> {
         );
         self.require_zero("Start value is 0", q.value());
         self.require_zero("Start initial_value is 0", q.initial_value());
+        self.condition(q.lexicographic_ordering_selector.clone(), |cb| {
```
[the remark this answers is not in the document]
z2trillion: Yeah, that's no longer correct. It should now be row.rw_counter ==  prev_row.rw_counter + 1.

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -249,8 +319,26 @@ impl<F: Field> ConstraintBuilder<F> {
...
+                        q.mpt_update_table.storage_key.clone(),
+                    ),
+                    (q.field_tag(), q.mpt_update_table.field_tag.clone()),
+                    (q.state_root(), q.mpt_update_table.new_root.clone()),
+                    (q.state_root_prev(), q.mpt_update_table.old_root.clone()),
+                    (q.value(), q.mpt_update_table.new_value.clone()),
+                    (q.initial_value(), q.mpt_update_table.old_value.clone()),
+                ],
+            );
+        });
```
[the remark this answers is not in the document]
z2trillion: I added `build_tx_receipt_constraints`.

zkevm-circuits/src/state_circuit/constraint_builder.rs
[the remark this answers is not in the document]
ed255: Thanks, I've opened an issue to update the spec https://github.com/privacy-scaling-explorations/zkevm-specs/issues/275

zkevm-circuits/src/state_circuit/constraint_builder.rs
miha-stopar: I had in mind adding these new types that you mention, but I agree this might be better to add in a new PR.

zkevm-circuits/src/state_circuit/constraint_builder.rs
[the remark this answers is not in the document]
ed255: thanks!

zkevm-circuits/src/state_circuit/test.rs
ed255: > You're right about `k`.
> 
> I think it should be possible to reduce the degree back down to 9, but I'd have to put some more work into (re)unifying the mpt lookups for Account and AccountStorage rows. I'd prefer to put this off for now :)

Of course!  I've opened an issue to track this https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/772 :)

ed255 (approved):
LGTM! Nice work!

miha-stopar (approved):
Nice! LGTM!

PR #770, opened 2022-09-15
Use `Challenge` api for Bytecode circuit

### LINKED ISSUE #753: Refactor Bytecode circuit with challenge API
Using the challenge API instead of the instance value for randomness in Bytecode circuit

Notes:

- The randomness will be provided from the super circuit instead of being queried from within the Keccak circuit. This applies to all the sub circuits
- Treat the randomness as an expression, need a small test.
- The circuit needs to be divided into 2 phases: one with the commitments that don't use the randomness; and another one that has access to the randomness and commits values that use it (via RLC).

han0110:
This PR aims to refactor Bytecode circuit to use `Challenge` api provided by `halo2`.

It takes a naive approach to assume we can use static phases (no need to be configured manually) among all circuits, so we don't need to alternate the `*Table::construct` too much.

Note that only in those tests in module `bytecode_circuit` the Bytecode circuit is used with real challenge, for other integrate tests because other circuits are not ready, so we use mock challenge for Bytecode circuit as usual.

Resolves #753.

### DIFF
--- a/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
+++ b/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
@@ -3,7 +3,7 @@ use crate::{
         and, constraint_builder::BaseConstraintBuilder, not, or, select, RandomLinearCombination,
     },
     table::{BytecodeFieldTag, BytecodeTable, DynamicTableColumns, KeccakTable},
-    util::Expr,
+    util::{Challenges, Expr},
 };
 use bus_mapping::evm::OpcodeId;
 use eth_types::{Field, ToLittleEndian, Word};
@@ -20,7 +20,7 @@ use super::param::PUSH_TABLE_WIDTH;
 /// Public data for the bytecode
 #[derive(Clone, Debug, PartialEq)]
 pub(crate) struct BytecodeRow<F: Field> {
-    code_hash: F,
+    code_hash: Word,
     tag: F,
     index: F,
     is_code: F,
@@ -37,7 +37,6 @@ pub struct UnrolledBytecode<F: Field> {
 #[derive(Clone, Debug)]
 /// Bytecode circuit configuration
 pub struct Config<F> {
-    randomness: Expression<F>,
     minimum_rows: usize,
     q_enable: Column<Fixed>,
     q_first: Column<Fixed>,
@@ -60,9 +59,9 @@ pub struct Config<F> {
 impl<F: Field> Config<F> {
     pub(crate) fn configure(
         meta: &mut ConstraintSystem<F>,
-        randomness: Expression<F>,
         bytecode_table: BytecodeTable,
         keccak_table: KeccakTable,
+        challenges: Challenges<Expression<F>>,
     ) -> Self {
         let q_enable = meta.fixed_column();
         let q_first = meta.fixed_column();
@@ -157,9 +156,9 @@ impl<F: Field> Config<F> {
                 ),
             );
             cb.require_equal(
-                "hash_input_rlc := hash_input_rlc_prev * randomness + byte",
+                "hash_input_rlc := hash_input_rlc_prev * challenges.keccak_input + byte",
                 meta.query_advice(hash_input_rlc, Rotation::cur()),
-                meta.query_advice(hash_input_rlc, Rotation::prev()) * randomness.clone()
+                meta.query_advice(hash_input_rlc, Rotation::prev()) * challenges.keccak_input()
                     + meta.query_advice(value, Rotation::cur()),
             );
             cb.require_equal(
@@ -362,7 +361,6 @@ impl<F: Field> Config<F> {
         });
 
         Config {
-            randomness,
             minimum_rows: meta.minimum_rows(),
             q_enable,
             q_first,
@@ -388,9 +386,9 @@ impl<F: Field> Config<F> {
         layouter: &mut impl Layouter<F>,
         size: usize,
         witness: &[UnrolledBytecode<F>],
-        randomness: F,
+        challenges: &Challenges<Value<F>>,
     ) -> Result<(), Error> {
-        self.assign_internal(layouter, size, witness, randomness, true)
+        self.assign_internal(layouter, size, witness, challenges, true)
     }
 
     pub(crate) fn assign_internal(
@@ -398,7 +396,7 @@ impl<F: Field> Config<F> {
         layouter: &mut impl Layouter<F>,
         size: usize,
         witness: &[UnrolledBytecode<F>],
-        randomness: F,
+        challenges: &Challenges<Value<F>>,
         fail_fast: bool,
     ) -> Result<(), Error> {
         let push_rindex_is_zero_chip = IsZeroChip::construct(self.push_rindex_is_zero.clone());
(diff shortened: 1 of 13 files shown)


zkevm-circuits/src/util.rs
```diff
@@ -34,3 +36,62 @@ pub fn power_of_randomness_from_instance<F: Field, const N: usize>(
...
+        Self {
+            evm_word: meta.challenge_usable_after(FirstPhase),
+            keccak_input: meta.challenge_usable_after(FirstPhase),
+        }
+    }
+
+    /// Get `Expression`s from `Layouter` and returns expressions of challenges.
+    pub fn exprs<F: Field>(&self, meta: &mut ConstraintSystem<F>) -> Challenges<Expression<F>> {
+        let mut challenges = array::from_fn(|_| None);
+        meta.create_gate("power of randomness from instance", |meta| {
```
ChihChengLiang, line 59: 1. Does "power of randomness from instance" describe what we are doing here?
2. Maybe not in this PR but in a future one, is there a way to abstract away this create_gate trick to get expressions from ConstraintSystem?

zkevm-circuits/src/util.rs
```diff
@@ -34,3 +36,62 @@ pub fn power_of_randomness_from_instance<F: Field, const N: usize>(
...
+impl Challenges {
+    /// Construct `Challenges` by allocating challenges in specific phases.
+    pub fn construct<F: Field>(meta: &mut ConstraintSystem<F>) -> Self {
+        Self {
+            evm_word: meta.challenge_usable_after(FirstPhase),
+            keccak_input: meta.challenge_usable_after(FirstPhase),
+        }
+    }
+
+    /// Get `Expression`s from `Layouter` and returns expressions of challenges.
```
ChihChengLiang, line 56: We have no layouter here right?
```suggestion
    /// Get `Expression`s from `ConstraintSystem` and returns expressions of challenges.
```

ChihChengLiang (approved):
Great refactoring! This PR also demonstrates how to use challenge API in the circuit context.
Added some comments. LGTM after comments are addressed.

zkevm-circuits/src/util.rs
han0110: 1. No, I copied but forgot to update the description, thanks for catching this!
2. Yes, let me add a wrapper do that.

han0110: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/770/commits/96fa4e253365a556fec718ed3550c568f5b4dbd5, and add a helper `query_expression` to avoid create dummy gate manually.

zkevm-circuits/src/util.rs
han0110: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/770/commits/96fa4e253365a556fec718ed3550c568f5b4dbd5

DreamWuGit:
> It takes a naive approach to assume we can use static phases (no need to be configured manually) among all circuits, so we don't need to alternate the `*Table::construct` too much.

@han0110 I am curious about meaning of `static phases` here.  is there any doc for it static phase & dynamic phase ? thanks !

han0110:
static phases means not configurable from Super circuit or other circuits, even there is a phase in the middle that is not necessary (I believe we don't have this issue for now).

zkevm-circuits/src/bytecode_circuit/dev.rs
```diff
@@ -30,29 +24,35 @@ impl<F: Field> Circuit<F> for BytecodeCircuitTester<F> {
...
+        (config, challenges)
     }
 
     fn synthesize(
         &self,
-        config: Self::Config,
+        (config, challenges): Self::Config,
         mut layouter: impl Layouter<F>,
     ) -> Result<(), Error> {
+        let challenges = challenges.values(&mut layouter);
```
ed255, line 43: If I'm not mistaken, the challenges used in the mock prover are deterministic and random looking (generated here https://github.com/han0110/halo2/blob/346afce254258361f65a4e298f8636a52871cb6d/halo2_proofs/src/dev.rs#L544)

Recently while debugging a lookup error it was useful to me to use randomness `0x100` so that I could see the original values after an RLC (where the original values were smaller than 256 each).  I think it would be useful to allow specifying custom challenge values for the mock proving for debugging.

ed255 (approved):
LGTM!  Very nice work!  I've left some questions but none of them require any change.

zkevm-circuits/src/table.rs
```diff
@@ -539,12 +541,14 @@ pub struct BytecodeTable {
 impl BytecodeTable {
     /// Construct a new BytecodeTable
     pub fn construct<F: Field>(meta: &mut ConstraintSystem<F>) -> Self {
+        let [tag, index, is_code, value] = array::from_fn(|_| meta.advice_column());
```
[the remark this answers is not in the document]
han0110: Yes!

zkevm-circuits/src/util.rs
```diff
@@ -23,14 +36,62 @@ pub fn power_of_randomness_from_instance<F: Field, const N: usize>(
...
     let columns = [(); N].map(|_| meta.instance_column());
-    let mut power_of_randomness = None;
+    query_expression(meta, |meta| {
+        columns.map(|column| meta.query_instance(column, Rotation::cur()))
+    })
+}
+
+/// All challenges used in `SuperCircuit`.
+#[derive(Clone, Copy, Debug)]
+pub struct Challenges<T = Challenge> {
```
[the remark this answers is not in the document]
han0110: Yes, that's a default generic type parameter, without specifying it uses `halo2::plonk::Challenge` as `T` (see https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#default-generic-type-parameters-and-operator-overloading).

So I thought it's annoying to type `Challenges<Challenge>` so I added this default type, if you feel it's better to always specify the generic type, I can roll it back.

zkevm-circuits/src/bytecode_circuit/dev.rs
han0110: This makes sense, but this requires to update `pse/halo2`, I have created an feature request issue https://github.com/privacy-scaling-explorations/halo2/issues/101 for this.

davidnevadoc:
LGTM! Nice work :+1:

davidnevadoc (approved):
LGTM! Nice work :+1:

zkevm-circuits/src/util.rs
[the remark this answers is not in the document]
ed255: I think that's perfect!  It's just a feature of Rust I didn't know about :P

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -157,9 +156,9 @@ impl<F: Field> Config<F> {
                 ),
             );
             cb.require_equal(
-                "hash_input_rlc := hash_input_rlc_prev * randomness + byte",
+                "hash_input_rlc := hash_input_rlc_prev * challenges.keccak_input + byte",
                 meta.query_advice(hash_input_rlc, Rotation::cur()),
-                meta.query_advice(hash_input_rlc, Rotation::prev()) * randomness.clone()
+                meta.query_advice(hash_input_rlc, Rotation::prev()) * challenges.keccak_input()
```
ed255, line 161: It seems that `hash_input_rlc` depends on the challange.  Should it be an advice in `SecondPhase`?

han0110: Yes, it should be in `SecondPhase`, will test with real prover then submit a fix soon.

PR #773, opened 2022-09-16
Fix order of block hash assignment

### LINKED ISSUE #716: In Circuit Input Builder, store the block hashes in expected order
See: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/709#pullrequestreview-1088202123

In the block table we keep the list of the last 256 block hashes https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#block_table with an index
The Circuit Input Builder should return the list of block hashes sorted by their indexes, so that we don't need to `rev()` it later.

swjng:
Fix #716 
[Handling `history_hashes` in circuit input builder](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/block.rs#L48) is valid, but [assigning block table](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/witness/block.rs#L114) has invalid order. Lastest block hash is stored in last element of `history_hashes`. But in assignment, first element of `history_hashes` is used for lastest block.
So imo, just fixing order in assignment can solve this issue. Are there any other issues in that?

### DIFF
--- a/zkevm-circuits/src/witness/block.rs
+++ b/zkevm-circuits/src/witness/block.rs
@@ -104,21 +104,23 @@ impl BlockContext {
                     ),
                 ],
             ],
-            self.history_hashes
-                .iter()
-                .rev()
-                .enumerate()
-                .map(|(idx, hash)| {
-                    [
-                        F::from(BlockContextFieldTag::BlockHash as u64),
-                        (self.number - idx - 1).to_scalar().unwrap(),
-                        RandomLinearCombination::random_linear_combine(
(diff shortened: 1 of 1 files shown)


han0110 (approved):
LGTM! Thanks for fixing this!

jeong0982:
@han0110 Can this PR be merged?

han0110:
Merge with only 1 review to avoid more rebases since it should be an obvious one.

PR #775, opened 2022-09-16
[Do not merge] parse prestate and block

ChihChengLiang:
This is work in progress attempt to reproduce #723 

Currently, I got this error, but I don't think I'm at the right place yet.
```
  left: `70683092540792552306903383139300639849103465775893174946906595960976200455679`,
 right: `94323640787704833666008867169815180691409683780709829507833618832391461736655`', bus-mapping/src/evm/opcodes/sha3.rs:47:9
```
Run `cargo run -p zkevm-circuits --features test` if you like to check.

cc @CPerezz @pinkiebell @AronisAt79 @han0110


ChihChengLiang:
Currently stuck at `thread 'main' panicked at 'arithmetic operation overflow', /Users/liangcc/.cargo/registry/src/github.com-1ecc6299db9ec823/primitive-types-0.11.1/src/lib.rs:38:1`

ChihChengLiang:
If I attempted to override traces I get this error message.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AccountNotFound(0x138ee18fb1d5a584fac4b086baa2da3b2defde47)', /Users/liangcc/projects/repos/zkevm-circuits/zkevm-circuits/src/test_util.rs:50:10
```
But strange, looking from the state data, it seems 0x138ee18f is created after some txs have been run.

ChihChengLiang:
@pinkiebell what's the error you encountered? I tried your patch [62fa9af](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/775/commits/62fa9af85cba46900f7c830d798357320f93f7e2) and found that ... the test passed! 
I also tried to apply to zkevm-circuit revision `d16ac9b8bda39c45c82a41d2d59054c85abbc9a1`, which is the version reported on #723. But that was a long commit history ago, it would need some work to make the test code here to build on there.

pinkiebell:
> @pinkiebell what's the error you encountered? I tried your patch [62fa9af](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/775/commits/62fa9af85cba46900f7c830d798357320f93f7e2) and found that ... the test passed!

I got a `arithmetic overflow`.
This was because `history_hashes` is a proper 256 items long but the code path didn't expected that.  [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/fix-zkevm-chain-error/zkevm-circuits/src/witness/block.rs#L114) . Sorry I messed up the line number before 😁 

> I also tried to apply to zkevm-circuit revision `d16ac9b8bda39c45c82a41d2d59054c85abbc9a1`, which is the version reported on #723. But that was a long commit history ago, it would need some work to make the test code here to build on there.

I think we can test this on the latest main. If that error doesn't happen anymore, then it was fixed in between.

ChihChengLiang:
When I rebased to main, the code works. So I think this is fixed.

pinkiebell:
> When I rebased to main, the code works. So I think this is fixed.

Oh wait, did you test w/ the keccak circuit?. Because looking at the test tool it only uses StateCircuit.

ChihChengLiang:
I'd say EvmCircuit is all we need. The prover log shows the error comes from the region "Execution step", which is part of the EVMCircuit. The name of the lookup, "Keccak", should come from the stringified "Table::Keccak" located here. https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/55c6dfbc64a32ee23820ae7119b27239a7e2ad3f/zkevm-circuits/src/evm_circuit/execution.rs#L704

PR #780, opened 2022-09-19
feat: Add helper function to calcualte TxLog address key

### LINKED ISSUE #768: Add a helper function to calculate TxLog address key
In the RwTable, for the TxLog entries we're packing several values into the "address" key like `index | field_tag | log_id`.
In several places we have code like this to calculate this key:
```rust
U256::from(*index as u64)
    + (U256::from(*field_tag as u64) << 32)
    + (U256::from(*log_id) << 48))
    .to_address()
```
This appears in
- zkevm-circuits/src/witness/rw.rs
- zkevm-circuits/src/table.rs
- zkevm-circuits/src/copy_circuit.rs

It also appears in the form of expression in
- zkevm-circuits/src/evm_circuit/execution/logs.rs
- zkevm-circuits/src/evm_circuit/util/constraint_builder.rs

We can introduce helper functions to calculate this value and expression

leolara:
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/768

Questions @ed255 :
 + I am not sure about the function name and the module I put it in, please suggest if you think something better.
 + You mention something about the files zkevm-circuits/src/evm_circuit/execution/logs.rs and zkevm-circuits/src/evm_circuit/util/constraint_builder.rs but I didn't understand it
 + I put parameter field_tag not as a borrow because it has the copy trait, I hope that is the right thing. Is it?

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/logs.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/logs.rs
@@ -126,7 +126,11 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         );
 
         let copy_rwc_inc = cb.query_cell();
-        let dst_addr = build_tx_log_expression(0, TxLogFieldTag::Data, &cb.curr.state.log_id);
+        let dst_addr = build_tx_log_expression(
+            0.expr(),
+            TxLogFieldTag::Data.expr(),
+            cb.curr.state.log_id.expr(),
+        );
         let cond = memory_address.has_length() * is_persistent.expr();
         cb.condition(cond.clone(), |cb| {
             cb.copy_table_lookup(
--- a/zkevm-circuits/src/evm_circuit/execution/logs.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/logs.rs
@@ -129,7 +129,7 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         let dst_addr = build_tx_log_expression(
             0.expr(),
             TxLogFieldTag::Data.expr(),
-            cb.curr.state.log_id.expr(),
+            cb.curr.state.log_id.expr() + 1.expr(),
         );
         let cond = memory_address.has_length() * is_persistent.expr();
         cb.condition(cond.clone(), |cb| {
--- a/zkevm-circuits/src/evm_circuit/execution/logs.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/logs.rs
@@ -15,7 +15,7 @@ use crate::{
         witness::{Block, Call, ExecStep, Transaction},
     },
     table::{CallContextFieldTag, RwTableTag, TxLogFieldTag},
-    util::Expr,
+    util::{build_tx_log_expression, Expr},
 };
 use array_init::array_init;
 use bus_mapping::circuit_input_builder::CopyDataType;
@@ -126,8 +126,7 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         );
 
         let copy_rwc_inc = cb.query_cell();
-        let dst_addr = (1u64 << 32).expr() * TxLogFieldTag::Data.expr()
-            + (1u64 << 48).expr() * (cb.curr.state.log_id.expr() + 1.expr());
+        let dst_addr = build_tx_log_expression(0, TxLogFieldTag::Data, cb.curr.state.log_id);
         let cond = memory_address.has_length() * is_persistent.expr();
         cb.condition(cond.clone(), |cb| {
             cb.copy_table_lookup(
--- a/zkevm-circuits/src/evm_circuit/execution/logs.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/logs.rs
@@ -126,7 +126,7 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         );
 
         let copy_rwc_inc = cb.query_cell();
-        let dst_addr = build_tx_log_expression(0, TxLogFieldTag::Data, cb.curr.state.log_id);
+        let dst_addr = build_tx_log_expression(0, TxLogFieldTag::Data, &cb.curr.state.log_id);
         let cond = memory_address.has_length() * is_persistent.expr();
         cb.condition(cond.clone(), |cb| {
             cb.copy_table_lookup(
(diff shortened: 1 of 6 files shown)


ed255:
>     * I am not sure about the function name and the module I put it in, please suggest if you think something better.

Naming is hard haha.  I think `build_tx_log_address` is good :)

>     * You mention something about the files zkevm-circuits/src/evm_circuit/execution/logs.rs and zkevm-circuits/src/evm_circuit/util/constraint_builder.rs but I didn't understand it

There's another place where we build a `tx_log_address`, but using different types:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/673f5fe6e9e76ce3796a3f5b1d15d7c49ced66cd/zkevm-circuits/src/evm_circuit/execution/logs.rs#L129-L130

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/673f5fe6e9e76ce3796a3f5b1d15d7c49ced66cd/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs#L1058

Since those are different types, a new function will be required.  It could be something like:
```rust
pub(crate) fn build_tx_log_address_expr<F: Field>(index: Expression<F>, field_tag: Expression<F>, log_id: Expression<F>) -> Expression<F> {
    // ...
}
```

>     * I put parameter field_tag not as a borrow because it has the copy trait, I hope that is the right thing. Is it?

That's perfect!  The `TxLogFieldTag` is a very small type so it makes sense to copy it instead of borrowing it; specially since it has the Copy trait already.

zkevm-circuits/src/util.rs
```diff
@@ -34,3 +36,28 @@ pub fn power_of_randomness_from_instance<F: Field, const N: usize>(
 
     power_of_randomness.unwrap()
 }
+
+pub(crate) fn build_tx_log_address(index: u64, field_tag: TxLogFieldTag, log_id: u64) -> Address {
+    (U256::from(index) + (U256::from(field_tag as u64) << 32) + (U256::from(log_id) << 48))
+        .to_address()
+}
+
+pub(crate) fn build_tx_log_from_expr_to_expression<F: halo2_proofs::arithmetic::FieldExt>(
```
ed255, line 45:
```suggestion
pub(crate) fn build_tx_log_from_expr_to_expression<F: Field>(
```
Can you check if it works this way?  Usually we're using the `Field` trait (which I think contains the `FieldExt`)

ed255, line 45: I think having a single function for the tx log expression is enough.  Could you rename this one to `build_tx_log_expression` and delete the second one?

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -126,8 +126,7 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         );
 
         let copy_rwc_inc = cb.query_cell();
-        let dst_addr = (1u64 << 32).expr() * TxLogFieldTag::Data.expr()
-            + (1u64 << 48).expr() * (cb.curr.state.log_id.expr() + 1.expr());
+        let dst_addr = build_tx_log_expression(0, TxLogFieldTag::Data, cb.curr.state.log_id);
```
ed255: When used with the function that only takes expressions, this would look like:

`build_tx_log_expression(1.expr(), TxLogFieldTag::Data.expr(), cb.curr.state.log_id)`

Note that originally this `dst_addr` used `1` as index.  I don't know if that's correct, but I'd leave it the same to keep the same behavior.

leolara:
There are 3 cases:
 + u64, TxLogFieldTag, u64 -> Address
 + Expression, Expression, Expression -> Expression
 + u64 (can be 0), TxLogFieldTag, &Cell<F> -> Expression
 
 Perhaps The last interface could be eliminated and call the second one converting to Expression outside. But then there is the case where there is no index, I am not sure if adding a `0.expr() +` is bad

zkevm-circuits/src/util.rs
leolara: ok

ed255:
> @ed255 I don't know if I complicated the interface too much with three functions. I am not sure how expresssions are used, and I was afraid that adding a `0.expr() +` or doing conversions between expressions and integer and back to expression could be inefficient in the circuit. But I am not sure, please let me know if there is a better interface.

Having a `0.expr() + ...` is OK!
Conversion from integer to expression can be done without a problem.  The way back I think is not possible.

leolara:
@ed255 please check again :-)

zkevm-circuits/src/evm_circuit/execution/logs.rs
ed255: Actually I got mistaken, that `1` is not the index, it's actually that the log_id value is `cb.curr.state.log_id.log_id + 1`!  Sorry about that!

leolara:
@ed255 let's see if now it is all right

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -126,8 +126,11 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
         );
 
         let copy_rwc_inc = cb.query_cell();
-        let dst_addr = (1u64 << 32).expr() * TxLogFieldTag::Data.expr()
-            + (1u64 << 48).expr() * (cb.curr.state.log_id.expr() + 1.expr());
+        let dst_addr = build_tx_log_expression(
+            0.expr(),
+            TxLogFieldTag::Data.expr(),
+            cb.curr.state.log_id.expr(),
```
ed255, line 132:
```suggestion
            cb.curr.state.log_id.expr() + 1.expr(),
```

ed255 (changes requested):

ed255:
> @ed255 let's see if now it is all right

Please apply this change https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/780#discussion_r975358157

Also we require the committed code to be formatted with Rust fmt so that the formatting is always consistent.  Make sure you run `cargo fmt` on all your edited files (info https://github.com/rust-lang/rustfmt)  I suggest checking if your code editor supports calling `cargo fmt` every time you save a file for convenience :)

zkevm-circuits/src/evm_circuit/execution/logs.rs
leolara: I was conviced I did it :exploding_head:

leolara: I did it but didn't commit it

ed255 (approved):
LGTM! Thanks for the improvement!

ChihChengLiang (approved):
LGTM

PR #787, opened 2022-09-20
git: ignore submodules by default

pinkiebell:
Tools like `Cargo` will fetch dependencies w/ all submodules. Since the latest integration of ethereums-tests, this became a significant operation on CI runs. Luckily, cargo respects `submodule.<name>.update = none`.

Hence, submodules will be skipped by git but this can be explitly overriden via `git submodule update --init --checkout` - the `--checkout` flag is the important one.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,9 @@
 [submodule "evm-testvectors/tests"]
+	update = none
 	path = testool/tests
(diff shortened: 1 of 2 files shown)


lispc (approved):

ed255 (approved):
LGTM!

PR #792, opened 2022-09-21
Fix phase of `hash_input_rlc` in Bytecode circuit

han0110:
This PR aims to fix the phase of `hash_input_rlc` in Bytecode circuit cause it's containing value that depends on challenge usable after phase 1.

Bytecode circuit in `main` after #770 was just tested locally with real prover and failed expectedly, after this PR it's working again.

Thanks for @ed255 to spot that mistake.


ed255 (approved):
LGTM!

Issue #528, opened 2022-05-24
Merge matching structs between circuit input builder and evm circuit witness (where possible)


ed255:
The idea of having duplicate structs between circuit input builder and evm witness generation was to allow the development of the circuit input builder in parallel with the circuits while minimizing the changes that one implies to the other. But once things stabilize, I think we can merge those.

Related to https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/484#discussion_r875590637

See the latest discussions upon that:
> Unify the types `Block`,` Call`, `Step`, `Transaction` between these two:
>
>    - zkevm-circuits/src/witness
>    - bus-mapping/src/circuit_input_builder

Also see: 
> Remove duplicated fields in eth_block -> https://github.com/appliedzkp/zkevm-circuits/blob/feat%2Fannotate_evm_cols/bus-mapping/src/circuit_input_builder/block.rs#L54-L57
> (Work on refactoring and have just 2 blocks, 1 for RPC and the other for general purpose)

DreamWuGit:
Really good idea to reduce redundant  structs !!!

ChihChengLiang:
I can take this one.

ed255:
There are still some structs that can be de-duplicated after the PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1378 @ChihChengLiang will open a new issue to track those.

PR #795, opened 2022-09-22
EndBlock

### LINKED ISSUE #575: Implement EndBlock execution state
Currently we have an EndBlock gadget that is dummy (it doesn't have any constraints).

The EndBlock state was specified here https://github.com/privacy-scaling-explorations/zkevm-specs/pull/88 , nevertheless there are a few details that may be missing from the spec:
```python
# TODO: Introduce constrain_instance to constrain the equality between witness
#       and public input, for total_tx and total_rw
```
- Constraint that the number of tx processed in the EVM is the same as the number of txs in the tx table
- Constraint that the number of ops in the rw_table from the State circuit match the number of ops "emitted" from the EVM circuit.

Dependency:

- [x] [#711](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/711)

### LINKED ISSUE #587: Add lookups to EVM circuit to constrain number of padding and active rows
Once  #322 is done, we need to do add two lookups to the EVM circuit to prove that there exactly `n` active rows and `N_ROWS-n` padding rows in the rw table:
1. lookup(Start, rwc=1)
2. lookup(Start, rwc=N_ROWS-n)

See https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/558#issuecomment-1160765938 for context.

ed255:
- Implement EndBlock following updated spec at https://github.com/privacy-scaling-explorations/zkevm-specs/pull/286
- Update BeginTx so that it forbids Transaction.CallerAccount != 0 (in
  order to support Padding Txs as described in
  https://github.com/privacy-scaling-explorations/zkevm-specs/issues/273#issuecomment-1246837816)
- Update state transition constraints to support empty block.
- Extend ConstraintBuilder with:
    - Constraint location (step_first, step_last, step_any)
    - Add copy cells and fixed column to store constants to be accessed
      via copy constraints at fixed offsets.

Depends on:
- [x] https://github.com/privacy-scaling-explorations/zkevm-specs/pull/286

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/575
Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/587

### DIFF
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -731,7 +731,7 @@ mod tests {
         )
         .unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(
+        let mut builder = BlockData::new_from_geth_data_with_params(
             block.clone(),
             CircuitsParams {
                 max_rws: 8192,
@@ -755,8 +755,7 @@ mod tests {
         };
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder =
-            BlockData::new_from_geth_data_default(block.clone()).new_circuit_input_builder();
+        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
@@ -767,7 +766,7 @@ mod tests {
         let (code, _) = gen_sha3_code(0x20, 0x200, MemoryKind::EqualToSize);
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(
+        let mut builder = BlockData::new_from_geth_data_with_params(
             block.clone(),
             CircuitsParams {
                 max_rws: 2000,
@@ -794,8 +793,7 @@ mod tests {
         };
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder =
-            BlockData::new_from_geth_data_default(block.clone()).new_circuit_input_builder();
+        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -596,6 +596,7 @@ pub mod dev {
     #[derive(Default)]
     struct CopyCircuitTester<F> {
         block: Block<F>,
+        max_txs: usize,
         randomness: F,
     }
 
@@ -605,7 +606,11 @@ pub mod dev {
         }
 
         pub fn new(block: Block<F>, randomness: F) -> Self {
-            Self { block, randomness }
+            Self {
+                block,
+                randomness,
+                max_txs: 4,
+            }
         }
         pub fn r() -> Expression<F> {
             123456u64.expr()
@@ -653,13 +658,16 @@ pub mod dev {
         ) -> Result<(), halo2_proofs::plonk::Error> {
             let challenges = Challenges::mock(Value::known(self.randomness));
 
-            config
-                .tx_table
-                .load(&mut layouter, &self.block.txs, self.randomness)?;
+            config.tx_table.load(
+                &mut layouter,
+                &self.block.txs,
+                self.max_txs,
+                self.randomness,
+            )?;
             config.rw_table.load(
                 &mut layouter,
                 &self.block.rws.table_assignments(),
-                self.block.state_circuit_pad_to,
+                self.block.circuits_params.max_rws,
                 self.randomness,
             )?;
             config.bytecode_table.load(
@@ -689,7 +697,10 @@ pub mod dev {
 mod tests {
     use super::dev::test_copy_circuit;
     use bus_mapping::evm::{gen_sha3_code, MemoryKind};
-    use bus_mapping::{circuit_input_builder::CircuitInputBuilder, mock::BlockData};
+    use bus_mapping::{
+        circuit_input_builder::{CircuitInputBuilder, CircuitsParams},
+        mock::BlockData,
+    };
     use eth_types::{bytecode, geth_types::GethData, Word};
     use mock::test_ctx::helpers::account_0_code_account_1_no_code;
     use mock::TestContext;
@@ -720,7 +731,14 @@ mod tests {
         )
         .unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        let mut builder = BlockData::new_from_geth_data(
+            block.clone(),
+            CircuitsParams {
+                max_rws: 8192,
+                ..Default::default()
+            },
+        )
+        .new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
@@ -737,7 +755,8 @@ mod tests {
         };
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        let mut builder =
+            BlockData::new_from_geth_data_default(block.clone()).new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
@@ -748,7 +767,14 @@ mod tests {
         let (code, _) = gen_sha3_code(0x20, 0x200, MemoryKind::EqualToSize);
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        let mut builder = BlockData::new_from_geth_data(
+            block.clone(),
+            CircuitsParams {
+                max_rws: 2000,
+                ..Default::default()
+            },
+        )
+        .new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
@@ -768,7 +794,8 @@ mod tests {
         };
         let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
         let block: GethData = test_ctx.into();
-        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        let mut builder =
+            BlockData::new_from_geth_data_default(block.clone()).new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
--- a/zkevm-circuits/src/evm_circuit.rs
+++ b/zkevm-circuits/src/evm_circuit.rs
@@ -337,8 +337,9 @@ pub mod test {
     pub fn run_test_circuit_geth_data_default<F: Field>(
         block: GethData,
     ) -> Result<(), Vec<VerifyFailure>> {
-        let mut builder = BlockData::new_from_geth_data(block.clone(), CircuitsParams::default())
-            .new_circuit_input_builder();
+        let mut builder =
+            BlockData::new_from_geth_data_with_params(block.clone(), CircuitsParams::default())
+                .new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
             .unwrap();
@@ -350,7 +351,7 @@ pub mod test {
         block: GethData,
         circuits_params: CircuitsParams,
     ) -> Result<(), Vec<VerifyFailure>> {
-        let mut builder = BlockData::new_from_geth_data(block.clone(), circuits_params)
+        let mut builder = BlockData::new_from_geth_data_with_params(block.clone(), circuits_params)
             .new_circuit_input_builder();
         builder
             .handle_block(&block.eth_block, &block.geth_traces)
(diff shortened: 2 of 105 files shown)


han0110 (commented):
First pass review, overall looks great, also nice refactor on `ConstraintBuilder` 🎉

zkevm-circuits/src/table.rs
```diff
@@ -82,6 +82,8 @@ pub enum TxFieldTag {
     TxSignHash,
     /// CallData
     CallData,
+    /// Pad
+    Pad,
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/2046b53d68d7923e708f1bf19acae7b1dfcf9fab

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -335,9 +339,17 @@ impl<F: Field> ExecutionConfig<F> {
...
                 cb.require_equal("q_step == 1", q_step.clone(), 1.expr());
             });
+            // rw_counter starts at 1
+            cb.condition(q_step_first, |cb| {
+                cb.require_equal(
+                    "rw_counter is initialized to be 1",
+                    step_curr.state.rw_counter.expr(),
+                    1.expr(),
+                )
+            });
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/2046b53d68d7923e708f1bf19acae7b1dfcf9fab

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -581,13 +594,21 @@ impl<F: Field> ExecutionConfig<F> {
...
 
         // Enforce the logic for this opcode
-        let q_steps: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
+        let sel_step: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
             &|meta| meta.query_advice(q_step, Rotation::cur());
-        let q_steps_first: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
+        let sel_step_first: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
             &|meta| meta.query_selector(q_step_first);
+        let sel_step_last: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
+            &|meta| meta.query_advice(q_step, Rotation::cur()) * meta.query_selector(q_step_last);
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/2046b53d68d7923e708f1bf19acae7b1dfcf9fab

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -689,14 +697,22 @@ pub mod dev {
...
+        mock::BlockData,
+    };
     use eth_types::{bytecode, geth_types::GethData, Word};
     use mock::test_ctx::helpers::account_0_code_account_1_no_code;
     use mock::TestContext;
 
     use crate::evm_circuit::test::rand_bytes;
     use crate::evm_circuit::witness::block_convert;
 
+    const CIRCUITS_PARAMS: CircuitsParams = CircuitsParams {
```
ChihChengLiang, line 711: We are use `CircuitsParams::default()` and constants CIRCUITS_PARAMS everywhere. What do you think we define some constants like CIRCUITS_PARAMS_S(256 rws), CIRCUITS_PARAMS_M(512 rws), CIRCUITS_PARAMS_L(2048 rws), CIRCUITS_PARAMS_XL(8192 rws) in a file? The benefit would be 

1. less ad hoc magic number scattering around.
2. Easy to search and track what testing cases are using large params.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -112,20 +112,20 @@ impl<F: Field> EvmCircuit<F> {
...
+    // #[cfg(any(feature = "test", test))]
+    // /// Assign exact steps in block without padding for unit test purpose
+    // pub fn assign_block_exact(
+    //     &self,
+    //     layouter: &mut impl Layouter<F>,
+    //     block: &Block<F>,
+    // ) -> Result<(), Error> {
+    //     self.execution.assign_block(layouter, block, true)?;
+    //     Ok(())
+    // }
```
ChihChengLiang, line 128: We should avoid checking in the commented-off code.
```suggestion
```

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -320,24 +325,36 @@ impl<F: Field> ExecutionConfig<F> {
...
-            cb.condition(q_step_first, |cb| {
+            cb.condition(q_step_first.clone(), |cb| {
+                cb.require_equal("q_step == 1", q_step.clone(), 1.expr());
+            });
+            // q_step needs to be enabled on the last row
+            cb.condition(q_step_last, |cb| {
                 cb.require_equal("q_step == 1", q_step.clone(), 1.expr());
             });
+            // rw_counter starts at 1
+            cb.condition(q_step_first, |cb| {
```
ChihChengLiang, line 351: Can we merge conditions here with line 343?

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -581,13 +599,21 @@ impl<F: Field> ExecutionConfig<F> {
...
+        let sel_step: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
             &|meta| meta.query_advice(q_step, Rotation::cur());
-        let q_steps_first: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
+        let sel_step_first: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
             &|meta| meta.query_selector(q_step_first);
+        let sel_step_last: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> =
+            &|meta| meta.query_selector(q_step_last);
+        let sel_not_step_last: &dyn Fn(&mut VirtualCells<F>) -> Expression<F> = &|meta| {
+            meta.query_advice(q_step, Rotation::cur())
+                * (1.expr() - meta.query_selector(q_step_last))
```
ChihChengLiang, line 610:
```suggestion
                * not::expr(meta.query_selector(q_step_last))
```

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -723,33 +748,52 @@ impl<F: Field> ExecutionConfig<F> {
...
-                    rw_counter: if block.txs.is_empty() {
-                        0
-                    } else {
-                        // if it is the first tx,  less 1 rw lookup, refer to end_tx gadget
-                        last_tx.steps.last().unwrap().rw_counter + 9 - (last_tx.id == 1) as usize
-                    },
-                    execution_state: ExecutionState::EndBlock,
-                    ..Default::default()
-                };
+                let dummy_tx = Transaction::default();
```
ChihChengLiang, line 751: We defined a dummy() for Transaction above but we use Transaction::default here, is this intended?

ChihChengLiang (commented):
First pass

zkevm-circuits/src/evm_circuit.rs
ed255: Thanks for mentioning this! I forgot to remove it.

zkevm-circuits/src/evm_circuit/execution.rs
ed255: Ah, that `Transaction::dummy()` is for the `circuit_input_builder::Transaction`, whereas this one is `witness::Transaction`.

I created a `dummy` function for `circuit_input_builder::Transaction` to show the intention that it's a dummy tx that contains dummy data, instead of using a default which to me seems like it could create a valid default tx.

zkevm-circuits/src/evm_circuit/execution.rs
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/ffbeaf6e168e519926d18d93ae4c0db5a9540161

zkevm-circuits/src/evm_circuit/execution.rs
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/ffbeaf6e168e519926d18d93ae4c0db5a9540161

han0110 (approved):
LGTM!

zkevm-circuits/src/evm_circuit/execution.rs
ed255: Note: we have a pending task which is trying to unify the types between `circuit_input_builder` and `witness` modules (which we currently have duplicated).  See https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/528

zkevm-circuits/src/evm_circuit/execution/end_block.rs
```diff
@@ -1,45 +1,143 @@
...
+                // Since every tx lookup done in the EVM circuit must succeed
+                // and uses a unique tx_id, we know that at
+                // least there are total_tx meaningful txs in
+                // the tx_table. We conclude that the number of
+                // meaningful txs in the tx_table is total_tx.
+            });
 
-        // TODO: For the rest steps, propagate the rw_counter and call_id to next step.
+            // 1. Verify rw_counter counts to the same number of meaningful rows in
+            // rw_table to ensure there is no malicious insertion.
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/d8ae8675d5ca3867fdc44367e49991455dde14a1

zkevm-circuits/src/witness/block.rs
```diff
@@ -25,10 +30,16 @@ pub struct Block<F> {
...
+    // TODO: Rename to `max_evm_rows`
     /// Pad evm circuit to make selectors fixed, so vk/pk can be universal.
+    /// When 0, the EVM circuit contains as many rows for all steps + 1 row
+    /// for EndBlock.
     pub evm_circuit_pad_to: usize,
-    /// Length to rw table rows in state circuit
-    pub state_circuit_pad_to: usize,
+    // /// Length to rw table rows in state circuit.  When 0, the State circuit
+    // /// contains as many rows as rw operations + 1 row for a Start.
+    // pub state_circuit_pad_to: usize,
```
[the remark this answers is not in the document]
ed255: Updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/d8ae8675d5ca3867fdc44367e49991455dde14a1

zkevm-circuits/src/copy_circuit.rs
ed255: Note: In https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/795/commits/bc2d34c9b51d9254c7e1f5c70c6b4cd87e0d04e6 I've cleaned up some of these `CIRCUITS_PARAMS` because one was not used, and the other once was only used once and didn't need to be `const`.

About your suggestion: Initially I thought about doing something similar to what you mention, but then I found that there are many configurations of particular cases that would only be used once or twice, so I just defined a default that works for 90% of the cases, and for the rest I thought declaring the parameters in the test itself is not that bad.  For example, notice that one test requires `max_txs = 4`, so we would need 5 different constants for parameters sets.

It's true that this way it would be easy to track tests using large params, but I think currently we could just find those that are not using the default params and they would be the large ones.

lispc:
lgtm.  One topic to discuss: 

in this PR new_from_geth_data is renamed so there are over 100 files changed.

I am a bit concern that there are currently more than 20 PRs pending. This kind of renaming will likely bring them more possibility of conflict that needs some time to resolve.  Is it ok to use other solutions, eg add a method `new_from_geth_data_with_config`, and keep `new_from_geth_data` as is?

The renaming can be done months later when most developments are done and we are cleaning tech debt?

lispc (approved):

ChihChengLiang (approved):
LGTM. Added a nitpick.

zkevm-circuits/src/evm_circuit/execution/gasprice.rs
```diff
@@ -122,6 +122,10 @@ mod test {
         )
         .unwrap();
 
-        assert_eq!(run_test_circuits(ctx, None), Ok(()));
+        let res = run_test_circuits_default(ctx, None);
+        if let Err(ref e) = res {
+            println!("{:#?}", e);
+        }
```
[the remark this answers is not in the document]
ed255: oops! Thanks for spotting that! Deleted.

lispc:
can we merge this now?

lispc:
is the default max_rw 256 too small? the integration tests failed, and this returndatasize https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3343240930/jobs/5536258158 also failed.

ed255:
> is the default max_rw 256 too small? the integration tests failed, and this returndatasize https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3343240930/jobs/5536258158 also failed.
> 
> I think only the percentage of unit tests which needs to set max_rw manually should better not be more than 10%.

Ah sorry for causing these issues :sweat_smile: Increasing the default `max_rws` sounds good to me!  I can do it later today.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -798,26 +839,19 @@ impl<F: Field> ExecutionConfig<F> {
...
-                                    log::warn!(
-                                        "evm circuit offset larger than padding: {} > {}",
-                                        offset,
-                                        block.evm_circuit_pad_to
-                                    );
-                                }
-                                break;
-                            }
-                        }
+                    if !exact && offset >= evm_rows {
```
pinkiebell, line 843: @ed255 Off by one or intended?

ed255: ah this is tricky, but I think this is not a bug.
If there are 10 rows, the offset can only go from 0 to 9, so offset=10 should trigger the error.
The log message is wrong though, it should be `evm circuit offset larger than padding: {} >= {}`

pinkiebell: The offset calculation is a bit brittle and I attempt to fix it via #870 . I encountered a few more issues depending on circuit configuration and the number of transactions in geth data etc

ed255: Actually I checked it again and it was an off by one bug T.T  This is because at this line, the offset contains the next available offset.  It's OK for the next available offset to be equal to evm_rows, as long as we finish the loop here.  So it should be `offset > evm_rows` in the check.

I saw your PR!  Thanks for the effort on simplifying the offset handling logic!  I still think there are some things that need fixing that I completely missed on my EndBlock PR.

Issue #790, opened 2022-09-21
Prover test failure


ChihChengLiang:
- zkevm-circuit commit hash >> acf2c0b967ee3c073ce9195ef165d966666cecfa
- zkevm-chain commit hash  >> 68982510b93fd34e06f5d2f444a3807b90668b40

Reproducible examples: https://github.com/pinkiebell/tmp

```
Using circuit parameters: BLOCK_GAS_LIMIT=63000 MAX_TXS=3 MAX_CALLDATA=10500 MAX_BYTECODE=14000 MIN_K=19 STATE_CIRCUIT_PAD_TO=524032
```

glance of circuit bug
```
cat PROVER_LOG.txt| grep name | sort --unique
                        name: "",
                    name: "Execution step",
                    name: "assign bytecode",
                    name: "assign keccak rounds",
                    name: "ecc chip aux",
                    name: "ecdsa chip verification",
                    name: "length and data rlc",
                    name: "rw table",
                    name: "signature address verify",
                    name: "squeeze",
                name: "hash rlc check",
                name: "intermediate data rlc",
                name: "update data rlc",
            name: "Keccak",
            name: "composition_a",
            name: "composition_b",
            name: "composition_c",
            name: "composition_d",
            name: "keccak",
            name: "mpt_update exists in mpt circuit for Account last access",
            name: "overflow_a",
```

ChihChengLiang:
@pinkiebell @AronisAt79 can you update the zkevm-chain commit hash and zkevm-circuit commit hash on the description?

AronisAt79:
> @pinkiebell @AronisAt79 can you update the zkevm-chain commit hash and zkevm-circuit commit hash on the description?

done inline

pinkiebell:
Another note, a proving an empty block also has resulted in a `ConstraintSystemFailure`, that was with the real prover

ChihChengLiang:
> Another note, a proving an empty block also has resulted in a ConstraintSystemFailure, that was with the real prover

@ed255 do we need endblock to prove an empty block?

pinkiebell:
This time I have some debug information:
```
error: lookup input does not exist in table
  (L0, L1) ∉ (F38, F39)

  Lookup 'composition_a' inputs:
    L0 = x0
    ^
    | Cell layout in region 'ecdsa chip verification':
    |   | Offset | F40|
    |   +--------+----+
[...]

PR #797, opened 2022-09-23
Add missing ECDSA RangeChip load in SuperCircuit

pinkiebell:
Related: #790


pinkiebell:
The prover is still running but looks like it's past the hill. Thanks @han0110 ! You might to review it too :)

han0110 (approved):
LGTM!

Issue #747, opened 2022-09-08
SuperCircuit: Integration of Tables and Subcircuits


andyguzmaneth:
Currently the SuperCircuit contains the State and Copy circuits, but it seems that the rw_table and copy_table are duplicate.  The tables should be de-duplicated and reused between the EVM and State Circuit (for rw_table), and EVM and Copy Circuit (for copy_table)

aguzmant103:
Likely need to split this task into further ones.

PR #800, opened 2022-09-25
Unify `CopyTable` load and `CopyCircuit` assign.

davidnevadoc:
Modify `CopyTable` and `CopyCircuit` so they both use the same assign logic.
The columns `is_last`, `is_pad`, `is_code`, `value`  and the selector `q_step` have been moved from the `CopyCircuit` config to the `CopyTable` itself.
Part of solving issue #747


ed255 (approved):
LGTM! Thanks for working on this!

PR #803, opened 2022-09-27
Refactor Tx circuit and use challenge api

### LINKED ISSUE #752: Refactor Tx circuit with challenge API
Using the challenge API instead of the instance value for randomness in Tx circuit

Notes:

- The randomness will be provided from the super circuit instead of being queried from within the Keccak circuit. This applies to all the sub circuits
- Treat the randomness as an expression, need a small test.
- The circuit needs to be divided into 2 phases: one with the commitments that don't use the randomness; and another one that has access to the randomness and commits values that use it (via RLC).

han0110:
This PR aims to refactor Tx circuit with challenge API.

It reuses `MainGate`'s advice columns to create an extra random linear combination gate to get rid of 130 columns in `SignVerifyChip`, with the cost of few more rows to do `SignVerifyChip::assign_signature_verify` than previous approach. So now tests and benchmark of Tx circuit can be ran locally.

In respect to more rows (76 rows precisely) to verify a single signature, I think it should be negligible cause ecdsa verification already costs ~70k, and is still the bottleneck. Also even we have reached our bottleneck on ecdsa verification, we can actually just allocate another `MainGate` to double our capacity (each `MainGate` with `RangeChip` costs ~60 msm) without too much change to current code.

There is also a caveat about multi-phase: The `AssignedCell` will have `value: Value::unknown` in any phase that is not equal to current one, because we need to follow the principle that the function `to` of `assign_{fixed,advice}` will only be called once (it's a `FnMut` so we need to guarantee that). So it's a little bit awkward to calculate later phase witness with previous phase's `AssignedCell`, we need to pass the `Cell` and `Value` separately to make sure we can do the calculation and at the same time enforce the copy constraint.

And currently this behavior is not shown in `MockProver`, so it takes me some time to find this issue, I have also created an issue https://github.com/privacy-scaling-explorations/halo2/issues/102 to propose to make `MockProver` to do multi-phsae synthesis just like real prover.

## Todo

- [ ] Update spec https://github.com/privacy-scaling-explorations/zkevm-specs/pull/292

### DIFF
--- a/zkevm-circuits/src/tx_circuit/sign_verify.rs
+++ b/zkevm-circuits/src/tx_circuit/sign_verify.rs
@@ -5,21 +5,20 @@
 // - *_le: Little-Endian bytes
 
 use crate::{
-    evm_circuit::util::{not, RandomLinearCombination, Word},
+    evm_circuit::util::{not, rlc},
     table::KeccakTable,
-    util::Expr,
+    util::{Challenges, Expr},
 };
-use ecc::{EccConfig, GeneralEccChip};
+use ecc::{maingate, EccConfig, GeneralEccChip};
 use ecdsa::ecdsa::{AssignedEcdsaSig, AssignedPublicKey, EcdsaChip};
 use eth_types::sign_types::{pk_bytes_le, pk_bytes_swap_endianness, SignData};
 use eth_types::{self, Field};
-use gadgets::is_zero::{IsZeroChip, IsZeroConfig, IsZeroInstruction};
-use halo2_proofs::halo2curves::secp256k1::Secp256k1Affine;
 use halo2_proofs::{
     arithmetic::FieldExt,
-    circuit::{AssignedCell, Layouter, Region, Value},
+    circuit::{AssignedCell, Cell, Layouter, Value},
     halo2curves::secp256k1,
-    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Selector},
+    halo2curves::secp256k1::Secp256k1Affine,
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression, SecondPhase, Selector},
     poly::Rotation,
 };
 use integer::{AssignedInteger, IntegerChip, IntegerConfig, IntegerInstructions, Range};
@@ -31,14 +30,8 @@ use maingate::{
     AssignedValue, MainGate, MainGateConfig, MainGateInstructions, RangeChip, RangeConfig,
     RangeInstructions, RegionCtx,
 };
-use std::marker::PhantomData;
-
-/// Power of randomness vector size required for the SignVerifyChip
-pub const POW_RAND_SIZE: usize = 63;
-
-/// Number of rows required for a verification of the SignVerifyChip in the
-/// "signature address verify" region.
-pub const VERIF_HEIGHT: usize = 1;
+use num::Integer;
+use std::{iter, marker::PhantomData};
 
 /// Auxiliary Gadget to verify a that a message hash is signed by the public
 /// key corresponding to an Ethereum Address.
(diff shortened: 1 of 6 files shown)


zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -54,208 +47,148 @@ pub struct SignVerifyChip<F: Field, const MAX_VERIF: usize> {
...
+        meta.create_gate(name, |meta| {
+            let q_rlc = meta.query_selector(q_rlc);
+            let [a, b, c, d, e] = main_gate_config
+                .advices()
+                .map(|column| meta.query_advice(column, Rotation::cur()));
+            let [rlc, rlc_next] = [Rotation::cur(), Rotation::next()]
+                .map(|rotation| meta.query_advice(rlc, rotation));
+            let inputs = [e, d, c, b, a, rlc];
+            let powers_of_challenge = iter::successors(challenge.clone().into(), |power| {
+                (challenge.clone() * power.clone()).into()
```
ed255, line 177: In this expression we end up with `r^5`.  Does this affect the maximum expression degrees in the constraints?  Or since it comes from a challenge expression it has no extra cost that could affect the extended degree?

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -400,141 +348,213 @@ impl<F: Field, const MAX_VERIF: usize> SignVerifyChip<F, MAX_VERIF> {
...
+
+        let mut rlc = challenge.map(|_| F::zero());
+        for (chunk_idx, chunk) in inputs_be.chunks_exact(columns.len()).enumerate() {
+            ctx.enable(q_rlc)?;
+            let assigned_rlc = ctx.assign_advice(|| "{name}_rlc[{chunk_idx}]", config.rlc, rlc)?;
+            for ((idx, column), term) in (chunk_idx * chunk.len()..).zip(columns).zip(chunk) {
+                let copied =
+                    ctx.assign_advice(|| format!("{name}_byte[{idx}]"), column, term.value())?;
+                if let Some(cell) = term.cell() {
+                    ctx.constrain_equal(cell, copied.cell())?;
```
ed255, line 378: Let me know if I understand this correctly:
Whenever the `inputs_le` is not multiple of 5 (the number of columns in a main gate row), there are some front-padding 0s.  With this you are doing a copy constraint of those main gate cells, to a constant-zero cell right?  And afterwards the copy constraint is applied to the bytes.

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -400,141 +348,213 @@ impl<F: Field, const MAX_VERIF: usize> SignVerifyChip<F, MAX_VERIF> {
...
+        &self,
+        config: &SignVerifyConfig,
+        ctx: &mut RegionCtx<F>,
+        chips: &ChipsRef<F, NUMBER_OF_LIMBS, BIT_LEN_LIMB>,
+        name: &str,
+        q_rlc: Selector,
+        challenge: Value<F>,
+        inputs_le: impl IntoIterator<Item = Term<F>>,
+    ) -> Result<AssignedCell<F, F>, Error> {
+        let zero = chips.main_gate.assign_constant(ctx, F::zero())?;
```
ed255, line 362: If instead of creating a new constant for every call to `assign_rlc_le` we do it once for the entire circuit, do you think we could have a non-negligible optimization benefit?

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -54,208 +47,148 @@ pub struct SignVerifyChip<F: Field, const MAX_VERIF: usize> {
...
+            let [rlc, rlc_next] = [Rotation::cur(), Rotation::next()]
+                .map(|rotation| meta.query_advice(rlc, rotation));
+            let inputs = [e, d, c, b, a, rlc];
+            let powers_of_challenge = iter::successors(challenge.clone().into(), |power| {
+                (challenge.clone() * power.clone()).into()
+            })
+            .take(inputs.len() - 1)
+            .collect_vec();
+
+            vec![q_rlc * (rlc_next - rlc::expr(&inputs, &powers_of_challenge))]
```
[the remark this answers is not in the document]
ed255: I'm replying to myself: I saw that you're doing a copy constraint to the zero-constant cell in the assignment, so this is already covered :)

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -400,141 +348,213 @@ impl<F: Field, const MAX_VERIF: usize> SignVerifyChip<F, MAX_VERIF> {
...
+        let columns = config.main_gate_config.advices();
+        let inputs_le = inputs_le.into_iter().collect_vec();
+        let inputs_be = iter::repeat_with(|| Term::assigned(zero.cell(), Value::known(F::zero())))
+            .take(Integer::next_multiple_of(&inputs_le.len(), &columns.len()) - inputs_le.len())
+            .chain(inputs_le.into_iter().rev())
+            .collect_vec();
+
+        let mut rlc = challenge.map(|_| F::zero());
+        for (chunk_idx, chunk) in inputs_be.chunks_exact(columns.len()).enumerate() {
+            ctx.enable(q_rlc)?;
```
ed255, line 372: I wonder if `q_rlc` should be disabled in the last row :thinking:

Imagine you call `assign_rlc_le` consecutively.  You'd get a column `q_rlc` at 1 consecutively in all rows, but you have different RLCs, so they would mix together.  I also wonder why this is not happening; maybe `chips.main_gate.assign_constant(ctx, F::zero())?;` is taking a row, which would have `q_rlc = 0`?

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -400,141 +348,213 @@ impl<F: Field, const MAX_VERIF: usize> SignVerifyChip<F, MAX_VERIF> {
...
-            offset,
-            || Value::known(address),
+        let pk_hash_rlc = self.assign_rlc_le(
+            config,
+            ctx,
+            chips,
+            "pk_hash_rlc",
+            config.q_rlc_evm_word,
+            challenges.evm_word(),
+            iter::empty().chain(pk_hash_lo.into_iter().rev()).chain(
```
ed255, line 536: Just to be sure: here we have all the terms in `pk_hash_lo` are `Term::assigned` and all the elements in `pk_hash_hi` `Term::unassigned` right?

ed255 (approved):
LGTM!  Amazing work!  Not only you updated the circuit to use the challenge API, but also optimized it removing many rows, and also cleaned up some stuff :D 

I've left some questions (so that I make sure i understood everything) and a suggestion, please take a look :)

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: > since it comes from a challenge expression it has no extra cost that could affect the extended degree?

Yes, it's just like a constant.

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: Exactly, since the `width` of `MainGate` not always divides length of `inputs_le`, so we need to add zero padding in front of `input_be` by copy constraint.

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: > If instead of creating a new constant for every call to `assign_rlc_le` we do it once for the entire circuit

I agree so, but I feel it's ugly to implement in Tx circuit, it should be an API `constrain_constant` provided from `MainGate` just like what `Region` provides, and inside `MainGate` it tracks assigned constant cells and reuse to avoid such unnecessary cost.

I have created an feature request issue here https://github.com/privacy-scaling-explorations/halo2wrong/issues/47

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: Yes, the layout with example above is actually wrong! It should be:

```diff
        // Layout (take input with length 12 as an example)
        // | q_rlc |                          rlc                        |   a   |   b   |   c   |   d    |   e    |
        // | ----- | --------------------------------------------------- | ----- | ----- | ----- | ------ | ------ |
        // |   1   |                                                   0 |     0 |     0 |     0 |  be[0] |  be[1] |
        // |   1   |                                  be[0]*r^1 +  be[1] | be[2] | be[3] | be[4] |  be[5] |  be[6] |
        // |   1   | be[0]*r^6  + be[1]*r^5  + ... +  be[5]*r^1 +  be[6] | be[7] | be[8] | be[9] | be[10] | be[11] |
-       // |   1   | be[0]*r^11 + be[1]*r^10 + ... + be[10]*r^1 + be[11] |       |       |       |        |        |
+       // |   0   | be[0]*r^11 + be[1]*r^10 + ... + be[10]*r^1 + be[11] |       |       |       |        |        |
```

And in line 392-393 is where we actually assign the last row, without enabling the `q_rlc` just like you said. Will fix the layout document soon.

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: Yes, so `pk_hash_hi` are only witnessed for building the random linear combination of keccak input, but `pk_hash_lo` should be copied from `address` we have assigned ahead in order to get the value `is_address_zero`.

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -54,208 +47,148 @@ pub struct SignVerifyChip<F: Field, const MAX_VERIF: usize> {
...
+            "keccak_input_rlc",
+            main_gate_config.clone(),
+            q_rlc_keccak_input,
+            rlc,
+            challenges.keccak_input(),
         );
 
+        // Ref. spec SignVerifyChip 1. Verify that keccak(pub_key_bytes) = pub_key_hash
+        // by keccak table lookup, where pub_key_bytes is built from the pub_key
+        // in the ecdsa_chi
```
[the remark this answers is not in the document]
han0110: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/803/commits/885034bd501f5c440fe60aa026bd8389c67b1170.

zkevm-circuits/src/tx_circuit/sign_verify.rs
ed255: Handling the reuse of constants from the `MainGate` sounds great!

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -400,141 +350,207 @@ impl<F: Field, const MAX_VERIF: usize> SignVerifyChip<F, MAX_VERIF> {
...
+    ) -> Result<AssignedCell<F, F>, Error> {
+        let zero = chips.main_gate.assign_constant(ctx, F::zero())?;
+        let columns = config.main_gate_config.advices();
+        let inputs_le = inputs_le.into_iter().collect_vec();
+        let inputs_be = iter::repeat_with(|| Term::assigned(zero.cell(), Value::known(F::zero())))
+            .take(Integer::next_multiple_of(&inputs_le.len(), &columns.len()) - inputs_le.len())
+            .chain(inputs_le.into_iter().rev())
+            .collect_vec();
+
+        let mut rlc = challenge.map(|_| F::zero());
```
ChihChengLiang, line 372: What are we trying to get from the expression `challenge.map(|_| F::zero())`? If it is `Value::known(F::zero())`, why dont we do so directly?

ChihChengLiang (approved):
LGTM. I did only a light review so that this pr can be merged to unblock other prs. Added a comment.

zkevm-circuits/src/tx_circuit/sign_verify.rs
```diff
@@ -269,6 +204,36 @@ impl<F: Field> SignVerifyConfig<F> {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum Term<F> {
```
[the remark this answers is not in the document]
han0110: Make sense, added in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/803/commits/7f123a96ccda192a89f6775cf9cde4f650b65220.

zkevm-circuits/src/tx_circuit/sign_verify.rs
han0110: Yeah using `Value::known(F::zero())` directly seems more straightforward, updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/803/commits/7f123a96ccda192a89f6775cf9cde4f650b65220.

han0110:
Although the spec has not been merged, but since https://github.com/privacy-scaling-explorations/zkevm-specs/pull/292 only adds some more description to Tx circuit, so I think we can merge this first to unblock #816.

PR #817, opened 2022-10-03
geth-utils: remove 10mb trace limit

pinkiebell:
I hit this one while testing the circuits

### DIFF
--- a/external-tracer/src/lib.rs
+++ b/external-tracer/src/lib.rs
@@ -45,7 +45,7 @@ pub struct LoggerConfig {
 impl Default for LoggerConfig {
     fn default() -> Self {
         Self {
-            enable_memory: true,
+            enable_memory: false,
             disable_stack: false,
             disable_storage: false,
             enable_return_data: true,
--- a/geth-utils/lib/lib.go
+++ b/geth-utils/lib/lib.go
@@ -27,10 +27,6 @@ func CreateTrace(configStr *C.char) *C.char {
 	}
 
 	bytes, err := json.MarshalIndent(executionResults, "", "  ")
-	if len(bytes) > 10000000 {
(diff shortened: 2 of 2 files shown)


han0110:
@adria0 I saw this limitation is added in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/748, is there anything broken when the trace is too large? Otherwise I think it should be fine to process large trace.

ed255:
I think that after this was merged some months ago https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/614 we shouldn't encounter massive memory usage in the traces (as I believe this was coming from having a memory snapshot at each step of the trace).  We just need to make sure that we're always querying the trace with `EnableMemory: false`.  I think this may not be the case when using the internal tracer.  See: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2235254416aefd1bc9b5592e4e06cd58f8516e2a/external-tracer/src/lib.rs#L48

ed255 (approved):
LGTM!  I think that without memory snapshots this check is unnecessary.  (We may still need to update some places to avoid requesting the memory snapshots from the trace)

pinkiebell:
> I think that after this was merged some months ago #614 we shouldn't encounter massive memory usage in the traces (as I believe this was coming from having a memory snapshot at each step of the trace). We just need to make sure that we're always querying the trace with `EnableMemory: false`. I think this may not be the case when using the internal tracer. See:
> 
> https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2235254416aefd1bc9b5592e4e06cd58f8516e2a/external-tracer/src/lib.rs#L48
> 
> @pinkiebell could you try setting the default `enable_memory` to `false` and see if all unit tests still pass?

Looks good 👌

han0110 (approved):
LGTM!

PR #818, opened 2022-10-04
Increase rw_counter_offset when doing a copy table lookup

z2trillion:
Doing this allows us to do additional rw table lookups after copy table lookups.


ed255 (approved):
Nice!  LGTM!

PR #786, opened 2022-09-20
zkevm-circuits: move `mock` dependency to the `test` target

pinkiebell:

PR #820, opened 2022-10-05
fix: add `mock` dependency to zkevm-circuit `dev-dependencies`

silathdiir:
It seems that test build is failed when running `cargo test` (without `--all-features` or `--feature test`) in `zkevm-circuit` folder.
```
error[E0433]: failed to resolve: use of undeclared crate or module `mock`
   --> zkevm-circuits/src/copy_circuit.rs:694:9
    |
694 |     use mock::test_ctx::helpers::account_0_code_account_1_no_code;
    |         ^^^^ use of undeclared crate or module `mock`
```
`mock` dependency was set to optional in PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/786.
Fixed to add `mock` to `dev-dependencies` in `Cargo.toml` for tests.

### DIFF
--- a/zkevm-circuits/Cargo.toml
+++ b/zkevm-circuits/Cargo.toml
@@ -16,7 +16,7 @@ bus-mapping = { path = "../bus-mapping" }
 eth-types = { path = "../eth-types" }
 gadgets = { path = "../gadgets" }
 ethers-core = "0.17.0"
-ethers-signers = "0.17.0"
+ethers-signers = { version = "0.17.0", optional = true }
 mock = { path = "../mock", optional = true }
 strum = "0.24"
 strum_macros = "0.24"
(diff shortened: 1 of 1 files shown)


pinkiebell (approved):
LGTM. After looking at the deps, I saw that `ethers-signers` could also be made optional like the `mock` depdendency and added to `test = [...]`. Would be nice nice if you include that change too 👍

silathdiir:
> LGTM. After looking at the deps, I saw that `ethers-signers` could also be made optional like the `mock` depdendency and added to `test = [...]`. Would be nice nice if you include that change too 👍

Fixed to update `ethers-signers` to optional (in `dependencies` section) and associate it with feature `test`.

pinkiebell (approved):
LGTM

pinkiebell:
@silathdiir Feel free to merge 👌

PR #827, opened 2022-10-08
implement OOG Call state

### LINKED ISSUE #911: implement OOG Call error state circuit
PR #827 targets it.

DreamWuGit:
spec PR :https://github.com/privacy-scaling-explorations/zkevm-specs/pull/287


aguzmant103:
Hey @DreamWuGit I'm reviewing open PRs, for this one the next step is rebasing and getting a reviewer from PSE/another team?

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,522 @@
...
+            cb.curr.state.is_root.expr(),
+            is_to_end_tx,
+        );
+
+        // When it's a root call
+        cb.condition(cb.curr.state.is_root.expr(), |cb| {
+            // Do step state transition
+            cb.require_step_state_transition(StepStateTransition {
+                call_id: Same,
+                rw_counter: Delta(18.expr() + cb.curr.state.reversible_write_counter.expr()),
```
ChihChengLiang, line 176: This part is different from the spec. https://github.com/privacy-scaling-explorations/zkevm-specs/blob/0bfb00a41cc45564fe1c111c926cb8b028cfb053/src/zkevm_specs/evm/execution/oog_call.py#L94

ChihChengLiang:
@aguzmant103 I'm currently reviewing this one.
@adria0 would be great to have your review on this one, too, since you reviewed the spec before.

adria0 (commented):
I see that there's a lot of common code with `callop`.
I think that worth the time to try to factorize both things to avoid code duplication.

This also will simplify updates that needs to be applied in the code like Challenge intgration (#898)

@ChihChengLiang, @DreamWuGit wdyt?

CPerezz:
Sorry, I saw the review request now. Thanks @ChihChengLiang for unassigning me. Is it still needed that I take a look?

ChihChengLiang:
> I see that there's a lot of common code with `callop`. I think that worth the time to try to factorize both things to avoid code duplication.

@adria0 Can we do that in a future PR? It seems to require some work to achieve that in this one.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -789,9 +794,11 @@ impl<'a> CircuitInputStateRef<'a> {
                     false,
                     op,
                 );
-                self.tx.steps_mut()[step_index]
-                    .bus_mapping_instance
-                    .push(rev_op_ref);
+                if self.tx.steps().len() > step_index {
```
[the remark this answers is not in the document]
DreamWuGit: good question:  it is because in the buss mapping use `state.push_op_reversible` for tx access list account warm checking..  in this method if current call is failed, it will add reversion group of current `Op` (refer to https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L185) , for example, for oog call cases, geth steps in tx can be "`push 100 push 0xfefe ... push Call(OOG_Call step) " ` the last step will add into reversion of key & value pair `(tx.steps.len(), Op)`.  hence when loop reversion group, `self.tx.steps()[tx.steps.len()]` will corrupt . 
I have tried if not use `push_op_reversible` and take use of read operations without push to reversion group, don't need this `if` judgement any more. 
Both seems OK to me , what do you think ? @adria0

CPerezz:
> > 
> 
> Sure, @ChihChengLiang, let's do it in a future PR

Please @adria0 make sure an issue to track this future work is created and linked/mentions this one

adria0 (approved):
Overall, LGTM

bus-mapping/src/circuit_input_builder/input_state_ref.rs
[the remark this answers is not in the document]
DreamWuGit: I will ping @ed255 as CC is on holiday this week, maybe not use `state.push_op_reversible` but only read is better to avoid this `if` statement.

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,522 @@
...
+        let is_warm_prev = cb.query_bool();
+        cb.account_access_list_write(
+            tx_id.expr(),
+            callee_address.clone(),
+            is_warm.expr(),
+            is_warm_prev.expr(),
+            Some(&mut reversion_info),
+        );
+
+        // Verify transfer
```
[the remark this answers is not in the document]
DreamWuGit: removed in [54a8a11](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/54a8a1130023a2def96a24b60581f161d4b04b67) , just borrowed from CallOp before. thanks

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,522 @@
...
+
+        // Lookup values from stack
+        cb.stack_pop(gas_word.expr());
+        cb.stack_pop(callee_address_word.expr());
+        cb.stack_pop(value.expr());
+        cb.stack_pop(cd_offset.expr());
+        cb.stack_pop(cd_length.expr());
+        cb.stack_pop(rd_offset.expr());
+        cb.stack_pop(rd_length.expr());
+        cb.stack_push(is_success.expr());
```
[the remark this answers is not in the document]
DreamWuGit: I like it updated in [54a8a11](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/54a8a1130023a2def96a24b60581f161d4b04b67) thanks

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,522 @@
...
+}
+
+impl<F: Field> ExecutionGadget<F> for ErrorOOGCallGadget<F> {
+    const NAME: &'static str = "ErrorOutOfGasCall";
+
+    const EXECUTION_STATE: ExecutionState = ExecutionState::ErrorOutOfGasCALL;
+
+    fn configure(cb: &mut ConstraintBuilder<F>) -> Self {
+        let opcode = cb.query_cell();
+        cb.opcode_lookup(opcode.expr(), 1.expr());
```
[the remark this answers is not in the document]
DreamWuGit: added in [54a8a11](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/54a8a1130023a2def96a24b60581f161d4b04b67) thanks !

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -773,6 +773,11 @@ impl<'a> CircuitInputStateRef<'a> {
 
     /// Handle a reversion group
     fn handle_reversion(&mut self) {
+        if self.tx_ctx.reversion_groups.is_empty() {
```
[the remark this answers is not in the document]
DreamWuGit: debug info , removed it now! thanks

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
DreamWuGit: it is related to discussion https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827#discussion_r1031341590 , will update after it finalizes.

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,511 @@
...
+            PUSH32(Word::from(stack.rd_offset))
+            PUSH32(Word::from(stack.cd_length))
+            PUSH32(Word::from(stack.cd_offset))
+            PUSH32(stack.value)
+            PUSH32(Address::repeat_byte(0xfe).to_word())
+            PUSH32(Word::from(stack.gas))
+            CALL // make this call out of gas
+            PUSH32(Word::from(0))
+            PUSH32(Word::from(0))
+        };
```
han0110, line 507: These are never used and unnecessary

han0110 (commented):
Overall looks good. Also I think the tests are worthy to be simplified (remove all unused logic should make review faster).

bus-mapping/src/circuit_input_builder/input_state_ref.rs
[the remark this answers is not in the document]
DreamWuGit: sure, will update it.

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,511 @@
...
+        );
+
+        let value_is_zero = IsZeroGadget::construct(cb, sum::expr(&value.cells));
+        let has_value = 1.expr() - value_is_zero.expr();
+        cb.condition(has_value.clone(), |cb| {
+            cb.require_zero(
+                "CALL with value must not be in static call stack",
+                is_static.expr(),
+            );
+        });
```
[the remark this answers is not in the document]
DreamWuGit: for `ErrorOutOfGasSTATICCALL` I am planing to merge it into `oog_call` in the future, at the moment we can remove it.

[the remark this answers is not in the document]
han0110: I mean even we consider out of gas for `STATICCALL`, it's still not necessary.

Out of gas and violation of static context are 2 different kinds of exceptions, and in go-ethereum implementation out of gas will always be detected before the later one, so this will become an over constraint if they happen at the same time.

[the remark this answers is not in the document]
DreamWuGit: Got it , that is true.

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
DreamWuGit: these are used for callee bytecode in L508

han0110: They will never be executed, since the call from `caller` is out of gas.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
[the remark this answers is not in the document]
DreamWuGit: updated to read  in [7ca4a2f](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/7ca4a2fa8ddc5f740be472d36309b5323495f0e9)

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,511 @@
...
+        let cd_length = cb.query_rlc();
+        let rd_offset = cb.query_cell();
+        let rd_length = cb.query_rlc();
+        let balance = cb.query_word();
+
+        let tx_id = cb.call_context(None, CallContextFieldTag::TxId);
+
+        let mut reversion_info = cb.reversion_info_read(None);
+        let [is_static] =
+            [CallContextFieldTag::IsStatic].map(|field_tag| cb.call_context(None, field_tag));
```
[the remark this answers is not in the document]
DreamWuGit: updated in [7ca4a2f](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/7ca4a2fa8ddc5f740be472d36309b5323495f0e9) thanks!

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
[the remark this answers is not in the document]
DreamWuGit: remove it in [7ca4a2f](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/7ca4a2fa8ddc5f740be472d36309b5323495f0e9) thanks !

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
DreamWuGit: seems not,  I tried remove them and set callee bytecode just `STOP`, this condition will not met oog gas state.

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
DreamWuGit: will update spec to match the real circuit .

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
han0110: I see, just checked the steps and saw it now. I'd still suggest to simply these test cases to make it as simpler as possible (e.g. make call of `caller` out of gas, so `callee` doesn't need to hold bytecode), that should make review much easier.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -194,6 +194,24 @@ impl<'a> CircuitInputStateRef<'a> {
...
+        let op_ref = self.block.container.insert(Operation::new_reversible(
+            self.block_ctx.rwc.inc_pre(),
+            RW::READ,
+            op,
+        ));
+        step.bus_mapping_instance.push(op_ref);
+
+        Ok(())
+    }
+
```
han0110, line 214: I think this doesn't make too much sense, read doesn't mutate state so will never need reversion.

bus-mapping/src/evm/opcodes/error_oog_call.rs
```diff
@@ -0,0 +1,89 @@
...
+        let is_warm = state.sdb.check_account_in_access_list(&call_address);
+        state.read_op_reversible(
+            &mut exec_step,
+            TxAccessListAccountOp {
+                tx_id,
+                address: call_address,
+                is_warm: true,
+                is_warm_prev: is_warm,
+            },
+        )?;
```
han0110, line 72:
```suggestion
        state.push_op(
            &mut exec_step,
            RW::READ,
            TxAccessListAccountOp {
                tx_id,
                address: call_address,
                is_warm,
                is_warm_prev: is_warm,
            },
        )?;
```

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
DreamWuGit: spec PR https://github.com/privacy-scaling-explorations/zkevm-specs/pull/332 will update to match circuit diff.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
DreamWuGit: just use `state.push_op(` is indeed more cheaper , will remove this new method.

DreamWuGit: removed in [58beba4](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/827/commits/58beba4d5c114a534f898e68e1ec2434c44f7453) thanks !!

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -779,6 +779,30 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
...
+                account_address,
+                0.expr(),
+                0.expr(),
+                value,
+                value_prev,
+                0.expr(),
+                0.expr(),
+            ),
+        );
+    }
```
han0110, line 804: Since it's a read, `value` and `value_prev` should always be the same.

```suggestion
    pub(crate) fn account_access_list_read(
        &mut self,
        tx_id: Expression<F>,
        account_address: Expression<F>,
        value: Expression<F>,
    ) {
        self.rw_lookup(
            "account access list read",
            false.expr(),
            RwTableTag::TxAccessListAccount,
            RwValues::new(
                tx_id,
                account_address,
                0.expr(),
                0.expr(),
                value.clone(),
                value,
                0.expr(),
                0.expr(),
            ),
        );
    }
```

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -0,0 +1,493 @@
...
+
+        // Add callee to access list
+        let is_warm = cb.query_bool();
+        let is_warm_prev = cb.query_bool();
+        cb.account_access_list_read(
+            tx_id.expr(),
+            callee_address.clone(),
+            is_warm.expr(),
+            is_warm_prev.expr(),
+        );
```
han0110, line 103: Then we can get rid of `is_warm_prev`

```suggestion
        let is_warm = cb.query_bool();
        cb.account_access_list_read(
            tx_id.expr(),
            callee_address.clone(),
            is_warm.expr(),
        );
```

han0110 (commented):
Added a final nitpick, after this I think we can merge. Thanks for considering all the suggestions.

han0110 (approved):

PR #837, opened 2022-10-18
Make stack_pointer_offset an expression so self.condition can be applied

z2trillion:


zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -988,12 +988,12 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
...
     pub(crate) fn stack_pop(&mut self, value: Expression<F>) {
-        self.stack_lookup(false.expr(), self.stack_pointer_offset.expr(), value);
-        self.stack_pointer_offset += 1;
+        self.stack_lookup(false.expr(), self.stack_pointer_offset.clone(), value);
+        self.stack_pointer_offset = self.stack_pointer_offset.clone() + self.condition_expr();
     }
 
     pub(crate) fn stack_push(&mut self, value: Expression<F>) {
-        self.stack_pointer_offset -= 1;
+        self.stack_pointer_offset = self.stack_pointer_offset.clone() - self.condition_expr();
```
DreamWuGit, line 996: I am wandering why previous look up succeed:  the stack_pointer_offset is always reduced by 1 no matter the condition is.

z2trillion: I think that is because it's only used when the condition is None.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -988,12 +988,12 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
     // Stack
 
     pub(crate) fn stack_pop(&mut self, value: Expression<F>) {
-        self.stack_lookup(false.expr(), self.stack_pointer_offset.expr(), value);
-        self.stack_pointer_offset += 1;
+        self.stack_lookup(false.expr(), self.stack_pointer_offset.clone(), value);
```
roynalnaruto, line 991: This change is unnecessary, as the `self.stack_pointer_offset.expr()` will perform the same operation.

It's up to you whether you wish to revert this. But I personally prefer `.expr()` instead of `.clone()` over here because `Clone` does not tell me what type it'll be, but with `Expr`  I know it will be an expression. Up to you :)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -988,12 +988,12 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
...
+        self.stack_pointer_offset = self.stack_pointer_offset.clone() + self.condition_expr();
```
roynalnaruto, line 992: Similarly here. `clone` v `expr`

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
roynalnaruto, line 996: also here

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1132,11 +1132,7 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
                 rwc_inc: rwc_inc.clone(),
             },
         );
-        self.rw_counter_offset = self.rw_counter_offset.clone()
-            + match &self.condition {
-                Some(condition) => condition.clone(),
-                None => 1.expr(),
-            } * rwc_inc;
+        self.rw_counter_offset = self.rw_counter_offset.clone() + self.condition_expr() * rwc_inc;
```
roynalnaruto, line 1135: here as well.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1357,4 +1349,11 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
             expr.clone()
         }
     }
+
+    fn condition_expr(&self) -> Expression<F> {
+        match &self.condition {
+            Some(condition) => condition.clone(),
```
roynalnaruto, line 1355: and lastly here.

roynalnaruto (approved):
Everything looks good!

Just a minor point, whether we should prefer `.clone()` or `.expr()`. I am leaning towards `.expr()` and I've explained why. It's up to you whether you wish to make that change :)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
z2trillion: The reason that that works in the log gadget is because there are not additional stack lookups afterwards. The `StepStateTransition` in the gadget correctly (but inconsistent with `stack_pointer_offset`) specifies that the stack point only changes by `2 + topic_count`, so the error doesn't propagate into subsequent gadgets.

PR #841, opened 2022-10-21
invalid jump error state circuit

### LINKED ISSUE #909: implement invalid jump error state circuit
PR #841 targets this issue !

DreamWuGit:
spec PR https://github.com/privacy-scaling-explorations/zkevm-specs/pull/296
   - remove `IsPersistent` for error type lookup
   - implement invalid jump error circuit 
   - refactor  `fn_gen_error_state_associated_ops` to handle all error in the future.


DreamWuGit:
@han0110  would you like to have a review ?

pinkiebell:
Note: I cancelled the integration tests run because it stalled

DreamWuGit:
sync with latest spec .

ChihChengLiang (approved):
LGTM, with a nitpick.

ChihChengLiang:
@han0110 do you have time to review this one? Asking since you reviewed the spec too.

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
```diff
@@ -0,0 +1,48 @@
...
+            None
+        };
+        exec_step.error = state.get_step_err(geth_step, next_step).unwrap();
+        // assert op code can only be JUMP or JUMPI
+        assert!(geth_step.op == OpcodeId::JUMP || geth_step.op == OpcodeId::JUMPI);
+        let is_jumpi = geth_step.op == OpcodeId::JUMPI;
+        let dest = geth_steps[0].stack.last()?.to_address();
+        let mut condition = 0;
+        if is_jumpi {
+            condition = geth_step.stack.nth_last(1)?.as_usize();
```
han0110, line 28: The `condition` is not always an `usize` I believe.

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
```diff
@@ -0,0 +1,48 @@
...
+        if is_jumpi {
+            state.stack_read(
+                &mut exec_step,
+                geth_step.stack.nth_last_filled(1),
+                condition.into(),
+            )?;
+        }
+        // `IsSuccess` call context operation is added in gen_restore_context_ops
+
+        state.gen_restore_context_ops(&mut exec_step, geth_steps)?;
```
han0110, line 44: It's confusing that `gen_restore_context_ops` only returns with not generating ops for restoring context when it's error:

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4cfccfa6c3b251284ff61eeb907d548d59206753/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L1179-L1190

We should remove this part handle the `IsSuccess` read in `ErrorInvalidJump::gen_associated_ops` and then do:

```suggestion
        if !call.is_root {
            state.gen_restore_context_ops(&mut exec_step, geth_steps)?;
        }
```

Or rename the `gen_restore_context_ops` to a better name.

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
DreamWuGit: yes, origin word type in evm stack, will update to RLC

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
DreamWuGit: if follow the suggestion, every error case should have the root check codes, reuse it as common helper would be good. I prefer to rename it like `gen_restore_context_ops_if_necessary` , what do you think ? @han0110

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
DreamWuGit: change to word type in [7f9cc36](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/841/commits/7f9cc36031e15ea711380b044469c4542c936212) thanks !

bus-mapping/src/evm/opcodes/error_invalid_jump.rs
han0110: I see, perhaps we should just keep it as is, we could refactor in a better way in the future.

han0110 (commented):
LGMT!

han0110 (approved):

PR #845, opened 2022-10-23
Handle invalid opcode

z2trillion:
Fixes `serial_test_evm_circuit_erc20_openzeppelin_transfer_fail` integration test.

### DIFF
--- a/zkevm-circuits/src/witness/bytecode.rs
+++ b/zkevm-circuits/src/witness/bytecode.rs
@@ -47,10 +47,8 @@ impl Bytecode {
             let is_code = push_data_left == 0;
 
             if is_code {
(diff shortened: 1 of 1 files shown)


leolara:
hi @z2trillion, i think this is a more comprehensive fix https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/846 for this type of issue, we do not need try_from, we only need from, as we have a complete mapping from u8 -> OpcodeId

z2trillion:
> hi @z2trillion, i think this is a more comprehensive fix #846 for this type of issue, we do not need try_from, we only need from, as we have a complete mapping from u8 -> OpcodeId

You're right. I didn't realize that was a possibility.

PR #848, opened 2022-10-24
fix call recursive test with invalid jump error

### LINKED ISSUE #849: `callop_recursive` test encounter invalid jump error
after `callop_recursive` test imports static call, jump dest will not be fixed value, current test meet `InvalidJump`

DreamWuGit:
fix issue #849, also need to adjust gas amount, otherwise will meet OOG Gas error 😄

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/callop.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/callop.rs
@@ -858,10 +858,12 @@ mod test {
         }
         caller_bytecode.append(&bytecode! {
             PUSH32(Address::repeat_byte(0xff).to_word())
-            PUSH2(10000)
+            // gas 10032 for static call
+            PUSH2(10000 + 32*(!is_call as u64))  // 10000 for call
             .write_op(opcode)
             STOP
         });
+        let (jump_dest, gas) = if is_call { (43, 132) } else { (41, 129) };
         // The following bytecode calls itself recursively if gas_left is greater than
         // 100, and halts with REVERT if gas_left is odd, otherwise just halts
         // with STOP.
@@ -869,7 +871,7 @@ mod test {
             GAS
             PUSH1(100)
             GT
-            PUSH1(43)
+            PUSH1(jump_dest) // jump dest
             JUMPI
 
             PUSH1(0)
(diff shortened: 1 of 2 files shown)


zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -858,47 +858,53 @@ mod test {
         }
         caller_bytecode.append(&bytecode! {
             PUSH32(Address::repeat_byte(0xff).to_word())
-            PUSH2(10000)
+            // gas 10032 for static call
+            PUSH2(10000 + 32*(!is_call as u64))  // 10000 for call
```
ChihChengLiang, line 862: It would be easier to reason by showing the final value
```suggestion
            PUSH2(if is_call {10000} else {10032})
```

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -858,47 +858,53 @@ mod test {
...
+        let (jump_dest, gas) = if is_call { (43, 132) } else { (41, 129) };
         // The following bytecode calls itself recursively if gas_left is greater than
         // 100, and halts with REVERT if gas_left is odd, otherwise just halts
         // with STOP.
         let mut callee_bytecode = bytecode! {
             GAS
             PUSH1(100)
             GT
-            PUSH1(43)
+            PUSH1(jump_dest) // jump dest
```
ChihChengLiang, line 874: what do you think we expand directly here?
```suggestion
            PUSH1(if is_call {43} else {41}) // jump dest
```

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -858,47 +858,53 @@ mod test {
...
+
         if is_call {
             callee_bytecode.push(1, U256::from(0));
         }
+
         callee_bytecode.append(&bytecode! {
             PUSH20(Address::repeat_byte(0xff).to_word())
-            PUSH1(132)
+            // gas
+            PUSH1(gas)
```
ChihChengLiang, line 890: what do you think we expand directly here?
```suggestion
            PUSH1(if is_call {132} else {129}) // gas
```

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -858,47 +858,53 @@ mod test {
...
             SUB
             .write_op(opcode)
 
-            JUMPDEST // 43
+            JUMPDEST // 41 for static_call, 43 for call
             GAS
             PUSH1(1)
             AND
-            PUSH1(56)
+            PUSH1(56 - 2 *(!is_call as u8))
```
ChihChengLiang, line 899:
```suggestion
            PUSH1(if is_call {56} else {54})
```

ChihChengLiang (approved):
Thanks for the fix!
Looks good to me after the comments are addressed.

zkevm-circuits/src/evm_circuit/execution/callop.rs
DreamWuGit: seems more clear ! will take your suggestion, hh

adria0 (approved):
LGTM!

PR #852, opened 2022-10-27
Optim/next step queries

lispc:


naure:
Currently, CellManager::new queries all rows of the next step. However, only the first row with state registers is necessary for step transition. This PR removes the unused queries.

This cuts the number of polynomial evaluations in almost half; proof size by 42%; verification time 3X. (with make evm_bench)

han0110 (approved):
LGTM! Nice refactor! I'd also like to have a function to sanity check if all queries are used in the gates/lookups, then we can avoid such unnecessary cost (issue #859 opened)

ed255 (approved):
LGTM!  Such a big improvement with a small change :D

PR #857, opened 2022-10-27
evm_circuit: conditionally enable `check_rw_lookup`

pinkiebell:
Fixes a big performance 'regression'.
Can be enabled with `RUST_LOG=debug` and otherwise saves > 200ms for each step assignment


lispc:
great! i did realized each step takes >200ms...

lispc (approved):

lispc:
since it is a small PR i will merge it directly

PR #858, opened 2022-10-28
Add capacity option to keccak circuits

Brechtpd:
See https://github.com/privacy-scaling-explorations/zkevm-chain/issues/67

No fixed capacity set in the super circuit yet however, not sure what's expected there.

### DIFF
--- a/zkevm-circuits/src/keccak_circuit/keccak_bit.rs
+++ b/zkevm-circuits/src/keccak_circuit/keccak_bit.rs
@@ -84,7 +84,7 @@ pub struct KeccakBitConfig<F> {
 #[derive(Default)]
 pub struct KeccakBitCircuit<F: Field> {
     witness: Vec<KeccakRow<F>>,
-    size: usize,
+    num_rows: usize,
     _marker: PhantomData<F>,
 }
 
@@ -119,10 +119,10 @@ impl<F: Field> Circuit<F> for KeccakBitCircuit<F> {
 
 impl<F: Field> KeccakBitCircuit<F> {
     /// Creates a new circuit instance
-    pub fn new(size: usize) -> Self {
+    pub fn new(num_rows: usize) -> Self {
         KeccakBitCircuit {
             witness: Vec::new(),
-            size,
+            num_rows,
             _marker: PhantomData,
         }
     }
@@ -130,12 +130,13 @@ impl<F: Field> KeccakBitCircuit<F> {
     /// The number of keccak_f's that can be done in this circuit
     pub fn capacity(&self) -> usize {
         // Subtract two for unusable rows
-        self.size / (NUM_ROUNDS + 1) - 2
+        self.num_rows / (NUM_ROUNDS + 1) - 2
     }
 
     /// Sets the witness using the data to be hashed
     pub fn generate_witness(&mut self, inputs: &[Vec<u8>]) {
-        self.witness = multi_keccak(inputs, KeccakBitCircuit::r(), Some(self.capacity()));
+        self.witness = multi_keccak(inputs, KeccakBitCircuit::r(), Some(self.capacity()))
+            .expect("Too many inputs for given capacity");
     }
 }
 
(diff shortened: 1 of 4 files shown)


zkevm-circuits/src/keccak_circuit/keccak_bit.rs
```diff
@@ -607,8 +607,9 @@ impl<F: Field> KeccakBitConfig<F> {
         layouter: &mut impl Layouter<F>,
         inputs: &[Vec<u8>],
         r: F,
+        capacity: Option<usize>,
```
ed255, line 610: The capacity units is number of permutations / keccak_f's that the circuit can handle right?

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
```diff
@@ -129,13 +129,13 @@ impl<F: Field> KeccakBitCircuit<F> {
 
     /// The number of keccak_f's that can be done in this circuit
     pub fn capacity(&self) -> usize {
-        // Subtract one for unusable rows
-        self.size / (NUM_ROUNDS + 1) - 1
+        // Subtract two for unusable rows
+        self.size / (NUM_ROUNDS + 1) - 2
```
ed255, line 133: What is the units of `self.size`?  Is it rows?

In case `size` is rows, and there are two unusable rows, and each keccak_f takes `NUM_ROUNDS +1` rows; shouldn't this be `(self.size - 2) / (NUM_ROUNDS + 1)`.  Or maybe I misunderstood something?

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
```diff
@@ -985,6 +986,12 @@ fn multi_keccak<F: Field>(bytes: &[Vec<u8>], r: F) -> Vec<KeccakRow<F>> {
     for bytes in bytes {
         keccak(&mut rows, bytes, r);
     }
+    if let Some(capacity) = capacity {
+        // Pad with no data hashes to the expected capacity
+        while rows.len() < (1 + capacity * (NUM_ROUNDS + 1)) {
```
ed255, line 991: I think it would be worth checking for the case where the inputs `bytes` require more than `capacity`, and return an error or at least `log::err`

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
Brechtpd: That's right. Would you prefer a different name?

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
Brechtpd: > What is the units of self.size? Is it rows?

Yeah it's number of rows, I'll rename the variable to `num_rows` because that's a much better name.

> In case `size` is rows, and there are two unusable rows, and each keccak_f takes `NUM_ROUNDS +1` rows; shouldn't this be `(self.size - 2) / (NUM_ROUNDS + 1)`. Or maybe I misunderstood something?

The code subtracts `2` of the capacity (number of keccak_f's) of the circuit, so`2 * (NUM_ROUNDS + 1)` rows of the circuit will not be used to make sure no unusable rows are needed.

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
Brechtpd: Agreed can't hurt to fail as early as possible, I do expect it to fail quickly after this with a not-enough-rows-error though.

zkevm-circuits/src/keccak_circuit/keccak_bit.rs
ed255: Ah, I see.  So you're reserving the rows corresponding to 2 permutations/keccak_f's for the blinding factors right?

ed255 (approved):
LGTM!  Thanks for addressing my questions and feedback :)

ChihChengLiang (approved):
LGTM

PR #861, opened 2022-10-28
returndatacopy: properly handle OOB access

pinkiebell:


bus-mapping/src/evm/opcodes/returndatacopy.rs
```diff
@@ -27,18 +27,21 @@ impl Opcode for Returndatacopy {
...
-            let minimal_length = dest_offset.as_usize() + length;
+
             if data_ends <= return_data.len() {
-                memory.extend_at_least(minimal_length);
                 memory[mem_starts..mem_ends].copy_from_slice(&return_data[data_starts..data_ends]);
-            } else {
-                assert_eq!(geth_steps.len(), 1);
-                // if overflows this opcode would fails current context, so
-                // there is no more steps.
+            } else if let Some(actual_length) = return_data.len().checked_sub(data_starts) {
```
han0110, line 40: Will this branch get chance to be executed? Copying return data out of bound should lead to exception and should be catch by `get_step_err` then handled by its own exception execution state IIUC.

pinkiebell: `returndatacopy` should behave like `calldatacopy` thus, out of bounds access is valid and just plain zeroes.

han0110: But from the implementation, it seems `RETURNDATACOPY` copies out of bound is specially handled:

```go
// copied from https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L322-L342
func opReturnDataCopy(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
	var (
		memOffset  = scope.Stack.pop()
		dataOffset = scope.Stack.pop()
		length     = scope.Stack.pop()
	)

	offset64, overflow := dataOffset.Uint64WithOverflow()
	if overflow {
		return nil, ErrReturnDataOutOfBounds
	}
	// we can reuse dataOffset now (aliasing it for clarity)
	var end = dataOffset
	end.Add(&dataOffset, &length)
	end64, overflow := end.Uint64WithOverflow()
	if overflow || uint64(len(interpreter.returnData)) < end64 {
		return nil, ErrReturnDataOutOfBounds
	}
	scope.Memory.Set(memOffset.Uint64(), length.Uint64(), interpreter.returnData[offset64:end64])
	return nil, nil
}
```

pinkiebell: Holy crap. There is no mention of this in the yellow paper but looks like [EIP-211](https://eips.ethereum.org/EIPS/eip-211) explicitly requires this.

bus-mapping/src/evm/opcodes/returndatacopy.rs
```diff
@@ -27,18 +27,21 @@ impl Opcode for Returndatacopy {
...
             let mem_ends = mem_starts + length;
             let data_starts = offset.as_usize();
             let data_ends = data_starts + length;
-            let minimal_length = dest_offset.as_usize() + length;
+
             if data_ends <= return_data.len() {
-                memory.extend_at_least(minimal_length);
                 memory[mem_starts..mem_ends].copy_from_slice(&return_data[data_starts..data_ends]);
-            } else {
-                assert_eq!(geth_steps.len(), 1);
```
pinkiebell, line 39: @han0110 Given this results in a out-of-gas condition,  shouldn't this error here be handled correctly?

han0110: I think this branch should never be entered, if this step is an exception if should be caught here:

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/d098f5fc40f6d86c1f8ecdd527d306e1bde6e755/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L974

bus-mapping/src/evm/opcodes/returndatacopy.rs
han0110: yeah that's one of easy-to-forget features of EVM 🥲

bus-mapping/src/evm/opcodes/returndatacopy.rs
pinkiebell: Ok. I was def. hitting this assert but I don't remember the details of it. I will test this again 👌

pinkiebell: [here we go](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/874)

pinkiebell:
🪦

PR #864, opened 2022-10-29
fix is_final in keccak multi pack

### LINKED ISSUE #860: keccak soundness bug
considering this patch:

```
diff --git a/zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs b/zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
index 48370937..862516e6 100644
--- a/zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
+++ b/zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
@@ -1978,7 +1978,7 @@ fn keccak<F: Field>(rows: &mut Vec<KeccakRow<F>>, bytes: &[u8], r: F) {
                     is_final,
                     length: round_lengths[round],
                     data_rlc: round_data_rlcs[round],
-                    hash_rlc,
+                    hash_rlc: hash_rlc + F::from_u128(row_idx as u128),
                     cell_values: regions[round].rows[row_idx].clone(),
                 });
             }
```

Tests still pass after this patch. I think it is not the expected behavior. 

the tuples in keccak table is (is_enabled, input_rlc, input_len, output_rlc) while in keccak multi pack implementation, we set 

```
        let is_final = keccak_table.is_enabled;
```

Other the other hand, for the final round of final chunk, the `get_num_rows_per_round` rows can all have `is_final == 1`. Then we just inject some incorrect values in non-first rows with `is_final == 1` so tests can still pass.

So we should change codes so that there can only be 1 row in a slot with `is_final == 1` rather than `get_num_rows_per_round` rows?

lispc:
fix https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/860


han0110 (commented):
Is this still a draft? Otherwise it looks good to me.

han0110 (approved):
LGTM!

ed255 (approved):
LGTM! Thanks for the fix!

Issue #586, opened 2022-06-21
Add unit tests for math gadgets


ed255:
In `src/evm_circuit/util/math_gadget.rs` we have several math gadgets that are used in opcode gadgets.  While the opcode gadgets have unit tests (which indirectly test the math gadgets), we don't have unit tests for the math gadgets.  Adding these unit tests would be very useful because it will allow us to test the math gadgets individually with their own edge cases.

Doing these unit tests may require creating some testing functionality because these gadgets use the `ConstraintBuilder` which is associated to EVM steps; but for unit testing them there's no actual step.

The list of math gadgets that require unit testing:
- [ ] `IsZeroGadget<F>`
- [ ] `IsEqualGadget<F>`
- [ ] `BatchedIsZeroGadget<F, const N: usize>`
- [ ] `AddWordsGadget<F, const N_ADDENDS: usize, const CHECK_OVREFLOW: bool>`
- [ ] `MulWordByU64Gadget<F>`
- [ ] `RangeCheckGadget<F, const N_BYTES: usize>`
- [ ] `LtGadget<F, const N_BYTES: usize>`
- [ ] `LtWordGadget<F>`
- [ ] `ComparisonGadget<F, const N_BYTES: usize>`
- [ ] `PairSelectGadget<F>`
- [ ] `ConstantDivisionGadget<F, const N_BYTES: usize>`
- [ ] `MinMaxGadget<F, const N_BYTES: usize>`
- [ ] `MulAddWordsGadget<F>`
- [ ] `ShrWordsGadget<F>`
- [ ] `MulAddWords512Gadget<F>`
- [ ] `ModGadget<F>`

xiaodino:
We are working on it.

roynalnaruto:
Some of the math gadgets that were also re-defined in the `gadgets` crate already have unit tests:
https://github.com/privacy-scaling-explorations/zkevm-circuits/tree/main/gadgets/src

So @xiaodino can refer them :)

roynalnaruto:
@smtmfft IMO we should also refactor the `math_gadget` module into modules per gadget in it, mainly to lower the lines-of-code in the entire module. What do others think about this? If yes, I'm happy to do this in a separate PR, or this can be done within #865 as well.

smtmfft:
Hi, @roynalnaruto You mean make each math gadget has its own file?? I think that's better structure, If the code does not change too much, I think this test would still works for them and I can add them to those files later on.

[...]

PR #865, opened 2022-10-31
Add unit test for math gadgets

smtmfft:
Address #586


smtmfft:
Pending on refactory, see #586 and #900

PR #870, opened 2022-11-01
fix evm padding

### LINKED ISSUE #868: bug: padding not work in evm circuit
reproduce:

set evm_circuit_pad_to to 2000, then

```
cargo test -p zkevm-circuits --all-features --release stop_gadget_simple -- --nocapture
```

it will reports ""evm circuit offset larger than padding: 2000 > 2000"

@ed255

pinkiebell:
Fixes: #868

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -773,18 +773,20 @@ impl<F: Field> ExecutionConfig<F> {
 
                 let evm_rows = block.evm_circuit_pad_to;
                 let exact = evm_rows == 0;
-                let mut block_step = end_block_not_last;
+                let mut no_next_step = false;
                 let mut get_next = |offset: &usize| match steps.next() {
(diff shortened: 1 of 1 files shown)


ed255:
Here are some issues I found while reproducing https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/868 which are not yet applied in this PR:
1. We need a unit test that sets `evm_circuit_pad_to != 0`
2. The number of EndBlock rows is calculated incorrectly because it starts from the offset of the previous to last trace step!  So we need to add `self.get_step_height(cur_state)`
3. The EndBlock has a bug when it's enabled in a `no_step_last` (which will never happen if `evm_circuit_pad_to == 0`).  The bug is that the lookups are enabled in all locations (not just `no_step_last`).  They should either be only enabled in `no_step_last` (to follow the spec), or be enabled everywhere but then assign the witness everywhere as well.  I currently don't know which is the best option (but the second one is easy).

pinkiebell:
closed in favour of #879

PR #871, opened 2022-11-01
Merge return & revert state in the ExecutionState

icemelon:


icemelon:
cc @z2trillion

ChihChengLiang (approved):
LGTM. Great addition

han0110 (approved):
LGTM!

Also when the case `RETURN && is_create`, it should update current callee's `code_hash`, but I can't find it anywhere, is that a missing lookup?

z2trillion:
> Also when the case `RETURN && is_create`, it should update current callee's `code_hash`, but I can't find it anywhere, is that a missing lookup?

Yes, it's missing at the moment. I am working on updating the specs and the circuit to include this.

PR #872, opened 2022-11-02
Fully implemented `BeginTx`

han0110:
This PR aims to fully implement `BeginTx` to also handle tx to account with empty code.

Previously `BeginTx` to account with empty code is left unimplemented, so the block containing simple transaction will fail like https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/790#issuecomment-1294440185.

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -9,7 +9,7 @@ use crate::{
                 ConstraintBuilder, ReversionInfo, StepStateTransition,
                 Transition::{Delta, To},
             },
-            math_gadget::{IsZeroGadget, MulWordByU64Gadget, RangeCheckGadget},
+            math_gadget::{IsEqualGadget, IsZeroGadget, MulWordByU64Gadget, RangeCheckGadget},
             select, CachedRegion, Cell, RandomLinearCombination, Word,
         },
         witness::{Block, Call, ExecStep, Transaction},
@@ -20,6 +20,7 @@ use crate::{
 use eth_types::{evm_types::GasCost, Field, ToLittleEndian, ToScalar};
 use halo2_proofs::circuit::Value;
 use halo2_proofs::plonk::Error;
+use keccak256::EMPTY_HASH_LE;
 
 #[derive(Clone, Debug)]
 pub(crate) struct BeginTxGadget<F> {
@@ -39,6 +40,7 @@ pub(crate) struct BeginTxGadget<F> {
     sufficient_gas_left: RangeCheckGadget<F, N_BYTES_GAS>,
     transfer_with_gas_fee: TransferWithGasFeeGadget<F>,
     code_hash: Cell<F>,
+    is_empty_code_hash: IsEqualGadget<F>,
 }
 
 impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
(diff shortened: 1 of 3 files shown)


zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -155,62 +157,103 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
-            (CallContextFieldTag::IsCreate, tx_is_create.expr()),
-            (CallContextFieldTag::CodeHash, code_hash.expr()),
-        ] {
-            cb.call_context_lookup(true.expr(), Some(call_id.expr()), field_tag, value);
-        }
+        );
+
+        cb.condition(is_empty_code_hash.expr(), |cb| {
+            cb.require_equal(
+                "Tx to account with empty code should be successful",
```
ed255, line 171: This string would make me think that the check is for `is_success`.  I know that in this context `is_persistent == is_success`.  What do you think about saying "should be persistent" instead?

ed255 (approved):
LGTM! I've added a very small nitpick, feel free to ignore it if you don't agree :)

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: Make sense, fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/872/commits/ba855da94f8ee361e82456f880de5a1beb7a1f18.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -155,62 +157,103 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
+        );
+
+        cb.condition(is_empty_code_hash.expr(), |cb| {
+            cb.require_equal(
+                "Tx to account with empty code should be persistent",
+                reversion_info.is_persistent(),
+                1.expr(),
+            );
+            cb.require_equal(
+                "Go to EndTx when Tx to account with empty code",
```
DreamWuGit, line 176: so this assumes tx to empty account must be in root call ? if there is internal call to empty account  will not go to `EndTx`

DreamWuGit (approved):
LGTM! only one question as comment!

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
han0110: No, call to account with empty code could happen in both case, but for internal call, it just need to restore the caller's context.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/997293041e57bd727d38f24f7e976984b8ec4268/zkevm-circuits/src/evm_circuit/execution/callop.rs#L223-L245

DreamWuGit: thanks, Callop handle it, another question :  Can we ensure Tx to account with empty code must be persistent ?  tx can failed due to many reason  i.e. Out of Gas, right ?
```
 cb.condition(is_empty_code_hash.expr(), |cb| {
            cb.require_equal(
                "Tx to account with empty code should be persistent",
                reversion_info.is_persistent(),
                1.expr(),
            );
```

han0110: `BeginTx` handles the successful case, which doesn't consider invalid transaction. We need to handle  transactions with insufficient intrinsic gas or invalid signature, and treat it as no-op, but that not be our current focus.

DreamWuGit: during I handle error case, I found even for step(opcode) failed,  the tx steps also consist of begin_tx . for example, begin_tx --> push1 --> push2(OOG occurs) --> end_tx ,  So I understand successful case is for final tx status , not the call scope status , right ?

han0110: In this case, it's not an account with empty code, right?

DreamWuGit: yes, I suspect in the case "n account with empty code", if error happens, the `is_persistent` is not true then . maybe better to an issue to warn us to check and add such test later.

han0110: make sense, but I guess we could just left it to future implementer who implements invalid transaction. (invalid transaction I mean which never gets included in a L1 block, but we might need to support that in circuit in order to accept force-included transaction for censorship-resistance).

DreamWuGit: sounds good!

Brechtpd: > make sense, but I guess we could just left it to future implementer who implements invalid transaction. (invalid transaction I mean which never gets included in a L1 block, but we might need to support that in circuit in order to accept force-included transaction for censorship-resistance).

Oh we are actually working on this, didn't think this would be generally useful but makes sense. We're only interested in being able to skip over transactions with invalid nonce/insufficient ETH (because these depend on the latest chain state), but we'll also do insufficient intrinsic gas/invalid signature if other people need it (though for forced transaction inclusion I think you could also just check this in the smart contract because you only need the tx data itself to verify these, but then you need extra solidity code as well to be able to do this so make sense to reuse circuits I guess).

han0110: > Oh we are actually working on this, didn't think this would be generally useful but makes sense. We're only interested in being able to skip over transactions with invalid nonce/insufficient ETH

That's really nice! I think no matter what in the end we'd definitely like to move all transaction validation logic into circuit to reduce verifier cost (right now we could only validate all in smart contract).

Issue #875, opened 2022-11-02
Integration tests failing


pinkiebell:
[See workflow](https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3377880208/jobs/5607397059#step:13:219)

pinkiebell:
[Another workflow with the same error but more information](https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3381189420/jobs/5614830134#step:13:247) 🤔

ed255:
oh, the super circuit integration test takes a long time (like 50 minutes)... I think we'll have to disable it for PRs (a part from fixing the issue), and only let it run once the commit is merged to main.

ed255:
> oh, the super circuit integration test takes a long time (like 50 minutes)... I think we'll have to disable it for PRs (a part from fixing the issue), and only let it run once the commit is merged to main.

I noticed that we're not using the `verify_par()` function from the mock prover!  That should give a significant speed-up on the integration tests when using the super circuit.

PR #876, opened 2022-11-02
workflows/integration: limit circuit test threads

pinkiebell:
This fixes the integration test workflow that exits with out-of-memory. This started to become a problem after the super circuit was integrated.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -77,5 +77,5 @@ jobs:
       - run: ./run.sh --steps "gendata"
       - run: ./run.sh --steps "tests" --tests "rpc"
       - run: ./run.sh --steps "tests" --tests "circuit_input_builder"
-      - run: ./run.sh --steps "tests" --tests "circuits"
+      - run: RUST_TEST_THREADS=2 ./run.sh --steps "tests" --tests "circuits"
       - run: ./run.sh --steps "cleanup"


DreamWuGit (approved):
LGTM

DreamWuGit:
@pinkiebell  I am wondering this PR CI also failed , how to merge it ?

pinkiebell:
> @pinkiebell I am wondering this PR CI also failed , how to merge it ?

Ya. I already opened #875 regarding that error. Regarding merge:
`At least 1 approving review is required by reviewers with write access. `

ed255:
> @pinkiebell I am wondering this PR CI also failed , how to merge it ?

I think currently this repository is configured so that the Integration Tests workflow is **not** required to pass in order to merge the PR

ed255:
In https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/879 I've managed to fix the integration tests.  I've also applied the thread limit, although I've used the `--test-threads 1` flag for the test binary instead of the env var.  The reason for setting only 1 thread is because I believe the super circuit verification with the mock prover takes around 32 GB of memory, and if we're unlucky and 2 such tests run at the same time we will also get an OOM.  The integration-test-runner has 64 GB of memory.

pinkiebell:
> In #879 I've managed to fix the integration tests. I've also applied the thread limit, although I've used the `--test-threads 1` flag for the test binary instead of the env var. The reason for setting only 1 thread is because I believe the super circuit verification with the mock prover takes around 32 GB of memory, and if we're unlucky and 2 such tests run at the same time we will also get an OOM. The integration-test-runner has 64 GB of memory.

I see. I think the env variable makes more sense because the flag applies to all tests

ed255:
> > In #879 I've managed to fix the integration tests. I've also applied the thread limit, although I've used the `--test-threads 1` flag for the test binary instead of the env var. The reason for setting only 1 thread is because I believe the super circuit verification with the mock prover takes around 32 GB of memory, and if we're unlucky and 2 such tests run at the same time we will also get an OOM. The integration-test-runner has 64 GB of memory.
> 
> I see. I think the env variable makes more sense because the flag applies to all tests

oh right! I'll update the PR to use the end var

pinkiebell:
Closed in favor of #879

PR #878, opened 2022-11-03
Integration tests to use actual prover

### LINKED ISSUE #789: Add real prover tests for the integration tests
Currently we have 4 kinds of tests in this repository:
1. unit tests, using the mock prover.  Run for every PR
2. integration tests (using real blocks from geth), using the mock prover.  Run for every commit merged to main
3. benchmarks, using mock data and the real prover.  Run on demand via PR labels
4. testool, (using official Ethereum test vectors) using the mock prover.  Plan to run them daily

Sometimes we have bugs in our circuits that wont' show up in the mock prover but will appear in the real prover.  We should have tests using real data that use the real prover, which should be run either for every commit merged to main or daily.

The challenge here is that tests for the real prover require a machine with very high resources, so we should be smart about it.

leolara:
Notes:
 + Had to change rw_map in state circuit to ::default() in order to pass, I don't know if this is a bug in the circuit
 + super_circuit integration test was added after I did most of the code, so it is the only one that is not using the actual prover, I will add it in another PR.
 + super_circuit mock prover was using a lot of memory so I disconnected that test to test the rest of the tests. 

Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/789


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -365,12 +365,14 @@ pub mod test {
         run_test_circuit(block)
     }
 
-    pub fn run_test_circuit<F: Field>(block: Block<F>) -> Result<(), Vec<VerifyFailure>> {
```
leolara, line 368: previous public function did not have a documentation comment, and clippy didn't give a warning, I guess because the module is called test? So, I didn't document the other test functions

leolara:
@JohnWick2ETH please could you review

integration-tests/tests/circuits.rs
```diff
@@ -100,33 +237,49 @@ async fn test_tx_circuit_block(block_num: u64) {
...
+    let randomness = CopyCircuitTester::<Fr>::get_randomness();
+    let circuit = CopyCircuitTester::<Fr>::new(block, randomness);
+
+    let num_rows = 1 << DEGREE;
+    const NUM_BLINDING_ROWS: usize = 7 - 1;
+    let instance = vec![vec![randomness; num_rows - NUM_BLINDING_ROWS]];
 
+    test_run(rng, DEGREE, circuit, instance);
+}
 pub async fn test_super_circuit_block(block_num: u64) {
```
ed255, line 283: Can you add a comment with a `TODO` for the super circuit to add a test with the real prover?

integration-tests/tests/circuits.rs
```diff
@@ -60,32 +199,30 @@ async fn test_state_circuit_block(block_num: u64) {
     let storage_ops = builder.block.container.sorted_storage();
     trace!("storage_ops: {:#?}", storage_ops);
 
-    const DEGREE: usize = 17;
-
-    let rw_map = RwMap::from(&builder.block.container);
+    // TODO: use rw_map from block
+    // let rw_map = RwMap::from(&builder.block.container);
+    let rw_map = RwMap::default();
```
ed255, line 204: We should fix this.  A `RwMap::default()` will just generate an empty state circuit, which is not verifying anything.
Did you get a failure when running the mock prover for this test?  Or was it for the real prover?

integration-tests/tests/circuits.rs
```diff
@@ -35,21 +52,143 @@ const CIRCUITS_PARAMS: CircuitsParams = CircuitsParams {
...
+            verifier_params,
+            verifying_key,
+            strategy,
+            &[instances],
+            &mut verifier_transcript,
+        )
+        .expect("failed to verify bench circuit");
+    }
+
+    let mock_prover = MockProver::<Fr>::run(degree, &circuit, instance.clone()).unwrap();
```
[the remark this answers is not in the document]
ed255: I just saw that the integration tests from the workflow took 56 minutes, because the real prover tests were enabled.  We should split them and not run them on every PR.

ed255 (commented):
Overall looks good! Please take a look at the various comments and requests I've added :)

integration-tests/tests/circuits.rs
```diff
@@ -35,21 +52,143 @@ const CIRCUITS_PARAMS: CircuitsParams = CircuitsParams {
...
+    mock_prover
+        .verify_par()
+        .expect("mock prover verification failed");
+
+    let general_params = ParamsKZG::<Bn256>::setup(degree, &mut rng);
+    let verifier_params: ParamsVerifierKZG<Bn256> = general_params.verifier_params().clone();
+
+    let verifying_key = keygen_vk(&general_params, &circuit).expect("keygen_vk should not fail");
+    let proving_key =
+        keygen_pk(&general_params, verifying_key, &circuit).expect("keygen_pk should not fail");
```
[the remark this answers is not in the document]
ed255: I think this is a great idea!  And I believe we currently have some bugs that would make this fail haha. (I think there's a bug when a block has less txs than the max_txs of the circuit).

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -365,12 +365,14 @@ pub mod test {
...
-        let fixed_table_tags = detect_fixed_table_tags(&block);
+    pub fn test_circuit_degree<F: Field>(block: &Block<F>) -> u32 {
```
hero78119, line 368: This function name `test_circuit_degree` (and so as `test_circuit_instance`, `test_cicuit_from_block`...) looks like test sth, but it turn out that it's like getting XXX for testing circuit, cause a bit confusing. How about rename to get_degree(_from_block), get_instance, gen_circuit...etc

integration-tests/tests/circuits.rs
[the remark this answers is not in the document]
hero78119: Using macro is a good idea. With macro, instead of using `mock: true`, maybe can further separate into `test_with_mock_prover`/`test_with_real_prover` for codee clearness

integration-tests/tests/circuits.rs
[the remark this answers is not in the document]
leolara: ok, so I use the circuit::without_witnesses() the keys and then I reuse those for all the tests with something like a singleton?

integration-tests/tests/circuits.rs
```diff
@@ -100,33 +237,49 @@ async fn test_tx_circuit_block(block_num: u64) {
 
     let prover = MockProver::run(DEGREE, &circuit, vec![vec![]]).unwrap();
 
-    prover.verify_par().expect("tx_circuit verification failed");
+    prover.verify().expect("tx_circuit verification failed");
```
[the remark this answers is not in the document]
leolara: yes

integration-tests/tests/circuits.rs
```diff
@@ -35,21 +52,143 @@ const CIRCUITS_PARAMS: CircuitsParams = CircuitsParams {
...
+    eth_types::Block<eth_types::Transaction>,
+) {
     let cli = get_client();
     let cli = BuilderClient::new(cli, CIRCUITS_PARAMS).await.unwrap();
-    let (builder, _) = cli.gen_inputs(block_num).await.unwrap();
+
+    cli.gen_inputs(block_num).await.unwrap()
+}
+
+fn test_run<ConcreteCircuit: Circuit<Fr>, R: RngCore>(
```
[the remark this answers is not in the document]
leolara: I did it like in here:

https://github.com/zcash/halo2/blob/main/halo2_proofs/src/plonk/prover.rs#L40

but I don't have a personal preference, I change it

integration-tests/tests/circuits.rs
leolara: ok

integration-tests/tests/circuits.rs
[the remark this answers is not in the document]
leolara: yes, sorry

zkevm-circuits/src/evm_circuit.rs
leolara: ok, I would do get_test_degree, get_test_instance and so on...

integration-tests/tests/circuits.rs
```diff
@@ -35,21 +52,143 @@ const CIRCUITS_PARAMS: CircuitsParams = CircuitsParams {
...
+        )
+        .expect("failed to verify bench circuit");
+    }
+
+    let mock_prover = MockProver::<Fr>::run(degree, &circuit, instance.clone()).unwrap();
+    mock_prover
+        .verify_par()
+        .expect("mock prover verification failed");
+
+    let general_params = ParamsKZG::<Bn256>::setup(degree, &mut rng);
```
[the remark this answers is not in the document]
leolara: Like a singleton that returns the same params for the same degree would be a good idea to start with? Or you mean doing all tests with degree 18?

integration-tests/tests/circuits.rs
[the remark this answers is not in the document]
ed255: > ok, so I use the circuit::without_witnesses() the keys and then I reuse those for all the tests with something like a singleton?

Yes!  To be able to reuse the same setup for all tests, while keeping the tests with the same structure, I recommend you to use https://docs.rs/lazy_static/1.4.0/lazy_static/ to have this global singleton.  It should be one global variable per circuit.  With lazy_static you can add the setup code, and then each test will make sure that the setup code is only run once (so running tests in parallel is fine too).

integration-tests/tests/circuits.rs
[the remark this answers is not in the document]
ed255: Ah, thanks for the pointer!  I would guess that they used that long name because `C` was already used?  Anyway, since you found this reference in the halo2 code (and the use case here is the same), I would leave it up to you to decide if you want to change it or not :)

integration-tests/tests/circuits.rs
leolara:
```
thread 'serial_test_state_circuit_erc20_openzeppelin_transfer_succeed' panicked at 'proof generation should not fail: Synthesis', integration-tests/tests/circuits.rs:96:10
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: circuits::test_run
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: tokio::runtime::scheduler::current_thread::Context::enter
   7: tokio::macros::scoped_tls::ScopedKey<T>::set
   8: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
   9: tokio::runtime::Runtime::block_on
  10: core::ops::function::FnOnce::call_once
```

how do I find out where is the problem?

integration-tests/src/integration_test_circuits.rs
```diff
@@ -0,0 +1,332 @@
...
+    max_txs: 4,
+    keccak_padding: None,
+};
+
+fn get_general_params(degree: u32) -> ParamsKZG<Bn256> {
+    match GEN_PARAMS.lock().unwrap().get(&degree) {
+        Some(params) => params.clone(),
+        None => {
+            let params = ParamsKZG::<Bn256>::setup(degree, RNG.clone());
+            GEN_PARAMS.lock().unwrap().insert(degree, params.clone());
```
ed255, line 65: This looks like a double lock on `GEN_PARAMS`.  Did you test this successfully?

integration-tests/src/integration_test_circuits.rs
```diff
@@ -0,0 +1,332 @@
...
+    trace!("storage_ops: {:#?}", storage_ops);
+
+    // TODO: use rw_map from block
+    // let rw_map = RwMap::from(&builder.block.container);
+    let rw_map = RwMap::default();
+
+    let circuit = StateCircuit::<Fr>::new(rw_map, 1 << 16);
+    let instance = circuit.instance();
+
+    test_mock(DEGREE, &circuit, instance.clone());
```
ed255, line 218: I think it would be better if all `test_mock` are only run when `!actual`.
Like: `if actual { test_actual } else { test_mock }`.

This is because the test_mock for big circuits (even with `verify_par`) can take some time.

integration-tests/tests/circuits.rs
ed255: When implementing the `Circuit` trait (from halo2), the error type for the `synthesize` method is already defined, and there's no way to add extra info in it.  The synthesis step usually does a lot of processing with the inputs (in order to calculate intermediate witness values), and there are many places that can return error.  Since we can't encode more details in the error itself, we have added lots of log::error right before the error is returned, like this:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/de26a9d4a8bfa1f416d65db805c091c4156d3ae9/zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs#L425-L430

Did you see any log before the panic backtrace?

ed255:
Overall the PR looks good to me!  The main thing that would be good to fix is the state_circuit error.  But if we can't manage to resolve it quickly, I think it's OK to move on with this PR and fix it via a new PR.

Some points that I believe can be improved in the future, but I think we can skip in this PR (to avoid delaying it too much) are:
- Reuse the `verifying_key` and `proving_key` between different tests of the same circuit (to make sure the circuit setup is equal no matter the input)
- Currently `integration-tests/tests/circuits_actual.rs` and `integration-tests/tests/circuits_mock.rs` are mostly duplicates.  I think we can find a way to reduce the code duplication.

integration-tests/src/integration_test_circuits.rs
leolara: Yes, in some way this is a oversight from my part as I should have seen it. The fact of adding Mutex is just an easy way to allow a mutable lazy_init, so it was artificial and I didn't look enough into it.

However, it works, because I think rust drops it as an optiomisation when the lock is not used anymore right after the lock. But as I saw in other code that is not always the same and we should not relay on this. There were a ticket open by pinkie about this.

I will make it lock just once.

leolara:
@ed255
 + I was working on caching the keys already so it is done, with the exception of state circuit where degree is not static it has been done.
  + I tried code deduplication for what you say but the need for macros make it difficult or make the code obscure if you force. So I don't know what is the best tool for this, so we can leave it for another PR.
  + I think we should move the problem of the state_circuit rwmap to another issue.

leolara:
I have fixed the rw_map issue using new_from_block as adiviced by @ed255

leolara:
I have added these tasks that have come out of this process but will be done in other PRs:

 + https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/966
 + https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/967
 + https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/968

integration-tests/tests/circuits.rs
leolara: Solved by using new_from_block

leolara:
The bytecode circuit is not passing despite puting it size to 2usize.pow(BYTECODE_CIRCUIT_DEGREE).

leolara:
State circuit also doesn't pass using new_from_block

leolara:
@ed255 passing all checks

ed255 (approved):
LGTM!

ed255:
@kunxian-xia could you take another look to this PR and approve if you find it ok?

The current status is that some tasks will be done in future PR following this list https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/878#issuecomment-1338833835
The mock prover tests are all passing, and some real prover tests are not passing and will be fixed in future PRs (we need to investigate what is failing).

kunxian-xia:
LGTM!

PR #885, opened 2022-11-08
Add challenge API to state circuit

### LINKED ISSUE #855: Refactor State circuit with challenge API

adria0:
This PR removes the directly injected randomness for RLP `Word` commitments and replaces it with the provided Challenge API value (for witness) and expression (for constraints).

Note that in some points is required to change from `F` to `Value<F>` (API provides `Value<F>` as the value of the challenge) meaning that, we must must use methods like `map` to propagate the "knowness" of different values.


ed255 (changes requested):

ed255 (approved):
LGTM! Nice work! :dancing_men:

miha-stopar (approved):
LGTM!

PR #887, opened 2022-11-09
fix empty block ExecutionState potiential mismatch issue

hero78119:
fix bug mentioned in #867

### DIFF
--- a/circuit-benchmarks/src/evm_circuit.rs
+++ b/circuit-benchmarks/src/evm_circuit.rs
@@ -10,14 +10,12 @@ use zkevm_circuits::table::{BlockTable, BytecodeTable, RwTable, TxTable};
 
 #[derive(Debug, Default)]
 pub struct TestCircuit<F> {
-    block: Block<F>,
+    block: Option<Block<F>>,
 }
 
-impl<F: Default> TestCircuit<F> {
-    pub fn new() -> TestCircuit<F> {
-        let mut test_circuit = TestCircuit::default();
-        test_circuit.block = Block::new();
-        test_circuit
+impl<F: Field> TestCircuit<F> {
+    pub fn new(block: Block<F>) -> Self {
+        TestCircuit { block: Some(block) }
     }
 }
 
@@ -59,7 +57,8 @@ impl<F: Field> Circuit<F> for TestCircuit<F> {
         config: Self::Config,
         mut layouter: impl Layouter<F>,
     ) -> Result<(), Error> {
-        config.assign_block(&mut layouter, &self.block)?;
+        let block = self.block.as_ref().unwrap();
+        config.assign_block(&mut layouter, block)?;
         Ok(())
     }
 }
@@ -68,6 +67,8 @@ impl<F: Field> Circuit<F> for TestCircuit<F> {
 mod evm_circ_benches {
     use super::*;
     use ark_std::{end_timer, start_timer};
+    use bus_mapping::{circuit_input_builder::CircuitsParams, mock::BlockData};
+    use eth_types::geth_types::GethData;
     use halo2_proofs::plonk::{create_proof, keygen_pk, keygen_vk, verify_proof};
     use halo2_proofs::poly::kzg::commitment::{KZGCommitmentScheme, ParamsKZG, ParamsVerifierKZG};
     use halo2_proofs::poly::kzg::multiopen::{ProverSHPLONK, VerifierSHPLONK};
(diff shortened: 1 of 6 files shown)


zkevm-circuits/src/witness/block.rs
```diff
@@ -4,13 +4,17 @@ use bus_mapping::circuit_input_builder::{self, CircuitsParams, CopyEvent, ExpEve
...
+    table::BlockContextFieldTag,
+    evm_circuit::step::ExecutionState,
+};
 
 use super::{step::step_convert, tx::tx_convert, Bytecode, ExecStep, RwMap, Transaction};
 
-/// Block is the struct used by all circuits, which constains all the needed
+/// Block is the struct used by all circuits, which constraints all the needed
 /// data for witness generation.
+/// TODO: figure out better way to deal with potential ExecutionState mismatch when invoke default trait
```
han0110, line 17: Since `Block::default()` is no longer valid, I think it's reasonable to remove the `Default` derive and use `Block::new` to create a valid empty block.

Another way is to call `Block::new()` inside `Block::default`, tho it might be confusing sometimes.

circuit-benchmarks/src/evm_circuit.rs
```diff
@@ -13,6 +13,14 @@ pub struct TestCircuit<F> {
     block: Block<F>,
 }
 
+impl<F: Default> TestCircuit<F> {
+    pub fn new() -> TestCircuit<F> {
+        let mut test_circuit = TestCircuit::default();
+        test_circuit.block = Block::new();
+        test_circuit
+    }
+}
```
han0110, line 22: With `TestCircuit::new` I think we can also remove the `Default` derive of `TestCircuit`.

han0110 (approved):
LGTM! Thanks for catching this!

circuit-benchmarks/src/evm_circuit.rs
```diff
@@ -83,7 +91,7 @@ mod evm_circ_benches {
             .parse()
             .expect("Cannot parse DEGREE env var as u32");
 
-        let circuit = TestCircuit::<Fr>::default();
+        let circuit = TestCircuit::<Fr>::new();
```
ed255, line 94: I would like to suggest a different approach, that hopefully will be more robust in the future.  Building the witness of an empty block, like it is done in this test https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/dac2741dd36d45f3a3ded8fc782f657c936bb8be/zkevm-circuits/src/evm_circuit.rs#L416

So it should look like this:
```rust
let geth_data: GethData = TestContext::<0, 0>::new(None, |_| {}, |_, _| {}, |b, _| b)
    .unwrap()
    .into();
let mut builder =
    BlockData::new_from_geth_data_with_params(geth_data.clone(), CircuitsParams::default())
        .new_circuit_input_builder();
builder
    .handle_block(&geth_data.eth_block, &geth_data.geth_traces)
    .unwrap();
let block = block_convert(&builder.block, &builder.code_db);

let circuit = TestCircuit { block };
```

And then remove the `Default` from `Block` and from `TestCircuit`

hero78119: Thanks for the suggestions! Current version, once adding new field, compiler wont hint a implicit default setting, which rely on developer awareness. Removing default trait entirely eliminate this potiential mismatch from earth, and `block` should be created via empty block anyway.

Since many place rely on block `default` trait, such as super_circuit, exp_circuit, i will modify them accordingly :)

hero78119: Updated pull request. Only remove `Default` from `Block`, in order to keep consistent coding style for all circuit `without_witnesses` which still rely circuit to have `Default` trait. For `block` field in circuit, it will be wrapped with `Option<Block>`. 

Let me know whether you have better idea :)

ed255 (approved):
LGTM!

circuit-benchmarks/src/evm_circuit.rs
ed255: It does look a bit strange to have `Option<Block>` in the circuits, but then it makes sense that a circuit without the witness doesn't have a block.  I can't think of a better idea, so this way looks good to me :)

ChihChengLiang:
Hi @hero78119, thanks for the great contribution! Could you do a rebase before we merge it?

ChihChengLiang:
@hero78119 one more `make fmt` and we're good to go :)

hero78119:
Hi @ChihChengLiang still CI running, this time should be ok now.
(Just realized it's not format error, instead just some dependencies which only being used by test mod need to be requested in test mod)

PR #896, opened 2022-11-11
add keccak_padding padding option

pinkiebell:
And make use of it inside the super circuit.


ChihChengLiang (approved):
LGTM. 

For some background for other reviewers, this is part of https://github.com/privacy-scaling-explorations/zkevm-chain/issues/67. We need to specify the max rounds of keccak we can perform in the production circuit so that the circuit has soundness.

PR #906, opened 2022-11-15
workflows/integration: only run integration tests on request

pinkiebell:
1. A bot will comment on a new PR with a help message
2. If someone comments with or edit a comment to `/test` this workflow will be trigged to attach a `trigger-integration-tests` label
3. The label will trigger this workflow again but this time inside the PR context - this triggers github to display `checks` accordingly.
4. The label is removed again because it's considered only a one-time trigger.
5. integration tests are run.

- The label trigger is required for github to recognize any status changes regarding the workflow.
- A custom github access token is required to let the workflow trigger itself again. This doesn't work with githubs default token.
- This workflow uses `pull_request_target` to allow to run it on pull requests via forks and explicitly checkouts the ref of the pull requests.


pinkiebell:
I can't test it here, because the workflow needs to be on `main` 😁
Regarding the access token, github seems to have a beta feature that allows us to manage them from the organizational setttings view.

pinkiebell:
Testbed: https://github.com/abcdeftest/test/pull/1

ed255 (approved):
LGTM!

I have a few question:
1. I didn't find the label `trigger-integration-tests` defined for this repository.  Do we need to add it?
2. I would like to understand how the `add-label` and `remove-label` interact with each other.  In the same job, the label will be added and removed?  Or is it added in one run, and removed in the next one?

pinkiebell:
Screenshot 2022-11-18 at 12 34 10
Note: Github has a beta feature for keeping track of personal access token of members

ChihChengLiang:
Do you mean the integration test can only be triggered by someone with an access token?

How does this PR affect the external contributors? Do they need our help to trigger the integration tests to reach the mergeable state?

pinkiebell:
> Do you mean the integration test can only be triggered by someone with an access token?

No, we just have to setup a personal access token and add it to the repository secrets.

> 
> How does this PR affect the external contributors? Do they need our help to trigger the integration tests to reach the mergeable state?

Right now, everyone can comment on the PR with `/test` and trigger a workflow run.

ChihChengLiang:
Understood. Sounds good to me.

ChihChengLiang (approved):

ChihChengLiang:
@pinkiebell The integration tests are not triggered. Does this pr prevent it from running?

pinkiebell:
> @pinkiebell The integration tests are not triggered. Does this pr prevent it from running?

This only works once this workflow is merged to main.

pinkiebell:
@ChihChengLiang Yes. Looks it prevents it. I think what happens is that this workflow already overrides the origin but isn't triggered because it needs to be on the main branch for that to happen. Ouch!

ChihChengLiang:
Alright, let's ape merge then.

PR #917, opened 2022-11-18
Get curr_memory_word_size from constraint builder

z2trillion:
`MemoryExpansionGadget::construct` is only ever called with `curr_memory_word_size` is always `cb.curr.state.memory_word_size.expr()` so it doesn't need to get an argument to the function.


roynalnaruto (approved):
Looks good!

ChihChengLiang (approved):
Nice fix! The code looks tidier now.

PR #918, opened 2022-11-21
fix super circuit bench

lispc:
super circuit bench does not work well now due to 2 reasons:

1. bytecode cirucuit `size` underflows, and it tries to assign witness till row 1<<64..
2. DEGREE env var also be used in keccak circuit, resulting incorrect behavior.

after this fix:

```
running 1 test
End:     Setup generation with degree = 19 .........................................4.456s
test super_circuit::tests::bench_super_circuit_prover has been running for over 60 seconds
Start:   SuperCircuit Proof generation with degree = 19
End:     SuperCircuit Proof generation with degree = 19 ............................2754.823s
Start:   SuperCircuit Proof verification
End:     SuperCircuit Proof verification ...........................................293.819ms
test super_circuit::tests::bench_super_circuit_prover ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 8 filtered out; finished in 2871.63s
```


zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -406,6 +406,7 @@ impl<F: Field> Config<F> {
         let length_is_zero_chip = IsZeroChip::construct(self.length_is_zero.clone());
 
         // Subtract the unusable rows from the size
+        debug_assert!(size > self.minimum_rows);
```
ed255, line 409: The `bench` profile has `debug-assertions` and `overflow-checks` disabled, so in a benchmark this check will not be done (and neither the possible overflow in the next line).
I think we don't have code in the critical path in this repository, so I think we should use `assert!` instead of `debug_assert!` everywhere.  Only in critical functions (like some specific parts of the halo2 proving system side) should we avoid having `assert`.

zkevm-circuits/src/super_circuit.rs
```diff
@@ -260,6 +260,12 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize, const MAX_RWS: u
             block.circuits_params.max_rws,
             Value::known(block.randomness),
         )?;
+        config.tx_table.load(
```
ed255, line 263: The TxTable is already loaded when calling `self.tx_circuit.assign`:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/3b06585106417a398a0af37422dc1fcadf5a614c/zkevm-circuits/src/tx_circuit.rs#L169

So there should be no need to load it again.

lispc: ok.. understand. so i think later we should unify the "tx table assignment in evm circuit" and "tx circuit assignment"..

ed255: In general the idea is that each table has a circuit that is responsible to verify its correctness, and the table is contained within the circuit.  The assign function of such circuit is also responsible of assigning/loading the table.
So:
- keccak circuit assigns the keccak table
- tx circuit assigns the tx table
- copy circuit assigns the copy table
- mpt circuit assigns the mpt table (not yet in main)
- bytecode circuit assigns the bytecode table
- exp circuit assigns the exp table
- state circuit assigns the rw table (this is currently not being done)
- block circuit assigns the block table (we don't have a block circuit yet)

Nevertheless we also want unit tests with single circuits, and for those we also need the tables loaded, for which we have the `Table.load` functions.

ed255 (approved):
LGTM!

PR #927, opened 2022-11-24
reduce super circuit degree to 9 to reduce extend_k from 4 to 3

lispc:
1. reduce state circuit degree to 9 
2. add a test to make sure super circuit degree <= 9
3. refactor super circuit a bit so later when we need to analyze the cost of each sub circuit (like inserting some `println!("after xx circuit, degree: {}", meta.degree())` ) , it will be easier.
4. also add a minor log change so offset in evm assignment log can be monotonous. The original debug log is redundant and a bit confusing:

image


CPerezz (approved):
LGTM!! Really nice job!!!

lispc:
Small pr. Merge it now

PR #937, opened 2022-11-25
Unify circuit types names and interfaces

ed255:
Given a circuit Foo:
- Unify the struct that contains the columns to FooCircuitConfig
- Unify the struct that contains the witness to FooCircuit
- Unify the config constructor to FooCircuitConfig::new, following the new SubCircuitConfig trait
- Unify the creation of a circuit from a block to FooCircuit::new_from_block following the new SubCircuit trait
- Unify the loading of fixed internal tables to load_aux_tables
- Unify the synthesis step of subcircuits to synthesize_sub following the new SubCircuit trait
- Remove generic associated const from circuit+config types as much as possible - Only possible in TxCircuit.  Not possible in PiCircuit due to halo2 Circuit trait
- Implement the Circuit trait to all subcircuits uniformly
- Move circuit tunning parameters to CircuitParams so that they can be taken from a single source
- Add an instance function to all circuits following the new SubCircuit trait


ed255:
Currently some testool unit tests are failing with the error `EthTypeError(Signature(InvalidSignature))`.  I've added the calculation of all the keccak inputs in the `witness::block_convert` (so that all circuits can consume the witness from a single source), which requires processing the transactions signatures.  I believe the failing tests in the testool have invalid signatures (which were not processed before), and that's why they are failing.

The rest of the tests should be passing.

Please if you have some time, take a look at the changes made and comment if you agree with this proposal or open a discussion to change!  Basically I've added this new traits and implemented them for all subcircuits:

```rust
/// SubCircuit is a circuit that performs the verification of a specific part of
/// the full Ethereum block verification.  The SubCircuit's interact with each
/// other via lookup tables and/or shared public inputs.  This type must contain
/// all the inputs required to synthesize this circuit (and the contained
/// table(s) if any).
pub trait SubCircuit<F: Field> {
    /// Configuration of the SubCircuit.
    type Config: SubCircuitConfig<F>;

    /// Create a new SubCircuit from a witness Block
    fn new_from_block(block: &witness::Block<F>) -> Self;

    /// Returns the instance columns required for this circuit.
    fn instance(&self) -> Vec<Vec<F>> {
        vec![]
    }
    /// Assign only the columns used by this sub-circuit.  This includes the
    /// columns that belong to the exposed lookup table contained within, if
    /// any; and excludes external tables that this sub-circuit does lookups
    /// to.
    fn synthesize_sub(
        &self,
        config: &Self::Config,
        challenges: &Challenges<Value<F>>,
    ) -> Result<(), Error>;
}

/// SubCircuit configuration
pub trait SubCircuitConfig<F: Field> {
    /// Config constructor arguments
    type ConfigArgs;

    /// Type constructor
    fn new(meta: &mut ConstraintSystem<F>, args: Self::ConfigArgs) -> Self;
}
```

zkevm-circuits/src/tx_circuit.rs
```diff
@@ -297,14 +313,60 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize>
...
-        )?;
+        )
+    }
+}
+
+impl<F: Field> SubCircuit<F> for TxCircuit<F> {
+    type Config = TxCircuitConfig<F>;
+
+    fn new_from_block(block: &witness::Block<F>) -> Self {
+        let mut rng = ChaCha20Rng::seed_from_u64(42);
```
pinkiebell, line 324: I think it would be nice to either change the `new_from_block` constructor to also take a Rng or put that into the witness struct.

zkevm-circuits/src/witness/block.rs
```diff
@@ -42,6 +46,12 @@ pub struct Block<F> {
     pub circuits_params: CircuitsParams,
     /// Inputs to the SHA3 opcode
     pub sha3_inputs: Vec<Vec<u8>>,
+    /// State root of the previous block
+    pub prev_state_root: Word, // TODO: Make this U256
```
pinkiebell, line 50: Isn't a Word = U256?

zkevm-circuits/Cargo.toml
```diff
@@ -34,6 +34,7 @@ integer =   { git = "https://github.com/privacy-scaling-explorations/halo2wrong"
 libsecp256k1 = "0.7"
 num-bigint = { version = "0.4" }
 subtle = "2.4"
+rand_chacha = "0.3"
```
pinkiebell, line 37: This is related regarding my common on Rng

zkevm-circuits/src/super_circuit.rs
```diff
@@ -248,74 +261,40 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize, const MAX_RWS: u
...
+        self.state_circuit
+            .synthesize_sub(&config.state_circuit, &challenges, &mut layouter)?;
+        self.copy_circuit
+            .synthesize_sub(&config.copy_circuit, &challenges, &mut layouter)?;
+        self.exp_circuit
+            .synthesize_sub(&config.exp_circuit, &challenges, &mut layouter)?;
+        self.evm_circuit
+            .synthesize_sub(&config.evm_circuit, &challenges, &mut layouter)?;
+        self.pi_circuit
+            .synthesize_sub(&config.pi_circuit, &challenges, &mut layouter)?;
```
pinkiebell, line 297: what about the keccak_circuit ?

pinkiebell (commented):
Looks great so far 💪 . This will simplify a lot and I hope we can get rid of const generics in the future too 😁

pinkiebell:
> Currently some testool unit tests are failing with the error EthTypeError(Signature(InvalidSignature))

Yes, I think the ethereum tests suite include invalid signatures too. So we must handle those somehow

zkevm-circuits/Cargo.toml
pinkiebell: At the moment, it also seems we just using it to mock stuff. If so, we can drop chacha completely and use `StdRng` instead.

zkevm-circuits/src/tx_circuit.rs
ed255: This rng is used to create an auxiliary point that is required for the ECDSA Chip.  This point is assigned as witness, and is not seen by the verifier.  I checked with @davidnevadoc and it's OK to reuse the same point through different instantiations of the circuit; that's why I moved the rng to this function instead of taking it externally.  But I understand that this looks weird, like a hardcoded randomness that should only be acceptable in test code.  I'll check if I can replace the generation of a random point with an rng, by adding a specific point directly.

zkevm-circuits/src/witness/block.rs
ed255: Ah, I actually meant `H256` :sweat_smile:

zkevm-circuits/src/super_circuit.rs
ed255: oops

ed255:
> Let's me try to check what's happening with the testool checks

I just found it!  The signatures of the transactions were not being propagated to the transaction types passed to the rest of the code.

pinkiebell:
> > Let's me try to check what's happening with the testool checks
> 
> I just found it! The signatures of the transactions were not being propagated to the transaction types passed to the rest of the code.

Great! Does that mean we don't test/support invalid signatures yet?

ed255:
> Great! Does that mean we don't test/support invalid signatures yet?

The circuit that checks the transaction signatures is the TxCircuit.  Currently it doesn't support invalid signatures.  For proof of validity, invalid signatures support should not be required; but for a rollup we may require it.  As an extra note, the signature verification uses the EcdsaChip from halow2rong, which currently doesn't support invalid signatures.

So strictly speaking, only tests that contain the TxCircuit need signed signatures; for the rest of the tests (other circuits) the txs signatures are not needed.  Nevertheless in this PR I added the calculation of all keccak inputs required to proof a full block as part of the circuit's input block preparation.  And part of the keccak inputs involves recovering the sender public key from the signature transaction (this is because the Tx Circuit proves that the signature of the tx was signed with a public key, that when hashed gives the sender address).  And this is the step that was giving errors.

For the testool this was easy to solve, because the transactions are signed, but in a conversion step the signature was not being propagated.

For the unit tests this is a bit more complicated, because we have a block builder used to generate test inputs where we only specify the tx sender as an address (which means that the block builder doesn't know the private key and txs are not signed).  I think in the future we could update the block builder to work with wallets, so that all generated transactions are signed; but that's a significant refactor.  So for now I decided to skip generating the keccak inputs for the tx circuit for any tx that is not signed (and do a warn about it).

zkevm-circuits/src/tx_circuit.rs
ed255: Actually it's not clear if the EcdsaChip is safe with this random point passed as witness.  David started investigating this https://github.com/privacy-scaling-explorations/halo2wrong/issues/53 
But in any case, either a hardcoded point can be used, or the EcdsaChip needs to be updated, but the `new_from_block` shouldn't be concerned with any randomness for now :)

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
```diff
@@ -688,6 +707,47 @@ fn into_words(message: &[u8]) -> Vec<u64> {
...
+    pub fn new(bytecodes: Vec<UnrolledBytecode<F>>, size: usize) -> Self {
+        BytecodeCircuit { bytecodes, size }
+    }
+}
+
+impl<F: Field> SubCircuit<F> for BytecodeCircuit<F> {
+    type Config = BytecodeCircuitConfig<F>;
+
+    fn new_from_block(block: &witness::Block<F>) -> Self {
+        let bytecode_size = block.circuits_params.max_bytecode + 128;
```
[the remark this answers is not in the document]
ed255: I checked this `+128` number and it's to account for the unusable rows that halo2 needs.  It's not a parameter that the consumer should choose, and it can be calculated from the circuit configuration. Currently it shows up here as a magic number.  I agree that this should be addressed, although I would prefer to do it in a future PR to avoid adding more stuff here!
https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/949

zkevm-circuits/Cargo.toml
ed255: Marking this as resolved as for now we're using a "constant" point generated with a deterministic rng.

zkevm-circuits/src/bytecode_circuit/bytecode_unroller.rs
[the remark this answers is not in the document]
pinkiebell: Sounds good to me 👍

zkevm-circuits/src/test_util.rs
```diff
@@ -47,6 +48,7 @@ pub fn run_test_circuits<const NACC: usize, const NTX: usize>(
     config: Option<BytecodeTestConfig>,
 ) -> Result<(), Vec<VerifyFailure>> {
     let block: GethData = test_ctx.into();
+    dbg!(&block);
```
pinkiebell, line 51: leftover?

pinkiebell (approved):

zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
```diff
@@ -777,8 +805,11 @@ mod transform_to {
     }
 }
 
-impl<F: Field> KeccakPackedConfig<F> {
-    pub(crate) fn configure(meta: &mut ConstraintSystem<F>, r: Expression<F>) -> Self {
+impl<F: Field> SubCircuitConfig<F> for KeccakCircuitConfig<F> {
+    type ConfigArgs = Expression<F>;
```
han0110, line 809: Perhaps we can also pass in `keccak_table` as argument? Then all tables will be allocated in the top level, which might be more consistent?

And we can also remove the necessity to [allocate a dummy advice](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/925/files#diff-6a80b984a7756f4e24712f75e42b12ea3be5af4111bdbcfcc08d1b18d33d22acR57-R58) beforehand in the test of `KeccakCircuit` with challenge API.

han0110 (approved):
LGTM! It's so clear now without those const generic in sub-circuits.

I think it'd be even better to move all `impl Circuit for *Circuit` into `mod test`, since they should never be used independently, then we never accidentally call `*Circuit::synthesize` (currently there are `ExpCircuit`, `StateCircuit`, `TxCircuit`, `BytecodeCircuit` and `KeccakCircuit` implemented not in `mod test`).

Edit: After chatted with Edu, since currently even `zkevm-chain` is using `zkevm-circuits` with feature `test`, it seems worth nothing to do this now, so this is moved to issue https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/953).

ed255:
> @ed255 Thanks that clears it up. Regarding the rollup, we can either recover the signatures or just do a cheaper pre-validation so that the ecdsa chip can recover a signature

Ah, sorry I didn't expand too much on the rollup story: In a rollup we may have a mechanism to include "forced" transactions to avoid censorship.  That would be a transaction that is sent to L1 and the node is forced to include in a block.  Since these transactions wouldn't be validated, they could contain an invalid signature that makes it impossible to recover the public key; but they must still be processed in the circuits.  So the EcdsaChip would need to be extended to prove that a tx signature is invalid (currently the constraints won't pass with an invalid signature).

zkevm-circuits/src/pi_circuit.rs
```diff
@@ -658,54 +687,116 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize>
...
+        let rlc_rpi = rlc_rpi_col
+            .iter()
+            .rev()
+            .fold(F::zero(), |acc, val| acc * self.rand_rpi + val);
+
+        // let block_hash = public_data
+        //     .eth_block
+        //     .hash
+        //     .unwrap_or_else(H256::zero)
+        //     .to_fixed_bytes();
```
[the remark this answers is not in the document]
ed255: Ah, I just moved this code around, this was already there: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/43da558aac7178f1895925782874662d26a72642/zkevm-circuits/src/pi_circuit.rs#L863-L867

I guess this will be uncommented once we validate the block hash.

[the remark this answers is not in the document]
han0110: I see, make sense to keep it.

zkevm-circuits/src/test_util.rs
ed255: yeah! thanks for catching this :)

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -67,19 +72,46 @@ pub struct CopyCircuit<F> {
...
 
-impl<F: Field> CopyCircuit<F> {
+/// Circuit configuration arguments
+pub struct CopyCircuitConfigArgs<F: Field> {
+    /// TxTable
+    pub tx_table: TxTable, // NOTE: Can this be TxTable?
+    /// RwTable
+    pub rw_table: RwTable, // NOTE: Can this be RwTable?
+    /// BytecodeTable
+    pub bytecode_table: BytecodeTable, // NOTE: Can this be BytecodeTable?
```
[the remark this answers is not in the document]
ed255: ah yeah! These were `dyn LookupTable`, and I already applied the change, so the notes are unnecessary

zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
ed255: thanks for the suggestion!

kunxian-xia:
> > @ed255 Thanks that clears it up. Regarding the rollup, we can either recover the signatures or just do a cheaper pre-validation so that the ecdsa chip can recover a signature
> 
> Ah, sorry I didn't expand too much on the rollup story: In a rollup we may have a mechanism to include "forced" transactions to avoid censorship. That would be a transaction that is sent to L1 and the node is forced to include in a block. Since these transactions wouldn't be validated, they could contain an invalid signature that makes it impossible to recover the public key; but they must still be processed in the circuits. So the EcdsaChip would need to be extended to prove that a tx signature is invalid (currently the constraints won't pass with an invalid signature).

Why is that the L1 "forced" transaction will include an invalid signature? Do you means that the L1 "forced" message might be triggered by a smart contract?

ed255:
> Why is that the L1 "forced" transaction will include an invalid signature? Do you mean that the L1 "forced" message might be triggered by a smart contract? @ed255 

Let me update on this.  My thought was that an "L1" forced transaction would be sent by any L1 address (that may not be the same as the `forcet_tx.from`), and the L1 contract would not do any kind of validation, just keep it to be included in a future L2 block.  This `forced_tx` would have a unverified signature, so the zkEVM would need to support having txs with invalid signatures that must be skipped.

But after a quick discussion with Haichen (scroll) and Brecth (taiko) I learned that this won't be necessary; because in both zkRollup projects they plan for the L1 contract to perform the signature verification and reject the `forced_tx` if the signature is invalid.  So we won't have the case of encountering a tx with invalid signature included in a block.

PR #939, opened 2022-11-26
add unit tests for stackOnlyOp opcode

ChengYueJia:
Add some stackOnlyOp unit tests.
see more on #947


CPerezz:
Hi @ChengYueJia is this ready for review? If not, could you leave as Draft until you have the PR ready?

Also, IIRC all of the opcodes that are already implemented as StackOpcode have tests associated. At least things like: 
41 | COINBASE
42 | TIMESTAMP
43 | NUMBER
46 | CHAINID 

These al have tests associated already. So could you be more specific and review excatly if all the mentioned opcodes require indeed testing?

ChengYueJia:
> Also, IIRC all of the opcodes that are already implemented as StackOpcode have tests associated. At least things like: 41 >| COINBASE 42 | TIMESTAMP 43 | NUMBER 46 | CHAINID
> These al have tests associated already. So could you be more specific and review excatly if all the mentioned opcodes require indeed testing?

@CPerezz  Thanks for your reminder. I should submit an issue before pr. 
I've submitted the issue #947 to specify the needed testing.

> Hi @ChengYueJia is this ready for review? If not, could you leave as Draft until you have the PR ready?
After specify the need testing, I'll finish it. Before all of that,  this will be left as a draft.

ChengYueJia:
@CPerezz  Hi, there are completed. please have a view.

bus-mapping/src/evm/opcodes/stackonlyop.rs
```diff
@@ -145,28 +168,237 @@ mod stackonlyop_tests {
...
             ],
-            vec![StackOp::new(1, StackAddress(1023), Word::from(0x79bdf))],
+            vec![StackOp::new(1, StackAddress(1023), Word::from(result))],
+        );
+    }
+    #[test]
+    fn test_xor_operate() {
+        xor_opcode_impl(0x01, 0x01, 0x01.bitxor(0x01) as usize);
+        xor_opcode_impl(0x01, 0x00, 0x00.bitxor(0x01) as usize);
+        xor_opcode_impl(0x00, 0x00, 0x00.bitxor(0x00) as usize);
```
CPerezz, line 336: Considering how small the test numbers are, we could simply put the result directly instead of using `bitxor()`. But it's fine if you want to keep it as is.

bus-mapping/src/evm/opcodes/stackonlyop.rs
```diff
@@ -145,28 +168,237 @@ mod stackonlyop_tests {
...
+    #[test]
+    fn difficulty_opcode_impl() {
+        stack_only_opcode_impl::<0, 1>(
+            OpcodeId::DIFFICULTY,
+            bytecode! {
+                DIFFICULTY
+                STOP
+            },
+            vec![],
+            vec![StackOp::new(1, StackAddress(1023), word!("0x200000"))],
```
CPerezz, line 371: Can you get the difficulty param from the testing constants instead of hardcoding it??

CPerezz (changes requested):
Nice work. Just a request regarding the testing parameters.

bus-mapping/src/evm/opcodes/stackonlyop.rs
ChengYueJia: This is a good idea. Thanks.

ChengYueJia:
I've added the mock var. Please have another look.

mock/src/lib.rs
```diff
@@ -21,8 +21,14 @@ lazy_static! {
         address!("0x00000000000000000000000000000000c014ba5e");
     /// Mock gasprice value
     pub static ref MOCK_GASPRICE: Word = Word::from(1u8);
+    /// Mock BASEFEE value
+    pub static ref MOCK_BASEFEE: Word = Word::zero();
```
CPerezz, line 25: I'd probably use a value that is different from 0. Otherwise some tests might pass when shouldn't. But we should keep the MockBlock::default() with 0 still.

CPerezz (commented):
Just a comment on the MOCK_BASEFEE. The rest looks good.

mock/src/lib.rs
ChengYueJia: > I'd probably use a value that is different from 0. Otherwise some tests might pass when shouldn't. But we should keep the MockBlock::default() with 0 still.

I've tried to make `MOCK_BASEFEE: Word = Word::from(1u64)`, but when I run this test. It shows error, like below.
image


Seems like, if we did this, MOCK_BASEFEE will not be used in the unit test for BASEFEE.

Or, let MOCK_BASEFEE andbase_fee_per_gas in  MockBlock::default be equal.

CPerezz: Is it possible that there are tests that need to update the BaseFee to the new value used??

ChengYueJia: For now, seems like no one wanna update it.

CPerezz: Feel free to leave it to zero then. I'll take a deeper look during next weeks.

CPerezz (approved):
LGTM

aguzmant103:
Hi @icemelon can you help us add 1 reviewer from your team? Thanks!

@ChengYueJia FYI needs to be rebased

silathdiir (approved):
LGTM

PR #979, opened 2022-12-12
fix dummy error handling 

DreamWuGit:
since many of error states are also in dummy way both buss mapping and circuit side , recent commit make some error states go to `fn_gen_error_state_associated_ops`  and use wrong dummy `gens_ops`. fix this by first checking `fn_gen_error_state_associated_ops` cover it , if not, use normal dummy way to handle it which contains the error info in exec step.


han0110 (approved):
Nice fix, so now all unimplemented exception states will have `handle_return` called.

Also IMO I think the logic here could be simplified to:
```rust
        // For exceptions that have been implemented
        if let Some(fn_gen_error_ops) = fn_gen_error_state_associated_ops(&exec_error) {
            return fn_gen_error_ops(state, geth_steps);
        } else {
            // For exceptions that already enter next call context, but fail immediately
            // (e.g. Depth, InsufficientBalance), we still need to parse the call.
            if geth_step.op.is_call_or_create() && !exec_step.oog_or_stack_error() {
                let call = state.parse_call(geth_step)?;
                state.push_call(call);
            // For exceptions that fail to enter next call context, we need
            // to restore call context of current caller
            } else {
            }
            state.handle_return(geth_step)?;
            return Ok(vec![exec_step]);
        }
```

I just tested locally and it works, but not sure if there is any edge case we haven't covered (I guess there are many...), so it'd be nice if you could help to check if it works and perhaps refactor in this way, since I found I need to understand what `exec_step.oog_or_stack_error() && !geth_step.op.is_call_or_create()` means every time I review tihs part.

DreamWuGit:
> Nice fix, so now all unimplemented exception states will have `handle_return` called.
> 
> Also IMO I think the logic here (line 306-324) could be simplified to:

should be correct, I will test it .

DreamWuGit:
> > Nice fix, so now all unimplemented exception states will have `handle_return` called.
> > Also IMO I think the logic here (line 306-324) could be simplified to:
> 
> should be correct, I will test it .

that's OK , updated in [7c78298](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/979/commits/7c7829864fdce294329353b3c053fccc2596e4b2). thanks!.

han0110:
> that's OK , updated in [7c78298](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/979/commits/7c7829864fdce294329353b3c053fccc2596e4b2). thanks!.

Thanks a lot!

lispc (approved):

PR #983, opened 2022-12-13
fix(evm): get_test_degree

lispc:


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -408,32 +409,56 @@ pub mod test {
...
+            EvmCircuit::<F>::get_num_rows_required(block);
+        let num_rows_required_for_rw_table: usize = block.circuits_params.max_rws;
+        let num_rows_required_for_fixed_table: usize = detect_fixed_table_tags(block)
+            .iter()
+            .map(|tag| tag.build::<F>().count())
+            .sum();
+        let num_rows_required_for_bytecode_table: usize = block
+            .bytecodes
+            .values()
+            .map(|bytecode| bytecode.bytes.len())
```
ed255, line 422: I noticed that this was previously incorrect, so maybe you could add the fix here: The bytecode circuit takes 1 extra row for each bytecode to encode its length, so this would be `.map(|bytecode| bytecode.bytes.len() + 1)`.

Note that this also means we shouldn't parametrize the Bytecode circuit with a `max_bytecode` parameter (I may be the culprit for that) because it's not accurate (due to the Length rows that don't contain any byte).

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -408,32 +409,56 @@ pub mod test {
...
+            .sum();
+        let num_rows_required_for_copy_table: usize =
+            block.copy_events.iter().map(|c| c.bytes.len() * 2).sum();
+        let num_rows_required_for_keccak_table: usize = block.keccak_inputs.len();
```
ed255, line 426: I think this one is not so straight forward.  `block.keccak_inputs` contains the list of keccak byte array inputs.  Each array is processed in chunks of 136 bytes (`const RATE` in keccak circuit).  For each chunk there are `NUM_ROUNDS + 1`, and each round takes `KECCAK_ROWS` (which can be accessed by `get_num_rows_per_round` in the keccak circuit module; it's 5 by default but can be changed via env var)

ed255 (commented):
This is great! :smiley:
Thanks for fixing this and making it more clear.  I've added two comments on the calculated number of rows for 2 circuits which I think need editing.

zkevm-circuits/src/evm_circuit.rs
lispc: fixed!

zkevm-circuits/src/evm_circuit.rs
lispc: that is true for the full keccak circuit. But for the table part, currently each byte input has only one row (RLC_input, RLC_hashoutput).    So this function is used to estimated rows for evm circuit, i think block.keccak_inputs.len() will be good.    When later we estimate keccak circuit rows in super circuit, then we should use `sum(KECCAK_ROWS * 25 * (Math.ceiling(bytes.len() / 136)) for bytes in block. keccak_inputs ]`

lispc:
@ed255 fixed bytecode table but not keccak table. Explained above

zkevm-circuits/src/evm_circuit.rs
ed255: Oh you're completely right!  My analysis doesn't apply in this EVM test context.

ed255 (approved):
LGTM!

han0110 (approved):
LGTM!

Issue #750, opened 2022-09-08
Refactor EVM circuit with challenge API


aguzmant103:

ed255:
Reference example: https://github.com/han0110/halo2/blob/feature%2Fchallenge-api/halo2_proofs/examples/shuffle.rs

aguzmant103:
Notes: update the abstraction done in the EVM circuit to support challenge API. Need to refactor the opcodes.

aguzmant103:
Notes: in progress. 

While checking the errors there's a poltergeist there. The problem how to assign cells no "advised" columns on phase 1 and phase 2. Can do it manually but t's a nightmare, not very maintainable. Find a way for a solution that is more maintainable. 

Discussed some time ago how to stop using the word RLC, we could remove it and split the words into HIGH and LOW. Not sure the current progress. If the challenge APIs changes for word we have to revert it.

aguzmant103:
Pending Discussion Notes:

- Remove RLC and change everything to HIGH and LOW?

Future question, if splitting the circuit to different proofs, removing the challenge API has it's benefits. 
The challenge API one 
- is for word but 
- also using it for Keccak inputs, and we have never discussed removing that.
- We also use the challenge API for lookups and haven't discussed removing that.
- We might also use it for batched ECDSA integration.

aguzmant103:
Notes: meeting with Adria, Carlos and David, addressing the problem on how to use the phase 1 and 2 cell manager. Still in progress.

aguzmant103:
Notes: making progress. Pending: when the degree of expression is too high, it automatically needs to split the expressions + other things.

aguzmant103:
Notes: in progress, it seems everything is working but having 2 pending topics:
- Correctness: need to find systematic way to search for problems
- Performance: 1st implementation working 3-4 times slower, main problem using call identifier from Halo2. Now is 50% slower. Want to continue to debug and optimize this

PR is in draft, could be reviewed.
PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/990

PR #990, opened 2022-12-15
Challenge in EVM circuit WIP

adria0:
Uploaded the version I'm working now of evm circuit with challenge api

Uses
https://github.com/privacy-scaling-explorations/halo2/tree/tmp/challenge-evm-circuit

Current status, and stuff I'll be working on
- [X] ~Not passing copy_circuit_valid_sha3~
- [X] ~Its like 50% slower for now (anyway, I reduced from 300% slower to 50% slower replacing Expression::identifier by https://github.com/privacy-scaling-explorations/halo2/blob/f090828bb829ac817d20b3446b3676bc63af8d3b/halo2_proofs/src/plonk/circuit.rs#L930 )~ seems that performance now is the same.
- [x] Lookup RLC challenge in Phase3 using `lookup_input` challenge
- [x] fix `split_expression`
- [x] Rebase

Anyway, if anybody wants to take a look (not asking for a detailed review now, only for general comments), will be nice because I'm blocking other issues.


han0110:
Had a review on the major changes and every opcode today, here are some my thoughts:

- Since we have 3 phases, the witness assignment is called 3 times, so the slow down seems to make sense. In future PR we will need to address this issue by not doing unnecessary computation when the phase is not matching.
- The lookup RLC challenge is still using the same challenge `evm_word`, I think if we need to use another challenge `lookup_input` usable after second phase, then we'd need more third phase columns in `ColumnType::Lookup` (or need more phase2 columns if we manage to remove phase3, see next bullet).
- I think we could remove third phase by witnessing the `is_zero` of RLC value in first phase, since as far as I see there is only `is_zero(word_rlc1 - word_rlc2)` requiring the third phase

han0110:
> But is this happening with `MockProver`? I added a `dbg!` of the challange values in the `EvmCircuit::synthesize` and I'm only getting one print, with the values set. Should I expect here at least two rounds, one with `Value::unknown` and the other with `Value::known` ?

Ahhh sorry, you are completely right, it should just be one synthesize called in `MockProver`, only when real prover we will do multiple times synthesize.

aguzmant103:
Closes #750

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1410,10 +1426,9 @@ impl<'a, F: Field> ConstraintBuilder<'a, F> {
...
             None => lookup,
         };
-
         let compressed_expr = self.split_expression(
             "Lookup compression",
-            rlc::expr(&lookup.input_exprs(), self.power_of_randomness),
+            rlc::expr(&lookup.input_exprs(), self.lookup_powers_of_randomness),
             MAX_DEGREE - IMPLICIT_DEGREE,
         );
         self.store_expression(name, compressed_expr, CellType::Lookup(lookup.table()));
```
han0110, line 1434: This seems need to be stored in `LookupPhase3` since it contains challenge after phase 2

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -395,7 +412,12 @@ impl<F: Field> ExecutionConfig<F> {
         });
 
         let mut stored_expressions_map = HashMap::new();
+
         let step_next = Step::new(meta, advices, MAX_STEP_HEIGHT, true);
+        let word_powers_of_randomness: [Expression<F>; 31] =
+            challenges.evm_word_powers_of_randomness();
+        let lookup_powers_of_randomness: [Expression<F>; 31] =
+            challenges.lookup_input_powers_of_randomness();
```
han0110, line 420: It seems we only need `12` here at most for `CopyTable`.

han0110 (commented):
> for  the`lookup_input` challenge,  you you think that's enough with https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/990/commits/8120bd4cdecfbfe94224562c24c157c9f5cdb8cb ?

It looks good to me! Just left 2 more thoughts.

PR #997, opened 2022-12-21
bugfix for pi-circuit-bench

SuccinctPaul:


han0110 (approved):
I can't see where the bug is, but using the built `DEGREE` seems to be a good idea.

After the branch is updated we can merge this PR.

circuit-benchmarks/src/pi_circuit.rs
```diff
@@ -47,17 +43,17 @@ mod tests {
         ));
         let public_inputs = circuit.0.instance();
         let instance: Vec<&[Fr]> = public_inputs.iter().map(|input| &input[..]).collect();
-        let instances = &[&instance[..]][..];
```
ChengYueJia, line 50: Hi, @han0110 , the bug is over here. A very small one.
And I've updated

circuit-benchmarks/src/pi_circuit.rs
```diff
@@ -47,17 +43,17 @@ mod tests {
...
+        let instances = &[&instance[..]];
```
ChengYueJia, line 46: Hi, @han0110  , the bug is over here. A very small one.
And I've updated the branch.

Issue #996, opened 2022-12-19
Soundness Bug: Modulo Gadget


kcharbo3:
The [Modulo gadget](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/9400c6237bd3efb9bc979b49fb0f25df7d763853/zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs) assigns `k` to `a / n`. However, `k` is not actually constrained to `a / n`. Any prover can substitute their own values for `k` as long as it is `<256 bits`.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/9400c6237bd3efb9bc979b49fb0f25df7d763853/zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs#L72

If the prover can submit their own values for `k` (< 256 bits), they can prove an inaccurate modulo operation. 

Take the following values for example:
`n = 3`
`k = 2^255`
`r = 0`
`a = 2^255`

The statement `0 = 2^255mod3` is false. But this statement will prove successfully in the circuit. This is because the MulAddWord gadget is done over the 256 bit field. So this is the constraint that is checked (which is true in this case):
`3 * 2^255 + 0 = 2^255 (mod 2^256)`. 

But I don't think this should suffice for proving `0 = 2^255mod3` (as 2^255 is not divisible by 3) So there is a whole set of values that will break this modulo operation if the prover is allowed to choose their own `k` value. To prevent this, another constraint needs to be added on `k` (ideally so that `k*n < 256 bits`).

In slightly more formal terms:
Let `k, n, r, a < 256 bits`.
The current constraint `kn + r = a mod2^256` implies there exists some integer `j` such that `kn + r = a + j2^256`.
So this can be rewritten as `(kn - j2^256) + r = a`. (Note that the MulAddWord gadget allows for products greater than 256 bits)
Since we are trying to prove `r=amodn`, then there must exist some integer `h` such that `hn + r = a`.
Therefore if `r=amodn` is true, then we must be able to write `(kn - j2^256)` as an integer multiple of `n`. So, `n` must divide `j * 2^256`.  But this requirement is never explicitly constrained in the circuit.
So the prover can manipulate `k` and `j` to satisfy `(kn - j2^256) + r = a` with a `j * 2^256` value that is not divisible by `n`. So in this case, `r=amodn` is false but the circuit has still "proved" it to be true.

ChihChengLiang:
[...]

PR #999, opened 2022-12-21
fix modulo soundness due to overflow

hero78119:
Fix issue #996 

Changes summary

`ModGadget` 
- exposing witness `k` so it's able to be unit-test. With k exposing, therefore also move witness generation logic outside.
- add test to cover soundness bug.

`MulModGadget`, op implementation
- rename `k` to `k2`,  plus adding `k1`, also passing `k1` for `ModGadget` to constrain.

`MulAddWordsGadget` 
- use `saturating_sub` to avoid underflow panic during carry calculation.
- add test to cover underflow

### DIFF
--- a/zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs
+++ b/zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs
@@ -195,15 +195,20 @@ mod tests {
     #[test]
     fn test_mod_n_unexpected_rem() {
         // test soundness by manipulating k to make a' = k * n + r and a' >=
-        // 2^256 lead to overflow and trigger soundness bug: (a' != a) ^ a' ≡ a
+        // 2^256 cause overflow and trigger soundness bug: (a' != a) ^ a' ≡ a
         // (mod 2^256)
+        // Here, the attacker tries to convince you of a false statement `2 % 3 = 0` by
+        // showing you `2 = ((2^256 + 2) / 3) * 3 + 0`. In the `MulAddWordsGadget`, `k *
+        // n + r = a  (modulo 2**256)` would have been a valid statement. But the gadget
+        // would have the overflow = 1. Since we constrain the overflow to be 0 in the
+        // ModGadget, the statement would be invalid in the ModGadget.
         try_test!(
             ModGadgetTestContainer<Fr>,
             vec![
                 Word::from(2),
                 Word::from(3),
                 Word::from(0),
-                /* magic number (2^256 + 2) / 3, and 2^256 + 2 divisible by 3 */
+                /* magic number (2^256 + 2) / 3, and 2^256 + 2 is divisible by 3 */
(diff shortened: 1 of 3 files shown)


zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs
```diff
@@ -49,27 +49,31 @@ impl<F: Field> ModGadget<F> {
             1.expr() - lt.expr() - n_is_zero.expr(),
         );
 
+        // Constrain k * n + r no overflow
+        cb.add_constraint("overflow == 0 for k * n + r", mul_add_words.overflow());
```
ChihChengLiang, line 53: What do you think use `require_zero` here for clarity?

ChihChengLiang (approved):
Great fix. LGTM. I added a nitpick.

zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs
```diff
@@ -183,6 +194,20 @@ mod tests {
...
+        // test soundness by manipulating k to make a' = k * n + r and a' >=
+        // 2^256 lead to overflow and trigger soundness bug: (a' != a) ^ a' ≡ a
+        // (mod 2^256)
+        try_test!(
+            ModGadgetTestContainer<Fr>,
+            vec![
+                Word::from(2),
+                Word::from(3),
+                Word::from(0),
+                /* magic number (2^256 + 2) / 3, and 2^256 + 2 divisible by 3 */
```
ChihChengLiang, line 206:
```suggestion
                /* magic number (2^256 + 2) / 3, and 2^256 + 2 is divisible by 3 */
```

zkevm-circuits/src/evm_circuit/util/math_gadget/modulo.rs
```diff
@@ -183,6 +194,20 @@ mod tests {
 
     #[test]
     fn test_mod_n_unexpected_rem() {
+        // test soundness by manipulating k to make a' = k * n + r and a' >=
+        // 2^256 lead to overflow and trigger soundness bug: (a' != a) ^ a' ≡ a
+        // (mod 2^256)
```
ChihChengLiang, line 199:
```suggestion
        // (mod 2^256)
        // Here, the attacker tries to convince you of a false statement `2 % 3 = 0` by showing you `2 = ((2^256 + 2) / 3) * 3 + 0`. In the `MulAddWordsGadget`, `k * n + r = a  (modulo 2**256)` would have been a valid statement. But the gadget would have the overflow=1. Since we constrain the overflow to be 0 in the ModGadget, the statement would be invalid in the ModGadget.
```

ChihChengLiang (approved):
Added some more comments to help understand the soundness bug.

KimiWu123 (approved):

Issue #1009, opened 2022-12-26
bug: `get_step_err` fail to parse  write protection error for call


DreamWuGit:
when doing #1001 write protection error,  found that for call with nonzero value, `get_step_err` fail to get it , because it come to line  https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L1038
`thread 'evm_circuit::execution::error_write_protection::test::test_write_protection' panicked at 'called `Result::unwrap()` on an `Err` value: UnexpectedExecStepError("call failure without return", Step { pc: 0x0024, op: CALL, gas: 39976, gas_cost: 19103, depth: 2, error: None, stack: [0x2a, 0x2, 0x0, 0x0, 0xa, 0xc8, 0xffffffffffffffffffffffffffffffffffffffff, 0x2710], storage: {} })', bus-mapping/src/evm/opcodes.rs:300:72`

DreamWuGit:
fixed by PR #1010 .

PR #1010, opened 2022-12-26
fix bug #1009: parse call write protection error 

DreamWuGit:
fix bug #1009  and add corresponding trace test for call .


han0110 (approved):
LGTM! Nice fix.

silathdiir (approved):
LGTM

PR #1015, opened 2022-12-28
fix log copyrwc when is_persistent = false

DreamWuGit:
when is_persistent false, current copy_rwc_inc should assign with zero.  also add false persistent cases.


han0110 (approved):
LGTM

silathdiir (approved):
LGTM

PR #1018, opened 2022-12-30
opt(assignment): speedup EndBlock padding

lispc:
in evm circuit, if padding is enabled, then a lot of EndBlock will be assigned with the expensive `CachedRegion` way.

In this PR, these same non-last EndBlock steps are only handled by CachedRegion once, and then are repeated until the last EndBlock. 

For a 1<<18 circuit, assignment time reduced from 28s to 4s in my laptop

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution.rs
+++ b/zkevm-circuits/src/evm_circuit/execution.rs
@@ -741,6 +741,43 @@ impl<F: Field> ExecutionConfig<F> {
         }
     }
 
+    /// Assign columns related to step counter
+    fn assign_q_step(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        height: usize,
+    ) -> Result<(), Error> {
+        for idx in 0..height {
+            let offset = offset + idx;
+            self.q_usable.enable(region, offset)?;
+            region.assign_advice(
+                || "step selector",
+                self.q_step,
+                offset,
+                || Value::known(if idx == 0 { F::one() } else { F::zero() }),
+            )?;
+            let value = if idx == 0 {
+                F::zero()
+            } else {
+                F::from((height - idx) as u64)
+            };
+            region.assign_advice(
+                || "step height",
+                self.num_rows_until_next_step,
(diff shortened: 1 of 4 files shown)


zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -84,20 +86,52 @@ impl<'r, 'b, F: FieldExt> CachedRegion<'r, 'b, F> {
...
+            width_start: advice_columns[0].index(),
             height_start,
+            advice_columns,
         }
     }
 
+    /// Repeatedly assign the locally cached witnesses.
+    /// This method can be used as a "quick" path for assignment for repeated
+    /// padding rows
+    pub fn replicate_assignment_for_range<A, AR>(
```
ed255, line 106: I would suggest the following small change to this function comment in order to make it more clear what it does:

Add to the comment indicating that this function replicates the assignment of 1 row at `height_start` (which must be already assigned via the `CachedRegion`) into a range of rows indicated by `offset_begin, offset_end`.

ed255 (approved):
LGTM! Thanks for working on this improvement!
I've added two comments with suggestions for improvements/clarity.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -774,46 +775,56 @@ impl<F: Field> ExecutionConfig<F> {
...
-                                evm_rows
-                            );
-                            return Err(Error::Synthesis);
-                        }
-                        if exact || evm_rows - (offset + cur_state_height) == 1 {
-                            block_step = end_block_last;
-                            no_next_step = true;
+                let no_padding = evm_rows == 0;
+
+                let assign_q_step =
```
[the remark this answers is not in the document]
lispc: done

zkevm-circuits/src/evm_circuit/util.rs
lispc: done

han0110 (approved):
LGTM! Nice work!

Issue #1041, opened 2023-01-08
bug: behavior inconsistency between evm and circuits for address range


lispc:
in evm, it is valid to pass U256 value larger than U160 as address in opcodes like extcodesize/extcodehash, only the last 20bytes is used and the first 12bytes are ignored.

in current circuit, this behavior is not described, resulting some valid traces not accepted.

lispc:
need something like this

silathdiir:
I guess it may be caused by the first 12-bytes of address are not zero. And may fail on `stack_pop(address.expr())` in `configure`. Will confirm, thanks.

lispc:
oh i forget paste my patch @silathdiir  

https://github.com/scroll-tech/zkevm-circuits/commit/fa36737d8c5bc5e465431229f98f61595cd9debb

PR #1042, opened 2023-01-09
fix stack value `address` to `Word` type in gadgets

silathdiir:
Close https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1041

Fixed to not split `Address` (calling `to_address` function) when stack ~~push or~~ pop in both bus-mapping and assign exec step of circuit. The first 12-bytes may be lost.

Referenced [code_address](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/callop.rs#L42) in `CallOpGadget`.

### DIFF
--- a/bus-mapping/src/evm/opcodes/balance.rs
+++ b/bus-mapping/src/evm/opcodes/balance.rs
@@ -2,7 +2,7 @@ use crate::circuit_input_builder::{CircuitInputStateRef, ExecStep};
 use crate::evm::Opcode;
 use crate::operation::{AccountField, CallContextField, TxAccessListAccountOp, RW};
 use crate::Error;
-use eth_types::{GethExecStep, ToAddress, ToWord, Word, U256};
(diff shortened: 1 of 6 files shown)


lispc (approved):
let's wait another review

lispc:
oh i think you don't need to modify caller and address. sine for these 2 opcodes, address is output instead of input, it it must within U160. For extcode{size,hash},balance, address is input, it can be larger than U160.

silathdiir:
> oh i think you don't need to modify caller and address. sine for these 2 opcodes, address is output instead of input, it it must within U160. For extcode{size,hash},balance, address is input, it can be larger than U160.
> 
> Using 20bytes rather than 32bytes costs less cells.

Reverted the fixes in `ADDRESS` and `CALLER`. Thanks.

lispc:
we forgot extcodecopy..

PR #1043, opened 2023-01-09
feat(testool): add new implemented opcodes

adria0:
Some opcodes are implemented but we are not checking them, added. Also removed EVM error pre-checks in order to get more accurate tests results.

At this moment, there are unimplemented stuff that generates a `warning` instead to create an `unimplemented` error, I will propose a PR to conditionally compile the kind or error that is triggered in these unimplemented stuff in order to not break the current testnets.


ed255 (approved):
LGTM!

Issue #1026, opened 2023-01-05
Implement bytecode refactor spec in circuit


leolara:
We have done the following refactor to the spec:

https://github.com/privacy-scaling-explorations/zkevm-specs/issues/151

https://github.com/privacy-scaling-explorations/zkevm-specs/pull/340

https://github.com/privacy-scaling-explorations/zkevm-specs/pull/351

This task is to implement this new spec into the bytecode circuit implementation in this repo.

aguzmant103:
- Improving readability by expressing it as a state machine
- Changing constraints in the bytecode circuit and adjustments on some columns of the witness generation
- This will include before and after benchmarks just to make sure we're not regressing in performance
- Using RLC (prev code was already using it too)

PR #1044, opened 2023-01-09
Implement bytecode refactor spec in circuit

leolara:
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1026

---

We have done the following refactor to the spec:

https://github.com/privacy-scaling-explorations/zkevm-specs/issues/151

https://github.com/privacy-scaling-explorations/zkevm-specs/pull/340

https://github.com/privacy-scaling-explorations/zkevm-specs/pull/351

This task is to implement this new spec into the bytecode circuit implementation in this repo.


zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1041 @@
...
+            }
+
+            // TODO: why different code_hash for each row? Is this going to produce the same
+            // result for every row?
+            let code_hash = challenges.evm_word().map(|challenge| {
+                RandomLinearCombination::<F, 32>::random_linear_combine(
+                    row.code_hash.to_le_bytes(),
+                    challenge,
+                )
+            });
```
han0110, line 508: If we want we I think we could group rows of the same bytecode inside `UnrolledBytecode` into a same struct, then we can cache this value.

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1041 @@
...
+        bytecode_circuit::{bytecode_unroller::BytecodeRow, dev::test_bytecode_circuit_unrolled},
+        util::{is_push, keccak},
+    };
+    use bus_mapping::evm::OpcodeId;
+    use eth_types::{Bytecode, Word};
+    use halo2_proofs::halo2curves::bn256::Fr;
+
+    fn get_randomness<F: Field>() -> F {
+        F::from(123456)
+    }
```
han0110, line 797: Seems we could also remove this.

```suggestion
```

han0110 (approved):
LGMT! Nice refactor! Only left some unimportant comments.

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1041 @@
...
+            );
+
+            cb.require_equal(
+                "next.push_data_left == cur.is_code ? cur.push_data_size : cur.push_data_left - 1",
+                meta.query_advice(push_data_left, Rotation::next()),
+                select::expr(
+                    meta.query_advice(bytecode_table.is_code, Rotation::cur()),
+                    meta.query_advice(push_data_size, Rotation::cur()),
+                    meta.query_advice(push_data_left, Rotation::cur()) - 1.expr(),
+                ),
```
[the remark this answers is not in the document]
leolara: I think a comment could explain the optimisation in case that readability is poorer

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1041 @@
...
+                meta.query_advice(bytecode_table.index, Rotation::cur()) + 1.expr(),
+            );
+
+            cb.require_equal(
+                "next.hash == cur.hash",
+                meta.query_advice(bytecode_table.code_hash, Rotation::next()),
+                meta.query_advice(bytecode_table.code_hash, Rotation::cur()),
+            );
+
+            // TODO: check this
```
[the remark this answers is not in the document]
leolara: I just was insecure about my understanding of the challenge, but if you have seen it I will remove the comment.

zkevm-circuits/src/bytecode_circuit/circuit.rs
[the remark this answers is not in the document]
leolara: @han0110 actually when is_code is 1, we don't substract 1, and when is_code is 0 we do substract 1. hence, this would be equivalent to push_data_size - not(is_code)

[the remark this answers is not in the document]
han0110: Ah you are right!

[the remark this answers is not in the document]
leolara: actually, this does not work, because in one case is push_data_SIZE when in the other it is push_data_LEFT - 1

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1041 @@
...
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn assign_bytecode(
+        &self,
+        region: &mut Region<'_, F>,
+        bytecode: &UnrolledBytecode<F>,
+        challenges: &Challenges<Value<F>>,
+        push_rindex_is_zero_chip: &IsZeroChip<F>,
```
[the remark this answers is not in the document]
leolara: Done

zkevm-circuits/src/bytecode_circuit/circuit.rs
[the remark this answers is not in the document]
han0110: Yes... Sorry I didn't check the formula carefully.

leolara:
please @icemelon assign a reviewer

zkevm-circuits/src/bytecode_circuit/circuit.rs
leolara: This method does a bytecode as a time, so I think it can be solved just moving the expression outside the loop and using the code_hash from the first row of the bytecode.

However, this make the tests bytecode_invalid_hash_data don't pass as the code_hash of rows different from the header row are ignored by the witness generation. Hence, I need to modify this tests, but in the future we should inject an evil witness to test other constrains.

leolara: Done

icemelon:
I assigned myself and will take a look this week.

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1030 @@
...
+
+            let empty_hash = RandomLinearCombination::<F, 32>::random_linear_combine_expr(
+                EMPTY_HASH_LE.map(|v| Expression::Constant(F::from(v as u64))),
+                &challenges.evm_word_powers_of_randomness::<32>(),
+            );
+
+            cb.require_equal(
+                "assert cur.hash == EMPTY_HASH",
+                meta.query_advice(bytecode_table.code_hash, Rotation::cur()),
+                empty_hash,
```
ChihChengLiang, line 226: I remember we have `cb.empty_hash_rlc()`, but that method was available for `ConstraintBuilder` but not for `BaseConstraintBuilder`.
No change is needed here.

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1030 @@
...
+            );
+
+            cb.gate(and::expr(vec![
+                meta.query_fixed(q_enable, Rotation::cur()),
+                not::expr(meta.query_fixed(q_last, Rotation::cur())),
+                is_byte_to_header(meta),
+            ]))
+        });
+        meta.lookup_any(
+            "keccak256_table_lookup(cur.hash, cur.length, cur.value_rlc)",
```
ChihChengLiang, line 357:
```suggestion
            "keccak256_table_lookup(cur.value_rlc, cur.length, cur.hash)",
```

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -0,0 +1,1030 @@
...
+                ]);
+
+                let lookup_columns = vec![value_rlc, length, bytecode_table.code_hash];
+
+                let mut constraints = vec![(
+                    enable.clone(),
+                    meta.query_advice(keccak_table.is_enabled, Rotation::cur()),
+                )];
+
+                // TODO: perhaps write this explicitly so it is more readable the matching
```
ChihChengLiang, line 372: Agree we should make this more readable. One idea I have in mind is to have a function in keccak table side.
`match_columns(input_rlc, length, output_hash) -> Vec<(Column, Column)>`
The function is only responsible for yielding correct pairs of columns. We then add selectors and query cells here.

ChihChengLiang (approved):
LGTM. I left some comments, but they might be out of the scope of this PR. Feel free to ignore them.

zkevm-circuits/src/bytecode_circuit/circuit.rs
leolara: I was going to complain, but I actually think what you are saying makes sense.

PR #1046, opened 2023-01-09
feat(state): constrain non-exist proof for account and storage

ed255:
Add BatchedIsZeroChip in gadgets
Update the following opcode gadgets to handle non-existing accounts via code_hash == 0:
- BALANCE
- EXTCODEHASH
- EXTCODESIZE
- callop (CALLCODE, DELEGATECALL, STATICCALL, CALL)

Missing:
- EXTCODECOPY (the circuit is currently not implemented, I'll update it once https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720 is merged)

Important note: By handling the non-existence proofs from the state circuit (which requires detecting the special case of storage 0->0 or account,code_hash 0->0 to swap the proof type to non existence in the MPT lookup), the state circuit degree has increased from 9 to 12.  I could keep it at 9 by introducing an intermediate column and splitting the degree >9 expression.  Is it a problem to keep the degree at 12?  Should I keep it at 9 in an update to this PR?

Update: I've brought back the degree to 9, at the cost of introducing a new column in the State circuit (to hold an intermediary witness and splitting the highest degree expression).

Corresponding specs PR https://github.com/privacy-scaling-explorations/zkevm-specs/pull/354

### DIFF
--- a/bus-mapping/src/evm/opcodes/balance.rs
+++ b/bus-mapping/src/evm/opcodes/balance.rs
@@ -100,6 +100,7 @@ mod balance_tests {
     use eth_types::evm_types::{OpcodeId, StackAddress};
     use eth_types::geth_types::GethData;
     use eth_types::{address, bytecode, Bytecode, ToWord, Word, U256};
+    use keccak256::EMPTY_HASH_LE;
     use mock::TestContext;
     use pretty_assertions::assert_eq;
 
@@ -248,23 +249,33 @@ mod balance_tests {
             }
         );
 
+        let code_hash = Word::from_little_endian(&*EMPTY_HASH_LE);
         let operation = &container.account[indices[5].as_usize()];
         assert_eq!(operation.rw(), RW::READ);
         assert_eq!(
             operation.op(),
             &AccountOp {
                 address,
-                field: if exists {
-                    AccountField::Balance
-                } else {
-                    AccountField::NonExisting
-                },
-                value: balance,
-                value_prev: balance,
+                field: AccountField::CodeHash,
+                value: if exists { code_hash } else { U256::zero() },
+                value_prev: if exists { code_hash } else { U256::zero() },
             }
         );
+        if exists {
+            let operation = &container.account[indices[6].as_usize()];
+            assert_eq!(operation.rw(), RW::READ);
+            assert_eq!(
+                operation.op(),
+                &AccountOp {
+                    address,
+                    field: AccountField::Balance,
+                    value: balance,
+                    value_prev: balance,
+                }
+            );
+        }
 
-        let operation = &container.stack[indices[6].as_usize()];
+        let operation = &container.stack[indices[6 + if exists { 1 } else { 0 }].as_usize()];
         assert_eq!(operation.rw(), RW::WRITE);
         assert_eq!(
             operation.op(),
--- a/bus-mapping/src/evm/opcodes/balance.rs
+++ b/bus-mapping/src/evm/opcodes/balance.rs
@@ -2,7 +2,7 @@ use crate::circuit_input_builder::{CircuitInputStateRef, ExecStep};
 use crate::evm::Opcode;
 use crate::operation::{AccountField, CallContextField, TxAccessListAccountOp, RW};
 use crate::Error;
-use eth_types::{GethExecStep, ToAddress, Word, U256};
+use eth_types::{GethExecStep, ToAddress, ToWord, H256, U256};
 
 #[derive(Debug, Copy, Clone)]
 pub(crate) struct Balance;
@@ -57,23 +57,28 @@ impl Opcode for Balance {
         // Read account balance.
         let account = state.sdb.get_account(&address).1;
         let exists = !account.is_empty();
+        let balance = account.balance;
+        let code_hash = if exists {
+            account.code_hash
+        } else {
+            H256::zero()
+        };
+        state.account_read(
+            &mut exec_step,
+            address,
+            AccountField::CodeHash,
+            code_hash.to_word(),
+            code_hash.to_word(),
+        )?;
         if exists {
             state.account_read(
                 &mut exec_step,
                 address,
                 AccountField::Balance,
-                account.balance,
-                account.balance,
+                balance,
+                balance,
             )?;
-        } else {
-            state.account_read(
-                &mut exec_step,
-                address,
-                AccountField::NonExisting,
-                Word::zero(),
-                Word::zero(),
-            )?;
-        };
+        }
 
         // Write the BALANCE result to stack.
         state.stack_write(
--- a/bus-mapping/src/evm/opcodes/extcodehash.rs
+++ b/bus-mapping/src/evm/opcodes/extcodehash.rs
@@ -3,10 +3,9 @@ use crate::{
     circuit_input_builder::CircuitInputStateRef,
     evm::opcodes::ExecStep,
     operation::{AccountField, CallContextField, TxAccessListAccountOp, RW},
-    state_db::Account,
     Error,
 };
-use eth_types::{GethExecStep, ToAddress, ToWord, U256};
+use eth_types::{GethExecStep, ToAddress, ToWord, H256, U256};
 
 #[derive(Debug, Copy, Clone)]
 pub(crate) struct Extcodehash;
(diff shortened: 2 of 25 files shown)


zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -23,7 +24,8 @@ use rand::SeedableRng;
 use std::collections::{BTreeSet, HashMap};
 use strum::IntoEnumIterator;
 
-const N_ROWS: usize = 1 << 16;
+const N_ROWS: usize = 1 << 16; // TODO: Uncomment
+                               // const N_ROWS: usize = 1 << 8;
```
han0110, line 28: leftover?

zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -98,6 +106,7 @@ fn test_state_circuit_ok(
     let prover = MockProver::<Fr>::run(19, &circuit, power_of_randomness).unwrap();
     let verify_result = prover.verify();
     assert_eq!(verify_result, Ok(()));
+    // prover.assert_satisfied();
```
han0110, line 109:
```suggestion
```

han0110 (approved):
All LGTM! Nice work! Only left some nitpicks.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -460,15 +460,15 @@ pub mod test {
...
-        ]);
+        // dbg!([
+        //     num_rows_required_for_rw_table,
+        //     num_rows_required_for_fixed_table,
+        //     num_rows_required_for_bytecode_table,
+        //     num_rows_required_for_copy_table,
+        //     num_rows_required_for_keccak_table,
+        //     num_rows_required_for_tx_table,
+        //     num_rows_required_for_exp_table
+        // ]);
```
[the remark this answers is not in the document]
ed255: Ah I didn't meant to leave this commented, I just wanted to do it locally to avoid too much noise during tests.  I've updated it to `log::debug` following your suggestion as I think it's cleaner :)  Done in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/f59933089f977ec564f751c7051a824c0d5dd041

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -183,8 +181,8 @@ mod test {
         });
 
         test_root_ok(&account, false);
-        test_internal_ok(0x20, 0x00, &account, false);
-        test_internal_ok(0x1010, 0xff, &account, false);
+        // test_internal_ok(0x20, 0x00, &account, false);
+        // test_internal_ok(0x1010, 0xff, &account, false);
```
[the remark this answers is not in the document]
ed255: Ah yes! Resolved in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/f59933089f977ec564f751c7051a824c0d5dd041

zkevm-circuits/src/state_circuit/test.rs
ed255: yes :P Resolved in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/f59933089f977ec564f751c7051a824c0d5dd041

zkevm-circuits/src/state_circuit/test.rs
ed255: Resolved in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/f59933089f977ec564f751c7051a824c0d5dd041

ed255:
@icemelon @z2trillion could someone from scroll do a second review to this PR? Thanks!

bus-mapping/src/evm/opcodes/balance.rs
```diff
@@ -57,23 +57,28 @@ impl Opcode for Balance {
         // Read account balance.
         let account = state.sdb.get_account(&address).1;
```
z2trillion, line 58: You can just do `let (exists, account) = state.sdb.get_account(&address);`

bus-mapping/src/evm/opcodes/extcodehash.rs
```diff
@@ -55,30 +54,13 @@ impl Opcode for Extcodehash {
...
-
-        state.account_read(
-            &mut exec_step,
-            external_address,
-            AccountField::Balance,
-            balance,
-            balance,
-        )?;
-
+        let account = state.sdb.get_account(&external_address).1;
```
z2trillion, line 57: `let (exists, account) = ...`

bus-mapping/src/evm/opcodes/extcodesize.rs
```diff
@@ -51,25 +51,22 @@ impl Opcode for Extcodesize {
...
-            let code_hash = account.code_hash;
-            let code_hash_word = code_hash.to_word();
-            state.account_read(
-                &mut exec_step,
-                address,
-                AccountField::CodeHash,
-                code_hash_word,
-                code_hash_word,
-            )?;
+        let exists = !account.is_empty();
```
z2trillion, line 54: let (exists, account) = ...

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -271,6 +265,10 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
                 cb.power_of_randomness(),
             ),
         );
+        // is_empty_code_hash_expr is true when the account exists and has empty
+        // code hash, or when the account doesn't exist (which we encode with
+        // code_hash = 0).
+        let is_empty_code_hash_expr = is_empty_code_hash.expr() + callee_not_exists.expr();
```
z2trillion, line 271: I think a better name would be `no_callee_code`. `is_empty_code_hash_expr` looks like it should be just `is_empty_code_hash.expr()`.

zkevm-circuits/src/evm_circuit/execution/extcodesize.rs
```diff
@@ -25,10 +25,9 @@ pub(crate) struct ExtcodesizeGadget<F> {
     reversion_info: ReversionInfo<F>,
     tx_id: Cell<F>,
     is_warm: Cell<F>,
-    exists: Cell<F>,
     code_hash: Cell<F>,
-    code_size: Cell<F>,
-    code_size_bytes: [Cell<F>; 8],
+    not_exists: IsZeroGadget<F>,
+    code_size: RandomLinearCombination<F, 8>,
```
z2trillion, line 30: 8 -> N_BYTES_U64

zkevm-circuits/src/state_circuit.rs
```diff
@@ -56,9 +59,9 @@ pub struct StateCircuitConfig<F> {
     // For Rw::AccountStorage, identify non-existing if both committed value and
     // new value are zero. Will do lookup for ProofType::StorageDoesNotExist if
     // non-existing, otherwise do lookup for ProofType::StorageChanged.
-    // TODO: use BatchedIsZeroGadget here, once it doesn't depend on the evm circuit constraint
-    // builder.
-    is_non_exist: Column<Advice>,
+    is_non_exist: BatchedIsZeroConfig,
+    // Intermediarey witness used to reduce mpt lookup expression degree
```
z2trillion, line 63: Intermediarey -> Intermediary

z2trillion (changes requested):
Only had some minor changes requested. Overall this looks great.

zkevm-circuits/src/evm_circuit/execution/callop.rs
ed255: Thanks for the suggestion!  Applied in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/8c84bc31ed1ffc74c31508b5818e90d8762dd450

zkevm-circuits/src/evm_circuit/execution/extcodesize.rs
ed255: Tanks!  Applied in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/8c84bc31ed1ffc74c31508b5818e90d8762dd450

zkevm-circuits/src/state_circuit.rs
ed255: Tanks!  Applied in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046/commits/8c84bc31ed1ffc74c31508b5818e90d8762dd450

bus-mapping/src/evm/opcodes/balance.rs
ed255: I have tested this option and unfortunately it's not viable without further changes:
- Currently the flag returned by `get_account` is true when the account was found in the StateDB, but the account may be empty
- In the CircuitInputBuilder we're currently loading a bunch of empty accounts in the StateDB, so an empty account is found in the StateDB
- We're using the `found` flag as a check to make sure that all accounts that are touched during a block are preloaded into the StateDB (even if the account is empty).  This allows us to detect if we missed retrieving some account from geth that will be used in a block.  See all the cases where `Err(Error::AccountNotFound(address));` is returned.
  - If the flag `found` is replaced to mean `exists`, then we lose this check, and we will not be able to detect the case where we are dealing with an account that we didn't retrieve (so it's empty but shouldn't be).

So I think it's better to leave the flag returned by `get_account` to mean `found` (as in, the account is preloaded in StateDB even if it's empty), and use the `account.is_empty()` to really determine if the account exists or not.

z2trillion: got it. it is unfortunate that it's possible to for the db find accounts that the circuits will later claim don't exist. i'll look into whether it's possible to fix this naming issue.

z2trillion (approved):
thanks!

PR #1057, opened 2023-01-12
Refactor get_row_nums of evm circuit

### LINKED ISSUE #1039: Refactor get_required_num_rows  of EvmCircuit
For now, we have to do `EvmCircuit::<F>::configure` first before genrating witness data to get_required_num_rows.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/432b4ef9b2b368608b4f1283dce07951b0dd7d66/zkevm-circuits/src/evm_circuit.rs#L385-L389

Can we use a constant(`sum(executeState.hight`) as the required_num_rows of EvmCircuit ? Or other better way to do it?

The detail step info as below. https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/718#issue-1354538334

ChengYueJia:
Cause the height of step is immutable. So we can regard is as a param.

There are two way to do it.
1. Specify the heigh for each state. As below
    * Pros: easy and fast to view the explicit info and use.
    * Cons: Had to maintain it when we modify the StateGadget.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/8b890792d075c7c448206e9d074910014e7ff512/zkevm-circuits/src/evm_circuit/step.rs#L318-L321

2. Lazy load the Evm::configure to get the height map.
    * Pros: Had to maintain it when we modify the StateGadget.
    * Cons: First time to use is slow.

Finally, I choose the latter one for the maintain is much more expensive than lazy_load. 

Fix #1039


ChengYueJia:
Hi, @ed255 @lispc , could you have a view?

ed255 (approved):
LGTM!
I agree that the option 2 (doing the lazy_static hash_map with the results) is the correct way to go!
Nice work!

aguzmant103:
@icemelon can you help us find another reviewer? Thanks!

lispc (approved):

PR #1063, opened 2023-01-13
Remove dup `declare_tests!` macros

### LINKED ISSUE #968: Dedup code in the integration tests macros (declare_test)

davidnevadoc:
Closes #968

### DIFF
--- a/integration-tests/tests/circuits.rs
+++ b/integration-tests/tests/circuits.rs
@@ -1,39 +1,39 @@
 macro_rules! declare_tests {
-    (($name:ident, $block_tag:expr),$mock:expr) => {
+    (($name:ident, $block_tag:expr),$real_prover:expr) => {
         paste! {
             #[tokio::test]
             async fn [<serial_test_evm_ $name>]() {
                 log_init();
                 let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
-                test_evm_circuit_block(*block_num, $mock).await;
+                test_evm_circuit_block(*block_num, $real_prover).await;
             }
 
             #[tokio::test]
             async fn [<serial_test_state_ $name>]() {
                 log_init();
                 let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
-                test_state_circuit_block(*block_num, $mock).await;
+                test_state_circuit_block(*block_num, $real_prover).await;
             }
 
             #[tokio::test]
             async fn [<serial_test_tx_ $name>]() {
                 log_init();
                 let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
-                test_tx_circuit_block(*block_num, $mock).await;
+                test_tx_circuit_block(*block_num, $real_prover).await;
             }
 
             #[tokio::test]
             async fn [<serial_test_bytecode_ $name>]() {
                 log_init();
                 let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
-                test_bytecode_circuit_block(*block_num, $mock).await;
+                test_bytecode_circuit_block(*block_num, $real_prover).await;
             }
 
             #[tokio::test]
             async fn [<serial_test_copy_ $name>]() {
                 log_init();
                 let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
-                test_copy_circuit_block(*block_num, $mock).await;
+                test_copy_circuit_block(*block_num, $real_prover).await;
             }
         }
     };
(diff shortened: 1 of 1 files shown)


integration-tests/tests/circuits.rs
```diff
@@ -7,64 +7,91 @@ use integration_tests::log_init;
...
-                test_copy_circuit_block(*block_num, true).await;
+                test_copy_circuit_block(*block_num, $mock).await;
             }
         }
     };
 }
 
-declare_tests!(circuit_block_transfer_0, "Transfer 0");
+// Mock tests
+declare_tests!(circuit_block_transfer_0_mock, "Transfer 0", true);
```
ed255, line 51: I see that now the test functions names look like `foo` and `foo_mock` for each case. This will not be very friendly towards selecting only one set of tests.  Before this was easy because we had two features to select one set or the other: `circuits_actual` and `circuits_mock`, but also the test functions were in different modules (where as now they are all in the same module).

So I would suggest to keep the test functions in two separate modules so it's easy to test each set separately.  Also, currently there's 2 macro calls for each test block: I think this can be improved!

So here are my suggestions:
1. Remove the `#![cfg(feature = "circuits_actual")]` at the top because I think now it doesn't make sense.
2. Group the tests into two modules: `mock_prover` and `real_prover`
3. Add another macro so that we don't need to write twice the code for each block.  I think this could be achieved with 1 new macro that takes a list of `$name, $block_tag` and unrolls them two times: one inside `mod mock_prover {}` and the ohter in `mod real_prover {}`.  And in each module, have calls to `declare_tests!` each input in the list (in the first mod with true, the other one with false).  The unrolling can follow this example https://users.rust-lang.org/t/unroll-macro-argument-list/10510/2

ed255:
The current `run.sh` script used to run the integration tests expects that there exist for each set of tests a test file and a feature with the same name.  With this change we now have a single file `circuits.rs` that contains mock and real prover tests, and we only want to allow selecting only one of them.  Here's a possible solution:
```diff
diff --git a/integration-tests/run.sh b/integration-tests/run.sh
index 2d4f551f..8125210c 100755
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -3,7 +3,7 @@ set -e
 
 ARG_DEFAULT_SUDO=
 ARG_DEFAULT_STEPS="setup gendata tests cleanup"
-ARG_DEFAULT_TESTS="rpc circuit_input_builder circuits_mock"
+ARG_DEFAULT_TESTS="rpc circuit_input_builder circuits::mock_prover"
 
 usage() {
     cat >&2 << EOF
@@ -98,7 +98,7 @@ fi
 if [ -n "$STEP_TESTS" ]; then
     for testname in $ARG_TESTS; do
         echo "+ Running test group $testname"
-        cargo test --profile release --test $testname --features $testname -- --nocapture
+        cargo test --profile release --test $(echo $testname | sed -e 's/::/ /g') --all-features -- --nocapture
     done
 fi
```
It abuses the fact that in the script, adding a space between a string creates two arguments for the command.  So for the `circuits::mock_prover` the command will become: `cargo test --profile release --test circuits mock_prover --all-features -- --nocapture` which means run the test file `circuits.rs` and filter by pattern `mock_prover` (which will filter the `mock_prover` module).

ed255:
Ah, regarding the last error in the github actions, the workflow needs to be updated like this:
```diff
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -77,5 +77,5 @@ jobs:
       - run: ./run.sh --steps "gendata"
       - run: ./run.sh --steps "tests" --tests "rpc"
       - run: ./run.sh --steps "tests" --tests "circuit_input_builder"
-      - run: RUST_TEST_THREADS=1 ./run.sh --steps "tests" --tests "circuits_mock"
+      - run: RUST_TEST_THREADS=1 ./run.sh --steps "tests" --tests "circuits::mock_prover"
       - run: ./run.sh --steps "cleanup"
```
To follow the new test naming

integration-tests/tests/circuits.rs
```diff
@@ -0,0 +1,85 @@
...
+            async fn [<serial_test_copy_ $name>]() {
+                log_init();
+                let block_num = GEN_DATA.blocks.get($block_tag).unwrap();
+                test_copy_circuit_block(*block_num, $mock).await;
+            }
+        }
+    };
+}
+
+macro_rules! unroll_tests {
```
leolara, line 42: @davidnevadoc  I don't understand the use of the word "unroll" here, is it something that means something in the context of Rust macros?

ed255: Yes, the meaning of unroll here is related to Rust macros.  See this thread https://users.rust-lang.org/t/unroll-macro-argument-list/10510/2

Basically in the macro call you pass a list of values, and then the unroll generates some code for each value in the list.

ed255 (approved):
LGTM!

Issue #866, opened 2022-10-31
soundness problem of advice lookup


lispc:
here is an example of soundness bug

https://github.com/scroll-tech/zkevm-circuits/pull/212

let me explain, usually we use advice columns as lookup table, but it is difficult if not impossible to constrain that there is no evil row after the region where selectors are applied.

In the above example, all state circuit constraints are gated by a selector, but if we usually the selector be true on range [0, rws.len()) in debug environment, and [0, state_circuit_pad_to/max_rws] in prod environment ,and the max_rws currently is 1<<k-64.   So there are some rows where there are no constraints at all in this lookup table...

Two solutions for this type of problem:

(1) the selector must be exact same len with 1<<k-blinding_rows or same with l_active.
(2) don't allow "all lookup table are adivce" design pattern. At least one column should be fixed column. When doing lookup, we either assert (input1, input2) in (table1_fixed, table2_any), or assert an_input_product in table_product_with_at_least_one_fixed_col.

lispc:
btw in the original zcash/halo2, only fixed columns are allowed to be lookup table. So there are no this type of problem in their circuits.

han0110:
I think the first solution should be the originally expected one, but the second one sounds even easier to implement without too much overhead. I think it should be fine to adopt either one.

naure:
(1) Clearly all rows should be constrained. There should not be a concept of _selector_, meaning a flag that disables rules. Instead, there can be case flags, meaning a flag that chooses between _either_ this rule _or_ that rule. In the example given, the cases are `data` or `padding`, and the second rule enforces padding values, e.g. `tag == padding`.

(2) A fixed column sounds exactly like the mechanism of the blinding factors, so it will have the same problem. The actual data will need to be constrained and padded up to the fixed size provisioned by the fixed table.

The simplest seems to be to reserve a tag for padding.

lispc:
do you think end_block can handle this? or this is indeed a problem? @ed255

icemelon:
Another alternative solution is that we include the selector column in the lookup argument such that we only look up into the rows with the selector set to 1.

lispc:
> Another alternative solution is that we include the selector column in the lookup argument such that we only look up into the rows with the selector set to 1.

Yes that is my second proposal it is more robust / intuitive than first proposal. 2**k-blinding_rows may be a bit complex and error-prone dealing with multi subcircuits?

icemelon:
[...]

PR #1067, opened 2023-01-16
fix extcodecopy: non-exsit-proof, challenge api, and other fixes

lispc:
1.  use non-exist-proof
2. use challenge api
3. fix stack push order
4. add EXTERNAL_ADDRESS to TestContext
5. fix reversible_write_counter

previously, due to the incorrect param stack pushing order, an OOG error is triggered and handled by DummyGadget, so all the above errors cannot be catched by tests.

### DIFF
--- a/bus-mapping/src/evm/opcodes/extcodecopy.rs
+++ b/bus-mapping/src/evm/opcodes/extcodecopy.rs
@@ -114,7 +114,6 @@ fn gen_extcodecopy_step(
     } else {
         H256::zero()
     };
-
     state.account_read(
         &mut exec_step,
         external_address,
--- a/bus-mapping/src/evm/opcodes/extcodecopy.rs
+++ b/bus-mapping/src/evm/opcodes/extcodecopy.rs
@@ -243,7 +243,11 @@ mod extcodecopy_tests {
         });
 
         let bytecode_ext = Bytecode::from(code_ext.to_vec());
-        let code_hash = keccak256(code_ext.clone());
+        let code_hash = if code_ext.is_empty() {
+            Default::default()
+        } else {
+            keccak256(code_ext.clone())
+        };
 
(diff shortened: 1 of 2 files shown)


ed255 (approved):
LGTM overall! Thanks for the quick fix!

There's one detail I'd like to discuss.  With this circuit implementation, when the account doesn't exist the code_hash will be `0`.  Then a bytecode_length is done with this code_hash, which gives code_size = 0.  This BytecodeTable lookup will suceed because we have the 0 row in the BytecodeTable.
- I quickly looked at the Bytecode Circuit and I didn't find the assignation of an explicit 0 row.  Maybe I'm missing something, because AFAIK this row is required the EVM circuit to work (otherwise the bytecode lookups can never be disabled).
- In https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1046 I followed a different approach: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/a6267835e4c5dccb991d803e9a4b25fff58986a2/zkevm-circuits/src/evm_circuit/execution/extcodesize.rs#L57-L66 But maybe using the 0-row is smarter because we avoid requiring the IsZeroGadget.  So maybe we can update it later (also for consistency).

lispc:
the first issue may be related to https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/866 ? even we don't insert a full-0 row, it is there...

ed255:
> the first issue may be related to #866 ? even we don't insert a full-0 row, it is there...

Oh you're right...  We should address #866.  So the reason why 0-row lookups suceed to the Bytecode Table is because the bottom rows that haven't been assigned and appear before the blinding rows are set to 0.  But they could be set to anything by the prover, making any arbitrary bytecode lookup suceed (in the current implementation)

roynalnaruto (approved):
Looks good to me!

PR #1068, opened 2023-01-16
fix check rw lookup with correct challenge API

lispc:


ed255 (approved):
LGTM! Thanks for the fix!

PR #1069, opened 2023-01-16
fix delegate call with insufficient balance

DreamWuGit:
fix two issues:
1. only check insufficient balance when call/callcode, met error with delegatecall
2. use caller.address for balance reading.

### DIFF
--- a/bus-mapping/src/evm/opcodes/callop.rs
+++ b/bus-mapping/src/evm/opcodes/callop.rs
@@ -122,7 +122,7 @@ impl<const N_ARGS: usize> Opcode for CallOpcode<N_ARGS> {
         let is_call_or_callcode = call.kind == CallKind::Call || call.kind == CallKind::CallCode;
         let insufficient_balance = call.value > caller_balance && is_call_or_callcode;
 
-        log::info!(
+        log::debug!(
             "insufficient_balance: {}, call type: {:?}, sender_account: {:?} ",
             insufficient_balance,
             call.kind,
--- a/bus-mapping/src/evm/opcodes/callop.rs
+++ b/bus-mapping/src/evm/opcodes/callop.rs
@@ -119,7 +119,15 @@ impl<const N_ARGS: usize> Opcode for CallOpcode<N_ARGS> {
         debug_assert!(found);
 
         let caller_balance = sender_account.balance;
-        let insufficient_balance = call.value > caller_balance;
+        let is_call_or_callcode = call.kind == CallKind::Call || call.kind == CallKind::CallCode;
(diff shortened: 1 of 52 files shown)


bus-mapping/src/evm/opcodes/callop.rs
```diff
@@ -119,15 +119,23 @@ impl<const N_ARGS: usize> Opcode for CallOpcode<N_ARGS> {
...
-        let insufficient_balance = call.value > caller_balance;
+        let is_call_or_callcode = call.kind == CallKind::Call || call.kind == CallKind::CallCode;
+        let insufficient_balance = call.value > caller_balance && is_call_or_callcode;
+
+        log::info!(
+            "insufficient_balance: {}, call type: {:?}, sender_account: {:?} ",
+            insufficient_balance,
+            call.kind,
+            call.caller_address
+        );
```
lispc, line 130: don't need this high level logging. `insufficient_balance` is a "normal exec state", don't need to be highlight as log::info. debug and trace will be ok.

DreamWuGit: updated in [b5f92fb](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1069/commits/b5f92fb6dc693ba2e826ae879e46b61252a5dd8b) thanks !!

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -488,8 +488,7 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
         self.is_insufficient_balance
             .assign(region, offset, caller_balance, value)?;
 
-        let is_insufficient = value > caller_balance;
-
+        let is_insufficient = (value > caller_balance) && (is_call || is_callcode);
```
ed255, line 491: The corresponding circuit expression only does `is_insufficient = value > caller_balance`.  This is because when `not(is_call || is_callcode)` then `value = 0` so `value > caller_balance == 0 > caller_balance == false`.

So I believe adding the `&& (is_call || is_callcode` here is redundant.  Nevertheless maybe it makes things more clear?  This is just an observation, not a request to change anything.

ed255 (approved):
LGTM! Thanks for the fix!

ed255:
@lispc do you consider this PR approved? Or would you like to see some changes?

zkevm-circuits/src/evm_circuit/execution/callop.rs
lispc: i also prefer to remove the redundant `&& (is_call || is_callcode`.

lispc (approved):

PR #1076, opened 2023-01-17
state circuit allows non zero old_state_root

lispc:
and other improvements:

1. rename power_of_randomness to instance in state circuit tests. There is no power_of_randomness now.
2. change MptUpdates from HashMap to BTreeMap. So each run of tests produces same columns without non-deterministic. Useful for debugging.


han0110 (approved):
LGTM!

PR #1078, opened 2023-01-17
refactor copy circuit

andyrdt:
- Refactor copy circuit to take vec of `CopyEvent`, rather than a `Block`
- Restores negative tests for copy circuit

### DIFF
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -433,7 +433,7 @@ impl<F: Field> CopyCircuitConfig<F> {
     pub fn assign_copy_event(
         &self,
         region: &mut Region<F>,
-        mut offset: usize,
+        offset: &mut usize,
         tag_chip: &BinaryNumberChip<F, CopyDataType, 3>,
         lt_chip: &LtChip<F, 8>,
         challenges: Challenges<Value<F>>,
@@ -454,7 +454,7 @@ impl<F: Field> CopyCircuitConfig<F> {
                     region.assign_advice(
                         || format!("{} at row: {}", label, offset),
                         *column,
-                        offset,
+                        *offset,
                         || value,
                     )?;
                 }
@@ -462,13 +462,13 @@ impl<F: Field> CopyCircuitConfig<F> {
 
             // q_step
             if is_read {
-                self.q_step.enable(region, offset)?;
+                self.q_step.enable(region, *offset)?;
             }
             // q_enable
             region.assign_fixed(
                 || "q_enable",
                 self.q_enable,
-                offset,
+                *offset,
                 || Value::known(F::one()),
             )?;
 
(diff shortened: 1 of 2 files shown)


roynalnaruto (approved):
Minor point. Also you need to resolve some a merge conflict

roynalnaruto (approved):
Looks good!

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -610,30 +634,56 @@ impl<F: Field> CopyCircuitConfig<F> {
...
         Self {
-            block: Some(Block::default()),
-            max_txs: usize::default(),
+            copy_events: vec![],
+            _marker: PhantomData::default(),
+            #[cfg(test)]
+            test_data: CopyCircuitTestData::default(),
         }
     }
 }
```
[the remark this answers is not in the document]
z2trillion: previously this wouldn't have worked because Option<T>::default() is None, but now every field is just set to its default value, which is what the derive implementation will do.

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -425,101 +429,117 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
...
-                                F::from(copy_event.src_addr_end),
-                            )?;
-                        }
-
-                        offset += 1;
-                    }
+                for copy_event in copy_events.iter() {
+                    self.assign_copy_event(
+                        &mut region,
+                        offset,
```
ed255, line 533: Consider passing `&mut offset` so that offset is incremented via the `assign_copy_event` without requiring the extra `offset += 2 * copy_event.bytes.len();`

ed255 (approved):
LGTM! I've added a small note but it's just a suggestion, feel free to ignore it :)

zkevm-circuits/src/copy_circuit.rs
andyrdt: I like it - feels cleaner. I also changed the padding methods to take `&mut offset` as well.

z2trillion (approved):

lispc:
the CI on main failed after this PR. Three failed cases are all copy circuit related. @andyrdt 

https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3989894924/jobs/6842879646

ed255:
> the CI on main failed after this PR. Three failed cases are all copy circuit related. @andyrdt
> 
> https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/3989894924/jobs/6842879646

I'm working on it: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1103

andyrdt:
> 

thanks for the quick fix @ed255. I thought CI tests were run with PR check, but now I see they were skipped. I will be sure to run CI tests locally next time!

ed255:
> thanks for the quick fix @ed255. I thought CI tests were run with PR check, but now I see they were skipped. I will be sure to run CI tests locally next time!

Currently the integration tests are only run after a commit lands in main, so we don't see the results in the PR.  Previously we ran them on all PRs, but as we added more circuits to the tests they started taking more and more time to run and that became an issue.  Also we only have 1 server to run them, so when we had updated 3 PRs, each one had to wait for the previous one to run these tests.

Now we have a label called `trigger-integration-tests` that we can manually add to a PR and that cause the integration tests to run for that PR (we usually do this for PRs that touch the integration tests).  To add the label I think you need write permissions to the repositories.

Nevertheless this PR didn't strictly touch the integration tests, so it was not obvious that it could break them!  I think the current workflow works well enough :)

PR #1079, opened 2023-01-17
Handle accounts with non-zero code hashes in BeginTx

z2trillion:
This occurs when an address already exists because it has a non-zero balance.


ed255:
@z2trillion this PR has been draft for a while.  The code only introduces a new test, does it still make sense to move forward with this?

z2trillion:
8 months ago the test failed because the code hash for accounts in the process of being created wasn't being set the the correct value of keccak(nil). I think this got fixed now that we're using `transfer_with_fee` with `must_create = call.is_create()` https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/begin_end_tx.rs#L111.

I can add a comment to the test explaining what it catches, or I can just close this PR.

lispc:
> must_create = call.is_create()

Oh this is incorrect in fact. I remember some time ago we discussed this issue in pse repi issues. And i later fixed it in scroll repo.    Sometimes the address can already has some ether. So need to first read the codehash.

z2trillion:
> > must_create = call.is_create()
> 
> ~Oh this is incorrect in fact. I remember some time ago we discussed this issue in pse repo issues. And i later fixed it in scroll repo. Sometimes the address can already has some ether. So need to first read the codehash.~. ignore this..

This is actually still a problem. https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1079/commits/771f58fe2f9d590d670330c3fd0560b843e08eb5 adds a test that fails when the account that's being created already has a non-zero balance (and hence a non-zero code hash).

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -553,7 +553,7 @@ impl<'a> CircuitInputStateRef<'a> {
             sender_balance
         );
         // If receiver doesn't exist, create it
-        if (!receiver_exists && !value.is_zero()) || must_create {
+        if !receiver_exists && (!value.is_zero() || must_create) {
```
ChihChengLiang, line 556: This comment is not a request for change but a note for other reviewers:

I wrapped my head around this expression and finally figured out a more straightforward way to convince myself the proposed change is correct. 

- If the expression evaluates true, we create an account with an empty hash. Specifically, update the codehash from 0 to the EMPTY_CODE_HASH.
- In the previous expression, the `must_create` dominates the outcome. If the call is CREATE, then even if the receiver_exists, the receiver account is still created with the empty hash. 
- In the proposed expression, the `!receiver_exists` dominates the outcome. If the receiver already exists, there's no way to mess up the account code hash again. This is the behavior that makes sense.

The failing case is captured in the test case `create_tx_for_existing_account`.

I've included my previous elaborate boolean analysis in the "Details".


<details><summary>Details</summary>
<p>

Let A = (!receiver_exists), B = !value.is_zero(), and C = must_create,
- before: (A && B) || C
- after: A && (B || C)

The two expressions disagree on the following cases
- A = False,  B = False, C = True
- A = False,  B = True, C = True

In both cases, the receiver exists, and the call is CREATE. 

If the value is zero, the expression before still creates the account with the empty codehash, but the proposed one doesn't.

If the value is NOT zero, the expression before still creates the account with the empty codehash, but the proposed one doesn't.

|A | B | C | Output|
|-- | -- | -- | --|
0 | 0 | 0 | F
0 | 0 | 1 | T
0 | 1 | 0 | F
0 | 1 | 1 | T
1 | 0 | 0 | F
1 | 0 | 1 | T
1 | 1 | 0 | T
1 | 1 | 1 | T





|A | B | C | Output|
|-- | -- | -- | --|
0 | 0 | 0 | F
0 | 0 | 1 | F
0 | 1 | 0 | F
0 | 1 | 1 | F
1 | 0 | 0 | F
1 | 0 | 1 | T
1 | 1 | 0 | T
1 | 1 | 1 | T



</p>
</details>

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -920,4 +920,26 @@ mod test {
...
+        let new_address = get_contract_address(address, nonce + 1);
+
+        let ctx = TestContext::<1, 2>::new(
+            None,
+            |accs| {
+                accs[0].address(address).nonce(nonce).balance(eth(10));
+            },
+            |mut txs, _| {
+                txs[0].from(address).to(new_address).value(eth(2));
+                txs[1].from(address);
```
ChihChengLiang, line 937: Some comments would be helpful. 
```suggestion
                txs[0].from(address).to(new_address).value(eth(2)); // Initialize new_address with some balance and an empty code hash
                txs[1].from(address); // Run a CREATE tx on new_address
```

ChihChengLiang (approved):
LGTM with some minor feedback and notes.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -216,13 +216,13 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
 
         // TODO: And not precompile
-        cb.condition(not::expr(tx_is_create.expr()), |cb| {
-            cb.account_read(
-                tx_callee_address.to_word(),
-                AccountFieldTag::CodeHash,
-                code_hash.to_word(),
-            ); // rwc_delta += 1
-        });
+        // i think this needs to be removed....
```
ed255, line 219: Do you mean removing the "TODO" in line 218?  If that's the case, I agree it can be removed!

zkevm-circuits/src/evm_circuit/execution/extcodehash.rs
```diff
@@ -263,4 +263,28 @@ mod test {
             test_ok(Some(account), false);
         }
     }
+
+    #[test]
+    // Regression test to ensure that the code hash for an account that is is being initialized is
+    // the empty code hash.
+    fn create_tx_extcodehash() {
```
ed255, line 270: I think I'm missing something while reading this test.  How is this test checking that the initialized account has the empty code hash?

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ed255, line 556: We have a very similar logic in `transfer_to`:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/ee97ab1f957918defafb9bc38dd51adf222ddfec/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L630
Should we change that as well?

ed255 (approved):
LGTM!  I've left a few minor questions, and I think following CC's request to add some comments https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1079#discussion_r1336622556 would be really nice.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
z2trillion: `transfer_to` is only ever being called with `must_create` being false, so I just removed the argument.

zkevm-circuits/src/evm_circuit/execution/extcodehash.rs
z2trillion: The tx does not have `to` set, so it is an account creation tx. The result ADDRESS and then EXTCODEHASH should be the code hash of the current account, which in this case will be keccak(nil). Previously in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1079/commits/b682805243f4f7c73571af56511830b8c0bf9885, this would produce an bus-mapping error, as we would claim that the code hash was 0 instead.

ed255: Got it, thanks a lot for the explanation!

Issue #1086, opened 2023-01-18
evm: a lot of opcode gadgets need to be updated with `CellType::StoragePhase2`


lispc:
eg: gas_price cell should be a phase2 cell.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/f7941c1525f3d93153d207d1c33a3abde6624f72/zkevm-circuits/src/evm_circuit/execution/gasprice.rs#L80-L81

lispc:
modify MockProver to catch these types of errors. 

https://github.com/scroll-tech/halo2/pull/25/commits/e409a69746d14e946bd177f640825ea47225e8d8

yeah a lot of tests failed.

ed255:
We would like to address replacing all the word_rlc usage by word hi/lo very soon (probably in the coming weeks).  For this reason we haven't been very thorough with correctness/soundness reviews and analysis related to the usage of the Challange API.  And I think we won't focus too much on fixing these issues ourselves until the word hi/lo is implemented (which will remove a lot of Challenge API usage).  Having said that, any fix related to these issue will be welcome and reviewed!

lispc:
i think hi-lo cannot solve all the challenge API problem? It can solve evm_word and. lookup_input randomness, but will you plan to also remove the dynamic length keccak_input randomness?

We did some experiment earlier but did not continue finishing it 

https://github.com/scroll-tech/zkevm-circuits/pull/216

ed255:
> i think hi-lo cannot solve all the challenge API problem? It can solve evm_word and. lookup_input randomness, but will you plan to also remove the dynamic length keccak_input randomness?

Definitely, word hi-lo will not solve all Challenge API problems, but hopefully it solves many of them (and reduces some circuit complexity).  Our current plan is to keep using the challenge API (via RLC) for the lookups in EVM Circuit and the dynamic length keccak inputs.

> We did some experiment earlier but did not continue finishing it
> 
> [scroll-tech#216](https://github.com/scroll-tech/zkevm-circuits/pull/216)
> 
> btw jonathan made more improvements based on our version
> 
> https://github.com/jonathanpwang/zkevm-circuits/tree/keccak-only-subtree

Interesting! Thanks for sharing these links!

PR #1087, opened 2023-01-19
fix evm circuit phase2 cells

lispc:
fix https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1086

(these types of bugs can be detected with this modified MockProver https://github.com/scroll-tech/halo2/commit/e409a69746d14e946bd177f640825ea47225e8d8 )


han0110 (approved):
LGTM! Thanks for fixing this.

Also thanks for the reference of `MockProver` fix, I'll try to resolve the inconsistent behavior between `MockProver` and real prover to avoid this kind of mistake to happen.

PR #1090, opened 2023-01-19
Make EVM circuit work with multi-phase `MockProver`

han0110:
This PR aims to fix EVM circuit to have all advice columns in correct phase to work with multi-phase `MockProver` https://github.com/privacy-scaling-explorations/halo2/pull/129, and should also work with real prover.

Base on #1087.


lispc:
another related problem: i guess this line may be wrong. Need to be less than?

https://github.com/privacy-scaling-explorations/halo2/blob/99795dd4dfd48478fcbbd86a6610df814c282ad4/halo2_proofs/src/plonk/prover.rs#L207

for example, in evm circuit,  the stored_expressions in CachedRegion need to "pick" assigned values to do RLC and assign the results. It seems if we don't assign phase1 columns in phase2/phase3, it picks a zero value, and do incorrect RLC..

Still investigating and debugging..

han0110:
> another related problem: i guess this line may be wrong. Need to be less than?

hmm I kind of forget the reason to be that strict (I thought it's because the `to` is `FnMut` but it seems not). This is the original discussion on this topic https://github.com/zcash/halo2/pull/593#issuecomment-1146436341. Mainly I might just wanted to keep the convention that in unrelated phase we don't call the `to` (for instance in `keygen` the `to` in `assign_advice` will not be called, and in `create_proof` the `to` in `assign_fixed` will not be called as well).

But this indeed causes the trouble you said (in later phase we can't get the value from `AssignedCell` of previous phase), and that's the reason I did the change like this (note that the `to` is called twice): https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/97491b285145640b10d4e6a7ae3b6532f5566a97/zkevm-circuits/src/evm_circuit/util.rs#L148-L174

lispc:
> > another related problem: i guess this line may be wrong. Need to be less than?
> 
> hmm I kind of forget the reason to be that strict (I thought it's because the `to` is `FnMut` but it seems not). This is the original discussion on this topic [zcash/halo2#593 (comment)](https://github.com/zcash/halo2/pull/593#issuecomment-1146436341). Mainly I might just wanted to keep the convention that in unrelated phase we don't call the `to` (for instance in `keygen` the `to` in `assign_advice` will not be called, and in `create_proof` the `to` in `assign_fixed` will not be called as well).
> 
> But this indeed causes the trouble you said (in later phase we can't get the value from `AssignedCell` of previous phase), and that's the reason I did the change like this (note that the `to` is called again to always get the value):
> 
> https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/97491b285145640b10d4e6a7ae3b6532f5566a97/zkevm-circuits/src/evm_circuit/util.rs#L148-L174

thanks! this seems a better solution!

lispc (approved):

ed255 (approved):
LGTM!

Issue #1084, opened 2023-01-18
Review and add checks to verify account updates consistency


ed255:
Add checks for every account field read/write to make sure that:
- Verify that a READ dosn't change the field value
- Verify that the previous value matches the account field value in the StateDB
- Verify that no read is done to a field other than CodeHash to a non-existing
  account (only CodeHash reads with value=0 can be done to non-existing
  accounts, which the State Circuit translates to MPT
  AccountNonExisting proofs lookups).

Fix bugs related to these checks not passing

PR #1023, opened 2023-01-04
Refactor/callop call oog

KimiWu123:
fix  #940

PR #1092, opened 2023-01-19
feat: panic on invalid account lookups

ed255:
After this was merged https://github.com/privacy-scaling-explorations/zkevm-specs/pull/354 the rule for opcodes that can access non-existing accounts is to first read the CodeHash, and only if it's not 0 (meaning that the account exists), proceed with other account field reads.  If a field read is performed to a non-existing account, this may end up mapping to an MPT account field proof for a non-existing account (which are not supported); so the MPT proof would be impossible to generate.  We can add some checks to make sure we don't implement any circuit that generates this situation; but we can even add more sanity checks to detect possible bugs early.  In particular we can check for each account field lookup (either read or write): 

- That a READ doesn't change the field value
- That the previous value matches the account field value in the StateDB
- That no read is done to a field other than CodeHash to a non-existing
  account (only CodeHash reads with value=0 can be done to non-existing
  accounts, which the State Circuit translates to MPT
  AccountNonExisting proofs lookups).

If any of these two points don't hold, the corresponding chain of MPT proofs
can't be generated.

In order to perform this checks in a single place, I've moved the logic to update the accounts in the StateDB performed by the circuit input builder in a single place, after the check (instead of requiring each bus-mapping opcode implementation to do it by themselves).

With these checks I found some bugs in the `callop` and `error_oog_call`, which I have fixed.

As a bonus I added the method `debug_print_txs_steps_rw_ops` to the `witness::Block` which prints all the rw operations (with their content) of each step in each tx, which was very useful for debugging.

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1084

Depends on https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1069

### DIFF
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -276,8 +276,8 @@ impl<'a> CircuitInputStateRef<'a> {
             }
         }
         let account = self.sdb.get_account_mut(&op.address).1;
-        if false {
-            // DBG
+        // Here starts the sanity check
+        if true {
             let account_value_prev = match op.field {
                 AccountField::Nonce => account.nonce,
                 AccountField::Balance => account.balance,
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -262,6 +262,45 @@ impl<'a> CircuitInputStateRef<'a> {
         Ok(())
     }
 
+    fn sanity_check_account_op(&mut self, rw: RW, op: &AccountOp) {
+        if matches!(rw, RW::READ) {
+            if op.value_prev != op.value {
+                panic!(
+                    "RWTable Account field read where value_prev != value rwc: {}, op: {:?}",
+                    self.block_ctx.rwc.0, op
+                )
+            }
+        }
+        let account = self.sdb.get_account(&op.address).1;
+        let account_value_prev = match op.field {
+            AccountField::Nonce => account.nonce,
+            AccountField::Balance => account.balance,
+            AccountField::CodeHash => {
+                if account.is_empty() {
+                    Word::zero()
+                } else {
+                    account.code_hash.to_word()
+                }
+            }
+        };
+        if op.value_prev != account_value_prev {
+            panic!("RWTable Account field {:?} lookup doesn't match account value account: {:?}, rwc: {}, op: {:?}",
+                rw,
+                account,
+                self.block_ctx.rwc.0,
+                op
+            );
+        }
+        if !matches!(op.field, AccountField::CodeHash) {
+            if account.is_empty() {
+                panic!(
+                    "RWTable Account field {:?} lookup to non-existing account rwc: {}, op: {:?}",
+                    rw, self.block_ctx.rwc.0, op
+                );
+            }
+        }
+    }
+
     /// Push a read type [`AccountOp`] into the
     /// [`OperationContainer`](crate::operation::OperationContainer) with the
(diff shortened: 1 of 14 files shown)


bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -276,11 +333,9 @@ impl<'a> CircuitInputStateRef<'a> {
...
         value_prev: Word,
     ) -> Result<(), Error> {
-        self.push_op(
-            step,
-            RW::READ,
-            AccountOp::new(address, field, value, value_prev),
-        );
+        let op = AccountOp::new(address, field, value, value_prev);
+        self.check_update_sdb_account(RW::READ, &op);
+        self.push_op(step, RW::READ, op);
```
lispc, line 338: what about put the logic of check_update_sdb_account inside `push_op`? So in the future developers never omit this check.

```
fn push_op(rw, op, ...):
    if op.type == Account:
       check_update_sdb_account(rw, op)
    ...

```

lispc (approved):

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
```diff
@@ -65,15 +60,16 @@ impl<F: Field> ExecutionGadget<F> for ErrorOOGCallGadget<F> {
             is_warm.expr(),
         );
 
-        let balance = cb.query_word_rlc();
-        cb.account_read(
-            call_gadget.callee_address_expr(),
-            AccountFieldTag::Balance,
-            balance.expr(),
-        );
+        // TODO: Why is calle balance read here? There's no constraint related to this
```
ChihChengLiang, line 63: We can safely remove the balance read here. The old OOG callop uses zero nonce, zero balance, and empty code hash to determine the empty account. In #1023, balance and nonce check are removed, but the balance read was a leftover. 

Reference: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1023#discussion_r1066505492

ChihChengLiang (approved):
LGTM. I added some feedback.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -237,7 +240,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
                 //   - Write CallContext IsRoot
                 //   - Write CallContext IsCreate
                 //   - Write CallContext CodeHash
-                rw_counter: Delta(23.expr()),
+                rw_counter: Delta(22.expr() + (1.expr() - tx_is_create.expr())),
```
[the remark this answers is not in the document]
ed255: Thanks for catching this! updated.

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -429,6 +445,7 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
         call: &Call,
         step: &ExecStep,
     ) -> Result<(), Error> {
+        // dbg!(offset);
```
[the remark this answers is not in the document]
ed255: oops! thanks!

zkevm-circuits/src/evm_circuit/execution/error_oog_call.rs
ed255: ah thanks for the explanation! I've removed the commented code :)

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ed255: Good idea! I've applied it in a new commit, thanks!

PR #1110, opened 2023-01-25
fix(testool) more readable HTML output

adria0:
At this moment the generated HTML report is quite unreadable, this makes it more easy to read.

### Contents

- Store error details URL encoded to keep one line per entry in the CSV file instead removing spaces and CRs
- Trim results in diffs
- Add an index to the HTML file


ed255 (approved):
LGTM!

CPerezz (approved):
LGTM!

Issue #1091, opened 2023-01-19
Refactor RandomLinearCombination::<F, 32>::random_linear_combine_expr to accept challenge instead of powers


leolara:
Currently the signature of this method is:

```
pub(crate) fn random_linear_combine_expr(
        bytes: [Expression<F>; N],
        power_of_randomness: &[Expression<F>],
    ) -> Expression<F>
```

but the second parameter should be the randomness instead of an array of its powers. It was implemented like this to simulate the challenge before it was implemented.

Hence the signature should be:

```
pub(crate) fn random_linear_combine_expr(
        bytes: [Expression<F>; N],
        randomness: Expression<F>,
    ) -> Expression<F>
```

As discussed with @ed255

leolara:
Perhaps the same should be applied to rlc::expr? @ed255

ed255:
> Perhaps the same should be applied to rlc::expr? @ed255

Yes, the same reasoning could be applied to `rlc::expr`.

I've checked how the EVM circuit deals with the powers of randomness with the challange API and I see that they are constructed with these: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/98646613d517f473489d7b18d4ea8b53a43b5e11/zkevm-circuits/src/util.rs#L111-L130
So these expressions are computed once, and reused with all RLCs in the EVM circuit.

If we replace the rlc functions following your proposal, then the powers of random would be calculated many times (each time an rlc expression is required).  Since these expressions use the challenge, I'm not 100% sure if it may affect the verifier cost, because AFAIK the verifier needs to calculate the challenge, and maybe also evaluate the expressions used with it?  @han0110 could you help clarify this please?  Can such a change modify the verification cost?

han0110:
Previously it's taking a slice of powers because we were using instance columns as challenges, so to avoid degree is blown up to 32. But now with challenge API we don't need to worry about that.

As for prover/verifier cost I think the difference should be negligible (same amount of mul/add).

leolara:
This task also implies to update the constrain builder that accept a power of randomness array.

PR #1113, opened 2023-01-26
Refactor RandomLinearCombination::<F, 32>::random_linear_combine_expr to accept challenge instead of powers

leolara:
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1091


zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -542,30 +528,45 @@ pub(crate) mod from_bytes {
...
+            .map(|v| F::from(*v as u64))
+            .collect::<Vec<F>>();
+        if !values.is_empty() {
+            gen(values, randomness)
+        } else {
+            F::zero()
+        }
+    }
+
+    pub(crate) fn gen<V, I>(values: I, randomness: V) -> V
```
leolara, line 560: If I could use a trait that includes ::zero then I could return it when the values is an empty iterator. Now, this is only done for expr and value, I am thinking in creating an issue in FF.

ed255, line 560: It looks like `fn gen` is an internal helper function used in `expr` and `value`, so maybe it doesn't need the `pub(crate)`?

ed255 (approved):
LGTM! Great refactor, everything looks better now :smiley: And it's always nice to see more lines removed than added!

zkevm-circuits/src/evm_circuit/util.rs
ed255: You mean defining the trait `Field` in `halo2_proofs::arithmetic::Field` as `... + num_traits::identities::Zero` right?

leolara: I have been looking into this, I think https://crates.io/crates/num-traits is not standard, and we would need to convinced several open source projects to use that trait.

I think it is a pitty that core::ops does not include a Zero and One trait, perhaps we should request it.

leolara: My idea is that anyone could use gen for doing RCL with any type that has the traits needed. But yes, currently it is not used outside the module, and without the Zero trait is not that generic so perhaps we should make it private.

leolara:
> LGTM! Great refactor, everything looks better now 😃 And it's always nice to see more lines removed than added!

Always happy yo remove code :-)

zkevm-circuits/src/evm_circuit/util.rs
ed255: I see! In that case feel free to choose, (leave `pub(crate)` or remove it), both seem fine with me.

zkevm-circuits/src/witness/tx.rs
```diff
@@ -63,12 +61,9 @@ impl Transaction {
...
                     Value::known(F::zero()),
-                    challenges.evm_word().map(|evm_word| {
-                        RandomLinearCombination::random_linear_combine(
-                            self.gas_price.to_le_bytes(),
-                            evm_word,
-                        )
-                    }),
+                    challenges
+                        .evm_word()
+                        .map(|evm_word| rlc::value(&self.gas_price.to_le_bytes(), evm_word)),
```
ChihChengLiang, line 66: It is a dummy variable but naming it `challenge`, `randomness`, or `r` could be less confusing.
```suggestion
                        .map(|challenge| rlc::value(&self.gas_price.to_le_bytes(), challenge)),
```

zkevm-circuits/src/witness/tx.rs
```diff
@@ -92,12 +87,9 @@ impl Transaction {
...
                     Value::known(F::zero()),
-                    challenges.evm_word().map(|evm_word| {
-                        RandomLinearCombination::random_linear_combine(
-                            self.value.to_le_bytes(),
-                            evm_word,
-                        )
-                    }),
+                    challenges
+                        .evm_word()
+                        .map(|evm_word| rlc::value(&self.value.to_le_bytes(), evm_word)),
```
ChihChengLiang, line 92: Same here
```suggestion
                        .map(|challenge| rlc::value(&self.value.to_le_bytes(), challenge)),
```

ChihChengLiang (commented):
Added some nitpicks. Feel free to ignore them.

leolara:
Happy to see this PR has got so much attention 😊 I will fix these things today. I am with the integration thing

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -542,30 +528,45 @@ pub(crate) mod from_bytes {
...
+    where
+        I: IntoIterator<Item = V>,
+        <I as IntoIterator>::IntoIter: DoubleEndedIterator,
+        V: Clone + Add<Output = V> + Mul<Output = V>,
+    {
+        let mut values = values.into_iter().rev();
+        let init = values.next().expect("values should not be empty");
+
+        values.fold(init, |acc, value| acc * randomness.clone() + value)
```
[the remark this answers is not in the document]
leolara: So randomness is moved by the multiplation, that is why there is a clone. If you know a way to avoid it, let me know. I only have been doing rust for 6 months so perhaps there are techniques that are not obvious to me. This is based on the trait Mul:

https://doc.rust-lang.org/beta/core/ops/trait.Mul.html

that in principle moves Rhs. I guess I could type Rhs = V, but I don't think the types that we are using will have it implemented that way.

PR #1144, opened 2023-02-07
add `test-circuits` feature flag

pinkiebell:
Add an additional feature flag that exports the Circuit implementation for
sub-circuits. This prevents buildtime overhead and crossbuild errors when building with the `test` feature.

- evm circuit is an exception because of EXECUTION_STEP_HEIGHT (always exported)
- pi circuit is exported via PiTestCircuit because of associated types

Additionally:
- add Clone,Copy,Debug to ProofType
   Fixes build errors with newer rust versions


ed255 (approved):
LGTM!

ed255:
> Add an additional feature flag that exports the Circuit implementation for sub-circuits. This prevents overhead and crossbuild errors when building with the `test` feature.

Could you give some details on the overhead?  Is it a build-time overhead? run-time overhead? binary size overhead?

So the plan is to import `zkevm-circuits` with the `test-circuits` feature instead of `test` feature in the `zkevm-chain` right?

pinkiebell:
> > Add an additional feature flag that exports the Circuit implementation for sub-circuits. This prevents overhead and crossbuild errors when building with the `test` feature.
> 
> Could you give some details on the overhead? Is it a build-time overhead? run-time overhead? binary size overhead?
> 
> So the plan is to import `zkevm-circuits` with the `test-circuits` feature instead of `test` feature in the `zkevm-chain` right?

Yes. It's build overhead and build failures when cross compiling (a problem with emitted go output)

han0110 (approved):
LGTM!

PR #1145, opened 2023-02-07
Fix CI Checks workflow to support ephemeral GH runners

ntampakas:


pinkiebell (approved):
LGTM. What about the other workflows, gradually replacing them?

ntampakas:
> LGTM. What about the other workflows, gradually replacing them?

Exactly, that' s the plan!

pinkiebell:
@ChihChengLiang I think we need to 'force' merge here. Looks like `Test (ci-checks-runner)` is a stale check that can never be executed because it was changed via this PR.

ChihChengLiang:
Hi @pinkiebell @ntampakas, I've removed the branch protection rule that requires Test (ci-checks-runner)

ChihChengLiang (approved):
LGTM.

PR #1161, opened 2023-02-12
tx refund synchronization

ashWhiteHat:
Part of https://github.com/privacy-scaling-explorations/zkevm-specs/issues/364
I synchronized tx refund on spec and the same stuff for circuits.
https://github.com/privacy-scaling-explorations/zkevm-specs/pull/367

What I did are followings.

- give serial number to constraints
- add [7.3 constraint](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/state_circuits/07TxRefund.md#circuit-behavior)
- remove account destructed

I would appreciate it if you could confirm.
Thank you!


ed255 (commented):
The circuit constraints update look good!
Regarding the removal of AccountDestructed related things I think we should at least keep the `SELFDESTRUCT` opcode while it still exists (that is, until this EIP https://eips.ethereum.org/EIPS/eip-4758 or a similar one gets accepted).

So I think we should keep the `OpcodeId::SELFDESTRUCT`, and consequently also `OogError::SelfDestruct`, `DummySelfDestruct`, `func (a *Asm) SelfDestruct()`, `ExecutionConfig.selfdestruct_gadget`, `ExecutionState::ErrorOutOfGasSELFDESTRUCT`.

Then I think we should only remove the AccountDestructed logic associated with the StateCircuit: `Target::AccountDestructed`, `OpEnum::AccountDestructed`, `AccountDestructedOp`, `StateDB.destructed_account`, `Rw::AccountDestructed`.

ashWhiteHat:
Hi @ed255 
Thank you for the review.
That makes sense to me.
I modified according to your review.

I would appreciate it if you could confim.
Thank you.

ed255 (approved):
LGTM! Thanks for this sync update!

han0110 (approved):
LGTM!

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -320,19 +318,26 @@ impl<F: Field> ConstraintBuilder<F> {
...
-
+        // 7.1. `state root` is not changed
         self.require_equal(
             "state_root is unchanged for TxRefund",
             q.state_root(),
             q.state_root_prev(),
         );
+        // 7.2. `initial value` is 0
+        self.require_zero("initial TxRefund value is 0", q.initial_value());
+        // 7.3. First access for a set of all keys are 0 if READ
```
z2trillion, line 336: This constraint is redundant because on https://github.com/privacy-scaling-explorations/zkevm-circuits/blame/main/zkevm-circuits/src/state_circuit/constraint_builder.rs#L156, we already constrain `value` to be `initial_value`, which is 0 by 7.2 above.

PR #1167, opened 2023-02-13
Exp circuit annotations

leolara:
### Description

Adds columns annotations to Exp circuit.

### Issue Link

Part of https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1106

### Type of change

### Contents

- Adds columns annotations to Exp circuit.

### Rationale

Following example linked in issue.

### How Has This Been Tested?

Compilation didn't yield errors.

### Note

Not sure if calling `exp_table.annotate_columns(meta);` as has been done in execution gadget, I guess it can be done twice without problem.

### DIFF
--- a/gadgets/src/mul_add.rs
+++ b/gadgets/src/mul_add.rs
@@ -363,6 +363,36 @@ impl<F: Field> MulAddChip<F> {
 
         Ok(())
     }
+
+    /// Returns the list of ALL the gadget advice columns.
+    fn advice_columns(&self) -> Vec<Column<Advice>> {
+        vec![
+            self.config.col0,
+            self.config.col1,
+            self.config.col2,
+            self.config.col3,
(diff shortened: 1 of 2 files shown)


ed255:
> Not sure if calling `exp_table.annotate_columns(meta);` as has been done in execution gadget, I guess it can be done twice without problem.

Probably we will repeat that call in the EVM Circuit.  This means that when testing EVM Circuit there would be 1 call, when testing Exp Circuit 1 call, and when testing Super Circuit 2 calls (repeated).  But repeating this call is not a problem (as you guessed!) because it calls `annotate_lookup_any_column` from halo2 which just inserts a key-value in a hashmap (so if there was already an annotation it will be replaced).

ed255 (approved):
LGTM!

PR #1173, opened 2023-02-14
Feat/root circuit annotation

KimiWu123:
### Description

Add columns annotation for root circuit.

### Issue Link

part of https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1106

### Type of change

### Contents

- Add columns annotation for root circuit.

### Rationale

see #1106 

### How Has This Been Tested?

Pass all the test cases.

### DIFF
--- a/zkevm-circuits/src/root_circuit/aggregation.rs
+++ b/zkevm-circuits/src/root_circuit/aggregation.rs
@@ -221,7 +221,13 @@ impl AggregationConfig {
         let snarks = snarks.into_iter().collect_vec();
         layouter.assign_region(
             || "Aggregate snarks",
-            |region| {
+            |mut region| {
+                // annotate advices columns of `main_gate`. We can't annotate fixed_columns of
+                // `main_gate` bcs there is no methods exported.
+                for (i, col) in self.main_gate_config.advices().iter().enumerate() {
+                    region.name_column(|| format!("ROOT_main_gate_{}", i), *col);
+                }
+
                 let ctx = RegionCtx::new(region, 0);
 
                 let ecc_chip = self.ecc_chip::<M::G1Affine>();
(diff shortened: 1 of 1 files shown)


ChihChengLiang (approved):
LGTM

zkevm-circuits/src/root_circuit/aggregation.rs
```diff
@@ -613,6 +613,16 @@ pub mod test {
             layouter.assign_region(
                 || "",
                 |mut region| {
+                    // annotation columns
```
ed255, line 616: By checking the source code I see that the circuit `StandardPlonk` is just a test circuit used to generate proofs and then aggregate them with the AggregationCircuit (via the `TestAggregationCircuit` type).
The `StandardPlonk` is never checked with the `MockProver` so I think these annotations are not needed (and are not annotations to the RootCircuit).

The RootCircuit that would require annotations is `RootCircuit` defined in `zkevm-circuits/src/root_circuit.rs` and the region it uses seems to be defined in `zkevm-circuits/src/root_circuit/aggregation.rs` by `"Aggregate snarks"`.
The `RootCircuit` uses the `MainGate` and `Range`, and it's their associated columns that should be annotated.
I checked the `MainGateConfig` and it has a method that returns the advices, but no method to return the fixed columns, so I think we can only do partial annotation (unless we update `halo2wrong` to expose the fixed columns).
As for the `RangeConfig` I think we're less lucky and it doesn't expose any of its columns.

KimiWu123: Thanks! I was hesitant to add annotation to a test circuit or not. I thought it's fine because it's a 'test' circuit.

fixed in [c48e4d8](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1173/commits/c48e4d8e90fa6af01d3113cbbdeabd88312b3721)

ed255 (approved):
LGTM!

PR #1174, opened 2023-02-14
Refactor/column annotation

KimiWu123:
### Description

Trying to mitigating the effort of adding column annotation. Having a set of util functions, `assign_fixed`, `assign_advice`, `assign_fixed_multi_value` and `assign_advice_multi_value`. It calls `name_column` and `assign_fixed` or `assign_advice` accordingly in these utils functions 
- pros:
  - it's easier to add annotations
- cons:
  - it would be missing the annotation for each cell, like [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1174/files#diff-c6e71bd5ec1e036ab77c032b5f67c9c7a772d54d81dcb08e9246e22dcd3d9d5eL107), but Han has confirmed that on one is using it right now. 

### Issue Link

na

### Type of change

### Contents

- Embedding `name_column` in util functions.

### Rationale

mitigating the effort of adding column annotation and to avoid missing any column annotation.

### How Has This Been Tested?

Pass all the test cases.

### DIFF
--- a/gadgets/src/batched_is_zero.rs
+++ b/gadgets/src/batched_is_zero.rs
@@ -5,15 +5,15 @@
 //!   it exists, `0` otherwise
 //! - is_zero: 1 if all `values` are `0`, `0` otherwise
 
+use eth_types::Field;
 use halo2_proofs::{
-    arithmetic::FieldExt,
     circuit::{Region, Value},
     plonk::{Advice, Column, ConstraintSystem, Error, Expression, Phase, VirtualCells},
     poly::Rotation,
 };
 use std::marker::PhantomData;
 
-use crate::util::Expr;
+use crate::util::{assign_advice, Expr};
 
 // TODO: Configurable Phase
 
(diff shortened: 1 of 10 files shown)


KimiWu123:
Hi @CPerezz , want to know your opinion on this refactor. This pr is only a proposal and basically, having another `assign_fixed` and `assign_advice` and call `name_column` inside these two helper functions. 
- pros:
  - easier to add annotations
  - won't have any columns are missed
- cons:
  - we would miss the annotation for each cell, like [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1174/files#diff-c6e71bd5ec1e036ab77c032b5f67c9c7a772d54d81dcb08e9246e22dcd3d9d5eL107), but Han has confirmed that on one is using it right now. 

would like to hear your idea, no sure if it's a breaking change or over-engineering.

KimiWu123:
close it, quote from Edu
>  unfortunately this would mean that we are naming the same column over and over unnecessarily

PR #1207, opened 2023-02-14
uniform log messages for benchmark proof generation/verification

AronisAt79:
### Description

  Log messages across different prover benchmark tests differ slightly. This PR makes them uniform to make it easier to
  parse and report results

### Type of change

## Files Modified: 
```
circuit-benchmarks/src/bytecode_circuit.rs
circuit-benchmarks/src/copy_circuit.rs
circuit-benchmarks/src/evm_circuit.rs
circuit-benchmarks/src/packed_multi_keccak.rs
circuit-benchmarks/src/pi_circuit.rs
circuit-benchmarks/src/super_circuit.rs
circuit-benchmarks/src/copy_circuit.rs
circuit-benchmarks/src/exp_circuit.rs
circuit-benchmarks/src/tx_circuit.rs
```

### DIFF
--- a/circuit-benchmarks/src/bytecode_circuit.rs
+++ b/circuit-benchmarks/src/bytecode_circuit.rs
@@ -31,6 +31,9 @@ mod tests {
             .parse()
             .expect("Cannot parse DEGREE env var as u32");
 
+        //Unique string used by bench results module for parsing the result
+        const BENCHMARK_ID: &str = "Bytecode Circuit";
+
         // Contract code size exceeds 24576 bytes may not be deployable on Mainnet.
         const MAX_BYTECODE_LEN: usize = 24576;
 
@@ -67,7 +70,7 @@ mod tests {
         let mut transcript = Blake2bWrite::<_, G1Affine, Challenge255<_>>::init(vec![]);
 
         // Bench proof generation time
-        let proof_message = format!("Bytecode Circuit Proof generation with {} rows", degree);
+        let proof_message = format!("{} Proof generation with degree = {}",BENCHMARK_ID, degree);
         let start2 = start_timer!(|| proof_message);
         create_proof::<
             KZGCommitmentScheme<Bn256>,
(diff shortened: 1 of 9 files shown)


ChihChengLiang (commented):
This fix looks fragile to me. The parsing relies on the correctness of the log messages.
But the circuit developers need to be made aware of this dependency and might change the log messages accidentally in the future.

What do you think we define constants of benchmark IDs? For example

```rust
// Benchmark server parses the benchmark result using this string.
const BENCHMARK_ID_BYTECODE_CIRCUIT = "BytecodeCircuit";
```
Then in the proof message, we have

```rust
let proof_message = format!("{} Proof generation with {} rows", BENCHMARK_ID_BYTECODE_CIRCUIT, degree);
```
So that future developers wouldn't change the message accidentally and cause the parsing to fail.

AronisAt79:
> This fix looks fragile to me. The parsing relies on the correctness of the log messages. But the circuit developers need to be made aware of this dependency and might change the log messages accidentally in the future.
> 
> What do you think we define constants of benchmark IDs? For example
> 
> ```rust
> // Benchmark server parses the benchmark result using this string.
> const BENCHMARK_ID_BYTECODE_CIRCUIT = "BytecodeCircuit";
> ```
> 
> Then in the proof message, we have
> 
> ```rust
> let proof_message = format!("{} Proof generation with {} rows", BENCHMARK_ID_BYTECODE_CIRCUIT, degree);
> ```
> 
> So that future developers wouldn't change the message accidentally and cause the parsing to fail.

Thanks @ChihChengLiang. It makes sense. I will apply your proposal.

ChihChengLiang (approved):
LGTM. Thanks for applying the suggestions.

AronisAt79:
> LGTM. Thanks for applying the suggestions.

Thanks @ChihChengLiang. I pushed a new commit  with fmt changes cause rustfmt ci check was failing. Unless @CPerezz or/and @ed255 would like to also review, i will merge in one hour.

ed255 (approved):
LGTM!

Issue #1212, opened 2023-02-15
Unify circuits params dynamic / static encodings


ed255:
We use circuits parameters to achieve constant size circuits (that is, circuits that use a constant number of rows, where the fixed columns are assigned equally for different witness input lengths).
For some testing scenarios we want to adjust the number of rows used in a circuit to the minimum necessary given the test input, so that the testing time is reduced; in that case we want the number of rows to be calculated dynamically.

Currently the EVM circuit supports the dynamic approach by taking `max_evm_rows = 0`.
On the other hand, the Keccak circuit supports this by taking `keccak_padding = None`.
I think the other circuits don't support a special value to chose the circuit size dynamically.

We should unify the approach for dynamically sizing, for example by having the Kecak circuit support the 0 value as a "dynamic" flag, like the EVM Circuit does.

Alternatively, we could remove the dynamic size support from the circuits itself, and pass a dynamically circuit parameter calculated from the inputs in tests.

See: 
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/7fd693c92dd3f64d4d565098b8f0a380be0f4759/bus-mapping/src/circuit_input_builder.rs#L38-L59

aguzmant103:
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1148 waiting for this

Issue #1209, opened 2023-02-14
Add keccak circuit to integration tests


ed255:
Depends on https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1148 because it refactors the integration tests a bit

anylots:
How many transfer transactions can be included in a proof submission？

ed255:
> How many transfer transactions can be included in a proof submission？

Is this question related to this issue?

Issue #1208, opened 2023-02-14
SuperCircuit integration tests: make it non variadic


ed255:
Currently the SuperCircuit used in the integration tests is variadic depending on the inputs.  We can see this because the check `test_variadic` introduced in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1148 fails, and we also saw this already when trying the SuperCircuit with the real prover with chached proving keys (we observed ConstraintFailures).

I have taken a look at the fixed columns that vary between one run and the other, and they are fixed columns 3 to 10, which correspond to the Keccak Circuit.  This is no surprise, because the integration tests are using the circuits parameters `keccak_padding: None` which assigns the fixed columns of keccak dynamically.  We must change that for a value that works for the different integration tests.  A value of ~7000 should be enough (this is the max number of rows that the keccak fixed columns are assigned in the integration tests, gathered from observation)

leolara:
I think there was a task also to make keccak_padding not an `Option`? But an `int` so it is automatic when `0` like in the EVM circuit? I don't see it now.

ed255:
> I think there was a task also to make keccak_padding not an `Option`? But an `int` so it is automatic when `0` like in the EVM circuit? I don't see it now.

I think there was no issue for that, I've just created it https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1212

aguzmant103:
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1148 waiting for this

ed255:
This issue is probably the same as https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1128

ed255:
This issue is related to the failures in the Integration Testing that we can see in the main branch since https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1173
```
thread 'main' panicked at 'circuit fixed columns are not constant for different witnesses', /home/ubuntu/actions-runner/_work/zkevm-circuits/zkevm-circuits/integration-tests/src/integration_test_circuits.rs:254:17
test mock_prover::serial_test_super_circuit_multiple_transfers_0 ... FAILED

failures:
    mock_prover::serial_test_super_circuit_erc20_openzeppelin_transfer_fail
    mock_prover::serial_test_super_circuit_erc20_openzeppelin_transfer_succeed
    mock_prover::serial_test_super_circuit_multiple_erc20_openzeppelin_transfers
[...]

Issue #1128, opened 2023-02-02
Keccak integration tests should not have keccak_padding: None


leolara:
From https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/967#issuecomment-1410617907

---

by @ed255 :

Also I see that we have keccak_padding: None in the integration tests, we should also change that for an appropriate value to get a constant keccak circuit.

aguzmant103:
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1148 waiting for this

PR #1224, opened 2023-02-17
Keccak integration tests and more

leolara:
### Description

Add keccak to integration tests, and solve some issues that also fix variadic problem for Super Circuit integration test.

### Issue Link

Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1128
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1208
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1209
Closes https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1212

### Type of change

### Contents

- Keccak integration tests should not have keccak_padding: None. 
- SuperCircuit integration tests: make it non variadic.
- Add keccak circuit to integration tests. 
- Unify circuits params dynamic / static encodings. 

### Rationale

In the issues.

### How Has This Been Tested?

Integration test mock prover passes. The integration test actual prover does not pass because of unrelated bug https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1016


leolara:
Although the mock test passes and the fixed columns are the same, the actual prover gives an "ConstraintSystemFailure" error for all blocks with the keccak integration test.

Perhaps it is better to integrate the permutations check to the mock prover to see if that is the reason?

lispc:
oh i think i know the reason... i remember in the upstream keccak circuit, phase 2 cells and phase 1 cells are mixed in same columns.. so the commitment of phase1 columns(used to derive challenges), are not the real commitment that should be used in vanishing poly... so https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1016 is not only a "soundness" issue..      

we have a solution https://github.com/scroll-tech/zkevm-circuits/pull/308 if you think it is ok, we can PR here.

ed255:
> Although the mock test passes and the fixed columns are the same, the actual prover gives an "ConstraintSystemFailure" error for all blocks with the keccak integration test.

> Perhaps it is better to integrate the permutations check to the mock prover to see if that is the reason?

I think that's the best next step!

ed255:
> oh i think i know the reason... i remember in the upstream keccak circuit, phase 2 cells and phase 1 cells are mixed in same columns.. so the commitment of phase1 columns(used to derive challenges), are not the real commitment that should be used in vanishing poly... so #1016 is not only a "soundness" issue..
> 
> we have a solution [scroll-tech#308](https://github.com/scroll-tech/zkevm-circuits/pull/308) if you think it is ok, we can PR here.
> 
> super circuit in scroll fork works well, no problem when real proving

Thanks for sharing this information!

I have just tested the real prover for the keccak integration tests using the branch in this PR, but generating the setup with the block used in the proof and it still fails; so everything checks out and the issue must not be in the permutations but in the Challange API as you pointed out!

I checked that PR and looks good to me to be merged into our repo :)

CPerezz (approved):
LGTM!

Issue #1231, opened 2023-02-22
Circuit for `EXP` OOG error


silathdiir:
### Describe the feature you would like

Suppose to implement `ErrorOOGExpGadget` to handle OOG error of `EXP` opcode.

PR #1232, opened 2023-02-22
Implement circuit for `EXP` OOG error

silathdiir:
### Description

Add `ErrorOOGExpGadget` to handle OOG error of `EXP` opcode.

Gas calculation could reference [this go-ethereum function](https://github.com/ethereum/go-ethereum/blob/master/core/vm/gas_table.go#L351).

Spec Markdown PR https://github.com/privacy-scaling-explorations/zkevm-specs/pull/388

### Issue Link

Close https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1231

### Type of change

### How Has This Been Tested?

Tested with unit-test cases.

### DIFF
--- a/eth-types/src/evm_types.rs
+++ b/eth-types/src/evm_types.rs
@@ -136,7 +136,6 @@ impl GasCost {
     pub const MEMORY_EXPANSION_LINEAR_COEFF: Self = Self(3);
     /// Constant gas for LOG[0-4] op codes
     pub const LOG: Self = Self(375);
-    /// Constant gas for LOG[0-4] op codes
     /// Times ceil exponent byte size for the EXP instruction, EIP-158 changed
     /// it from 10 to 50.
     pub const EXP_BYTE_TIMES: Self = Self(50);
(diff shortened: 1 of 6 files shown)


eth-types/src/evm_types.rs
```diff
@@ -136,6 +136,10 @@ impl GasCost {
     pub const MEMORY_EXPANSION_LINEAR_COEFF: Self = Self(3);
     /// Constant gas for LOG[0-4] op codes
     pub const LOG: Self = Self(375);
+    /// Constant gas for LOG[0-4] op codes
```
roynalnaruto, line 139: This line might have sneaked in by mistake?

zkevm-circuits/src/evm_circuit/execution/error_oog_exp.rs
```diff
@@ -0,0 +1,216 @@
...
+    #[test]
+    fn test_oog_exp() {
+        test_ok(U256::zero());
+        test_ok(U256::one());
+        test_ok(1023.into());
+        test_ok(U256::MAX);
+        test_ok(rand_word());
+    }
+
+    fn test_ok(exponent: U256) {
```
roynalnaruto, line 189: Might be good to also create a test scenario for an internal call that goes OOG. To check for the context restoration

roynalnaruto (commented):
Overall LGTM! How about we also test the internal call scenario? As I see we only test OOG in the case of `is_root == 1`

roynalnaruto (approved):
Looks good!

lispc (approved):

PR #1239, opened 2023-02-23
improve debuggability: better check of rw lookup and rw value

lispc:
### Description

1. in evm circuit, check_rw_lookup only checks rw inside step instead of whole block. Improve running speedup greatly.
2. add a `check_value` method for RwMap. Without running StateCircuit we can quickly find almost problems related to rw::value.

### Issue Link

[_link issue here_]

### Type of change

### Contents

- [_item_]

### Rationale

### How Has This Been Tested?

[_explanation_]

<hr>

## How to fill a PR description 

Please give a concise description of your PR.

The target readers could be future developers, reviewers, and auditors. By reading your description, they should easily understand the changes proposed in this pull request.

MUST: Reference the issue to resolve

### Single responsability

Is RECOMMENDED to create single responsibility commits, but not mandatory.

Anyway, you MUST enumerate the changes in a unitary way, e.g.

```
This PR contains:
- Cleanup of xxxx, yyyy
- Changed xxxx to yyyy in order to bla bla
- Added xxxx function to ...
- Refactored ....
```

### Design choices

RECOMMENDED to:
- What types of design choices did you face?
- What decisions you have made?
- Any valuable information that could help reviewers to think critically

### DIFF
--- a/zkevm-circuits/src/witness/rw.rs
+++ b/zkevm-circuits/src/witness/rw.rs
@@ -4,6 +4,7 @@ use std::collections::HashMap;
 use bus_mapping::operation::{self, AccountField, CallContextField, TxLogField, TxReceiptField};
 use eth_types::{Address, Field, ToAddress, ToLittleEndian, ToScalar, Word, U256};
 use halo2_proofs::circuit::Value;
+use halo2_proofs::halo2curves::bn256::Fr;
 use itertools::Itertools;
 
 use crate::evm_circuit::util::rlc;
@@ -11,6 +12,8 @@ use crate::table::{
     AccountFieldTag, CallContextFieldTag, RwTableTag, TxLogFieldTag, TxReceiptFieldTag,
 };
 use crate::util::build_tx_log_address;
+
+use super::MptUpdates;
 
 /// Rw constainer for a witness block
 #[derive(Debug, Default, Clone)]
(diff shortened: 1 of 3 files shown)


zkevm-circuits/src/witness/rw.rs
```diff
@@ -38,6 +41,61 @@ impl RwMap {
...
+                    // value == prev_value
+                    let prev_value = prev_row.value_assignment::<Fr>(mock_rand);
+
+                    if value != prev_value {
+                        errs.push((idx, err_msg_non_first, *row, *prev_row));
+                    }
+                }
+            }
+        }
+        log::debug!("after rw value check, err num: {}", errs.len());
```
ed255, line 88: Maybe only show this log if `errs.len() > 0`?

zkevm-circuits/src/witness/rw.rs
```diff
@@ -38,6 +41,61 @@ impl RwMap {
...
+        for (idx, err_msg, row, prev_row) in errs {
+            log::debug!(
```
ed255, line 90: Did you consider using `log::error!` instead of `log::debug!` for this line and the previous, so that it shows up as a more important log?

ed255 (approved):
LGTM! This is great, thanks for the PR :) This will be very useful.

I added two small suggestions, but I'm approving already.

han0110 (approved):
LGTM!

Issue #1235, opened 2023-02-23
Seems `fee` should not be reversible in begin tx


silathdiir:
### Describe the bug

Reference go-ethereum code:

1. Seems `fee` is transfered in [buyGas function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L233).

2. [buyGas function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L215) is invoked by [preCheck function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L237).

3. `value` is transfered in [this Call invocation](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L359).

4. Call [get StateDB snapshot](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L184) and [revert snapshot](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L243).

So seems transfer `value` should be reversible but `fee` should not? It is implemented as [transfer_with_fee](https://github.com/scroll-tech/zkevm-circuits/blob/scroll-stable/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L486) and set both as reversible in current zkevm-circuit.

If this is an issue, suppose to fix it in this WIP PR https://github.com/scroll-tech/zkevm-circuits/pull/348.

han0110:
Yeah I think you are right, the fee shouldn't be reversible, so perhaps we need to split the `UpdateBalanceGadget` of `sender` into 2 and one irreversible (fee) and the other reversible (to `receiver`).

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/06e9d64789f0cf57a2c0260f963db7c8d520fd69/zkevm-circuits/src/evm_circuit/util/common_gadget.rs#L350-L355

PR #1242, opened 2023-02-23
Fix `fee` to not reversible in begin tx

silathdiir:
### Description

Reference go-ethereum code as below:

1. `fee` is transferred in [buyGas function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L233).

2. [buyGas function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L215) is invoked by [preCheck function](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L237).

3. `value` is transferred in [this Call invocation](https://github.com/ethereum/go-ethereum/blob/master/core/state_transition.go#L359).

4. Call [gets StateDB snapshot](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L184) and [reverts snapshot](https://github.com/ethereum/go-ethereum/blob/master/core/vm/evm.go#L243) (if error).

It seems that transfer `value` is reversible but `fee` is not.

We set both `fee` and `value` reversible in [transfer_with_fee function](https://github.com/scroll-tech/zkevm-circuits/blob/scroll-stable/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L486) and [TransferWithGasFeeGadget](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/util/common_gadget.rs#L353).

### Issue Link

Close https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1235

### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### Contents

Fix `transfer_with_fee` (bus-mapping) function and `TransferWithGasFeeGadget` (zkevm-circuits) to separate `fee` (not reversible) and `value` (reversible).

### How Has This Been Tested?

Tested with unit test cases of begin tx and other normal ones.

### DIFF
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -473,8 +473,41 @@ impl<'a> CircuitInputStateRef<'a> {
         if !found {
             return Err(Error::AccountNotFound(sender));
         }
-        let sender_balance_prev = sender_account.balance;
-        let sender_balance = sender_account.balance - value - fee;
+        let mut sender_balance_prev = sender_account.balance;
+        debug_assert!(
+            sender_account.balance >= value + fee,
+            "invalid amount balance {:?} value {:?} fee {:?}",
+            sender_balance_prev,
+            value,
+            fee
+        );
+        if !fee.is_zero() {
+            let sender_balance = sender_balance_prev - fee;
+            log::trace!(
+                "sender balance update with fee (not reversible): {:?} {:?}->{:?}",
+                sender,
+                sender_balance_prev,
+                sender_balance
+            );
+            self.push_op(
+                step,
+                RW::WRITE,
+                AccountOp {
+                    address: sender,
+                    field: AccountField::Balance,
+                    value: sender_balance,
+                    value_prev: sender_balance_prev,
+                },
+            );
+            sender_balance_prev = sender_balance;
+        }
+        let sender_balance = sender_balance_prev - value;
+        log::trace!(
+            "sender balance update with value: {:?} {:?}->{:?}",
+            sender,
+            sender_balance_prev,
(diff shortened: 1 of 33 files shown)


lispc (approved):

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -473,8 +473,41 @@ impl<'a> CircuitInputStateRef<'a> {
...
-        let sender_balance = sender_account.balance - value - fee;
+        let mut sender_balance_prev = sender_account.balance;
+        debug_assert!(
+            sender_account.balance >= value + fee,
+            "invalid amount balance {:?} value {:?} fee {:?}",
+            sender_balance_prev,
+            value,
+            fee
+        );
+        if !fee.is_zero() {
```
han0110, line 484: I think we should have the update even `fee` is zero? Otherwise the `rw_counter` won't match

han0110 (commented):
LGTM, but got one question.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
silathdiir: I see. Please let me fix the argument `fee` of `transfer_with_fee` from type `Word` to `Option<Word>`. Since the below `transfer` function calls [this transfer_with_fee](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L514).

han0110: ah i see, using `Option<Word>` should be good.

han0110 (approved):
All LGTM now!

Issue #1233, opened 2023-02-22
RootCircuit AssertitionFailure


pinkiebell:
### Describe the bug

This error message comes up when aggregating SuperCircuit Proof:
`RootCircuit::new: AssertionFailure("e(lhs, g2)·e(rhs, -s_g2) == O")`

Logs: https://github.com/privacy-scaling-explorations/zkevm-chain/tree/prover-error-super-ag-native_withdraw-24af9e3fdd0dcc607ad09b3f920ffd150f7edcfc/errors

### Concrete steps to reproduce the bug

Inside the zkevm-chain repo:
```
docker compose run --rm -it --entrypoint bash dev -c 'COORDINATOR_AGGREGATE_PROOF=true COORDINATOR_CIRCUIT_NAME=super ./scripts/test_prover.sh native_withdraw'
```

pinkiebell:
@han0110 The simple eth transfer now returns [this info via mock prover](https://github.com/privacy-scaling-explorations/zkevm-chain/actions/runs/4261222371/jobs/7415280360#step:5:1105)

PR #1270, opened 2023-03-01
fix tx_accesslist_account_write for warm addresses in BeginTx

pinkiebell:
### Description

Fixes state.tx_accesslist_account_write for warm addresses.

### Issue Link

Fixes: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1233

### Type of change

### How Has This Been Tested?

`cargo test -p zkevm-circuits --features test prover_error -- --nocapture --ignored`
with zkevm-chain fixtures

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -151,7 +151,8 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             tx_id.expr(),
             tx_callee_address.expr(),
             1.expr(),
-            // Freely choosen by prover, not a soundness issue
+            // No extra constraint being used here.
+            // Correctness will be enforced in build_tx_access_list_account_constraints
             is_caller_callee_equal.expr(),
             None,
         );
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -49,6 +49,7 @@ pub(crate) struct BeginTxGadget<F> {
     is_empty_code_hash: IsEqualGadget<F>,
     caller_nonce_hash_bytes: [Cell<F>; N_BYTES_WORD],
     create: ContractCreateGadget<F, false>,
+    is_caller_callee_equal: IsEqualGadget<F>,
 }
 
 impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
@@ -145,11 +146,13 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             0.expr(),
             None,
         );
+        let is_caller_callee_equal =
+            IsEqualGadget::construct(cb, tx_caller_address.expr(), tx_callee_address.expr());
         cb.account_access_list_write(
             tx_id.expr(),
             tx_callee_address.expr(),
             1.expr(),
-            0.expr(),
+            is_caller_callee_equal.expr(),
             None,
         );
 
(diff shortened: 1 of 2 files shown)


zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -145,11 +146,13 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             0.expr(),
             None,
         );
+        let is_caller_callee_equal =
```
pinkiebell, line 149: Don't know if this is the correct way to do it.
cc/ @han0110

han0110: Ah right, never thought about tx to self 🤦‍♂️

I think this would work, but to make it even easier we could have the line 155 to take a witness freely chosen by prover, because we don't really use its previous value here (we only need to make sure it's updated to `1`), but we still need to assign a valid value, otherwise the State circuit consistency check will fail (perhaps one of cause of https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1233)

pinkiebell: I can remove the IsEqualGadget if it doesn't impact soundness of the circuit 👌

pinkiebell: Is there a easy way to do a equality check on the Cell/Expression types? 😅

han0110: you mean to check if 2 cells are identical as `Expression` structure? but why are we doing that here? i think we could do a `query_bool` for this witness.

pinkiebell: I looked at the constraint builder but I don't grok how `query_bool` would return callee==caller. I guess I have to lookup these values beforehand? Anyhow, given that the prover is free to choose those values I opted into select::expr given that 'looks' simpler. Please correct me if I'm going the wrong way :)

pinkiebell: nvm. After being refreshed, I now see my misunderstanding 😆

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -145,11 +146,13 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
             None,
         );
+        let is_caller_callee_equal = cb.query_bool();
         cb.account_access_list_write(
             tx_id.expr(),
             tx_callee_address.expr(),
             1.expr(),
-            0.expr(),
+            // Freely choosen by prover, not a soundness issue
+            is_caller_callee_equal.expr(),
```
han0110, line 155: Strictly speaking it's not that freely chosen by prover, the prover still needs to pass the State circuit consistency check, so we don't need extra constraint here.

If `caller == callee`, then the consistency check will enforce `is_caller_callee_equal == 1`.

If `caller != callee`, then the State circuit initial value's rule will enforce `is_caller_callee_equal == 0`.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/30a2b10f57e508948fb12b3a7c3f63fdeb174df2/zkevm-circuits/src/state_circuit/constraint_builder.rs#L315-L318

han0110 (approved):
LGTM!

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
pinkiebell: I will update the comment 👍

pinkiebell:
I can confirm that this fixes #1233 . Enabled auto-merge

PR #1275, opened 2023-03-02
fix get_step_error for tx deploy case

### LINKED ISSUE #1267: get_step_err fail to get create error in deploy tx
### What command(s) is the bug in?

_No response_

### Describe the bug

current `get_step_err` can only get create related error for internal create/create2 cases in https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/input_state_ref.rs#L1132
but for tx deploy(tx.to = null)(which result in root call context) return nothing from following matching code.
```
  // When last step has opcodes that halt, there's no error.
        if matches!(next_step, None)
            && matches!(
                step.op,
                OpcodeId::STOP | OpcodeId::RETURN | OpcodeId::REVERT | OpcodeId::SELFDESTRUCT
            )
        {
            return Ok(None);
        }
```

DreamWuGit:
### Description

fix get_step_error for tx deploy case
### Issue Link

try to fix #1267 
### Type of change

### How Has This Been Tested?
add two tx deploy trace tests

This PR contains:
- fix `get_step_err` to return tx deploy error.
- add two tx deploy trace tests to verify fix

### DIFF
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -1070,12 +1070,15 @@ impl<'a> CircuitInputStateRef<'a> {
             return Ok(Some(ExecError::InvalidOpcode));
         }
 
+        let call = self.call()?;
         // When last step has opcodes that halt, there's no error.
         if matches!(next_step, None)
             && matches!(
                 step.op,
                 OpcodeId::STOP | OpcodeId::RETURN | OpcodeId::REVERT | OpcodeId::SELFDESTRUCT
             )
+            && !call.is_create()
+        // if it is tx deploy create, return also can cause err
         {
             return Ok(None);
         }
@@ -1085,7 +1088,6 @@ impl<'a> CircuitInputStateRef<'a> {
             .map(|s| s.stack.last().unwrap_or_else(|_| Word::zero()))
             .unwrap_or_else(Word::zero);
 
-        let call = self.call()?;
         let call_ctx = self.call_ctx()?;
         // get value first if call/create
         let value = match step.op {
@@ -1129,7 +1131,7 @@ impl<'a> CircuitInputStateRef<'a> {
                 });
             } else {
                 // Return from a {CREATE, CREATE2} with a failure, via RETURN
-                if !call.is_root && call.is_create() {
+                if call.is_create() {
                     let offset = step.stack.nth_last(0)?;
                     let length = step.stack.nth_last(1)?;
                     if length > Word::from(0x6000u64) {
--- a/bus-mapping/src/circuit_input_builder/tracer_tests.rs
+++ b/bus-mapping/src/circuit_input_builder/tracer_tests.rs
@@ -101,6 +101,32 @@ fn mock_internal_create() -> Call {
     }
 }
 
+fn mock_root_create() -> Call {
+    Call {
+        call_id: 0,
+        caller_id: 0,
+        last_callee_id: 0,
+        kind: CallKind::Create,
+        is_static: false,
+        is_root: true,
+        is_persistent: false,
+        is_success: false,
+        rw_counter_end_of_reversion: 0,
+        caller_address: *ADDR_A,
+        address: *ADDR_B,
+        code_source: CodeSource::Memory,
+        code_hash: Hash::zero(),
+        depth: 1,
+        value: Word::zero(),
+        call_data_offset: 0,
+        call_data_length: 0,
+        return_data_offset: 0,
+        return_data_length: 0,
+        last_callee_return_data_offset: 0,
+        last_callee_return_data_length: 0,
+    }
+}
+
 //
 // Geth Errors ignored
 //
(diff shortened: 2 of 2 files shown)


bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -1070,12 +1070,19 @@ impl<'a> CircuitInputStateRef<'a> {
             return Ok(Some(ExecError::InvalidOpcode));
         }
 
+        let call = self.call()?;
         // When last step has opcodes that halt, there's no error.
         if matches!(next_step, None)
-            && matches!(
+            && (matches!(
```
hero78119, line 1076: suggest below change for better readiblity 🚀 
```rust
if matches!(next_step, None) { // enumerating root call scope successful cases
       // case 1: root call with normal halt opcode termination
           if matches!(step.op, OpcodeId::STOP | OpcodeId::REVERT | OpcodeId::SELFDESTRUCT) {
               return Ok(None);
           }
       // case 2: root call is NOT Create (Create represented by empty tx.to) and halt by opcode::Return 
          if !call.is_create() && step.op == OpcodeId::RETURN {
              return Ok(None);
          }
       // more other case...
}
```

Side question: is `call.is_root == true` equivalent to `matches!(next_step, None)` ? if yes then also success then I vote `call.is_root == true` for better  readibility

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -1142,10 +1148,7 @@ impl<'a> CircuitInputStateRef<'a> {
                     } else if Word::from(200u64) * length > Word::from(step.gas.0) {
                         return Ok(Some(ExecError::CodeStoreOutOfGas));
                     } else {
-                        return Err(Error::UnexpectedExecStepError(
```
hero78119, line 1145: I think we should keep this UnexpectedExecStepError which is more conservative. (I tried rollback this change and all the test cases still pass)

hero78119 (changes requested):
Thanks for the changes!! just have comments regarding to readability 🐔

bus-mapping/src/circuit_input_builder/input_state_ref.rs
DreamWuGit: good comment， for `call.is_root == true` equivalent to `matches!(next_step, None)`, I think it is not aways, even though for internal call(`call.is_root == false`) next_step is not none in most cases.   there is no ensure for that.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
DreamWuGit: @hero78119  thanks for your try. in my testing, all trace tests can pass, but for `begin_tx_deploy_nonce`* test  in begin_tx.rs fail due to this line.   before make the change, they pass because ` return Ok(None)` when hit "RETURN" opcode. after this pr's change, success cases will hit L1145.  let me know if you got different result.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
DreamWuGit: add comment in [eaae0e0](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1275/commits/eaae0e01abde0b696fa97313def1bb14f0ef8b75) , removed root call from comment.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
hero78119: Ic. how about this way, by just move this explicitly whitelist condition as case 3
```rust
               // case 3 
                if call.is_create() && call.is_success() && step.op == OpcodeId::RETURN {
                return Ok(None);
            }
```
and need to add condition here to mark CREATE/CREATE2 call `is_success` or not
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/transaction.rs#L50-L65

With that to avoid adding a `Ok(None)` to capture all other cases. It also separate the success conditions with failed condition clearly in codebase. Besides, explicity success is always better, and left other cases is unknown error, wdyt~?

bus-mapping/src/circuit_input_builder/input_state_ref.rs
hero78119: Thanks for the clarification, I learned it still possible that got edge cases for internal call with `next_step = None`

bus-mapping/src/circuit_input_builder/input_state_ref.rs
DreamWuGit: > Ic. how about this way, by just move this explicitly whitelist condition as case 3
> 
> ```rust
>                // case 3 
>                 if call.is_create() && call.is_success && step.op == OpcodeId::RETURN {
>                 return Ok(None);
>             }
> ```

sound good to me, will have a try !!

DreamWuGit: updated in [93b2a24](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1275/commits/93b2a247623ee962ea185f3a2ea7b5008380eccc)  thanks for the suggestion!

DreamWuGit: > Update: I think `is_success` is handled correct for create/create2 op: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/circuit_input_builder/transaction.rs#L60-L63

that is correct !

hero78119 (approved):
LGTM and nice fix! thanks 💯

PR #1279, opened 2023-03-02
subcircuit trait for PiTestCircuit

pinkiebell:
### Description

This implements the `SubCircuit` trait for `PiTestCircuit`. PiTestCircuit isn't a subcircuit in technical terms and hence panics when `synthesize_sub` is called with a message to use the `PiCircuit` for embedding instead.

### Type of change

### Rationale

After moving some code of macro bodies I had to update the types and zkevm-chain now requires the `SubCircuit` trait being implemented for all circuits used. Technically, PiTestCircuit isn't a Subcircuit. But without creating a new trait it makes downstream usage more streamlined.

### How Has This Been Tested?

In zkevm-chain

### DIFF
--- a/zkevm-circuits/src/pi_circuit.rs
+++ b/zkevm-circuits/src/pi_circuit.rs
@@ -1456,12 +1456,39 @@ pub struct PiTestCircuit<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usi
 );
 
 #[cfg(any(feature = "test", test, feature = "test-circuits"))]
-impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize>
-    PiTestCircuit<F, MAX_TXS, MAX_CALLDATA>
+impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize> SubCircuit<F>
+    for PiTestCircuit<F, MAX_TXS, MAX_CALLDATA>
 {
+    type Config = PiCircuitConfig<F>;
+
+    fn new_from_block(block: &witness::Block<F>) -> Self {
+        let mut block = block.clone();
+        block.circuits_params.max_txs = MAX_TXS;
+        block.circuits_params.max_calldata = MAX_CALLDATA;
+
+        Self(PiCircuit::new_from_block(&block))
+    }
+
+    fn min_num_rows_block(block: &witness::Block<F>) -> (usize, usize) {
+        let mut block = block.clone();
+        block.circuits_params.max_txs = MAX_TXS;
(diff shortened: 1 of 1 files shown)


zkevm-circuits/src/pi_circuit.rs
```diff
@@ -1456,13 +1456,40 @@ pub struct PiTestCircuit<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usi
...
+        block.circuits_params.max_txs = MAX_TXS;
+        block.circuits_params.max_calldata = MAX_CALLDATA;
+
+        Self(PiCircuit::new_from_block(&block))
+    }
+
+    fn min_num_rows_block(block: &witness::Block<F>) -> (usize, usize) {
+        let mut block = block.clone();
+        block.circuits_params.max_txs = MAX_TXS;
+        block.circuits_params.max_calldata = MAX_CALLDATA;
```
ed255, line 1475:
```suggestion
        assert_eq!(block.circuits_params.max_txs, MAX_TXS);
        assert_eq!(block.circuits_params.max_calldata, MAX_CALLDATA);
```
With this you could also skip the clone.

zkevm-circuits/src/pi_circuit.rs
```diff
@@ -1456,13 +1456,40 @@ pub struct PiTestCircuit<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usi
...
-    PiTestCircuit<F, MAX_TXS, MAX_CALLDATA>
+impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize> SubCircuit<F>
+    for PiTestCircuit<F, MAX_TXS, MAX_CALLDATA>
 {
+    type Config = PiCircuitConfig<F>;
+
+    fn new_from_block(block: &witness::Block<F>) -> Self {
+        let mut block = block.clone();
+        block.circuits_params.max_txs = MAX_TXS;
+        block.circuits_params.max_calldata = MAX_CALLDATA;
```
ed255, line 1467:
```suggestion
        assert_eq!(block.circuits_params.max_txs, MAX_TXS);
        assert_eq!(block.circuits_params.max_calldata, MAX_CALLDATA);
```
We should always use the same values for circuit parameters in `circuit_params` and generic const of circuit types, so I think it's better to assert that this is true rather than overwrite the values, because it enforces us to follow the rule of using the same parameter values in types and values.  Would this be a problem in your scenario?
With this you could also skip the clone.

ed255 (approved):
LGTM!

Issue #1250, opened 2023-02-24
bug: returndatasize after successful creation should be 0


lispc:
### Describe the bug

according to https://eips.ethereum.org/EIPS/eip-211， “As an exception, CREATE and CREATE2 are considered to return the empty buffer in the success case and the failure data in the failure case.”

we did not handle this now

hero78119:
it's blocked by issues [Implement CREATE & CREATE2](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1130) 

After that, also need to check logic in bus mapping

PR #1283, opened 2023-03-03
follow EIP-211: create/create2 successful case must set empty returndatasize

hero78119:
### Description

Align specs in EIP-211: returndatasize should be 0 after create/create2 success call

### Issue Link

Closed https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1250

### Type of change

### Contents

- revamp `handle_restore_context` and `handle_return` to be able to handle create/create2 returndatasize 0 case
- add unittest to assure 1. create/create2 contract created successfully 2. bytecode with returndatasize/returndatacopy satisfy constraints
- add few comments for existing hardcode magic number

### Rationale

use magic number to access rws indices during assignment is inevitable. We might need a better design to address this issue.
One possible way is refactor, by bumping and returning index when every time calling rw_lookup, and exposing index value into XXXXXGadget like other `Cell` fields

### How Has This Been Tested?

`cargo test --package zkevm-circuits --lib --features test --features warn-unimplemented -- evm_circuit::execution::return_revert::test --nocapture`

<hr>

### DIFF
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -883,11 +883,8 @@ impl<'a> CircuitInputStateRef<'a> {
     pub fn handle_return(&mut self, step: &GethExecStep) -> Result<(), Error> {
         // handle return_data
         let (return_data_offset, return_data_length) = {
-            let call = self.call()?;
-            if !call.is_root {
+            if !self.call()?.is_root {
                 let (offset, length) = match step.op {
-                    // EIP-211 CREATE/CREATE2 call successful case should set RETURNDATASIZE = 0
-                    OpcodeId::RETURN if call.is_create() && call.is_success => (0, 0),
                     OpcodeId::RETURN | OpcodeId::REVERT => {
                         let offset = step.stack.nth_last(0)?.as_usize();
                         let length = step.stack.nth_last(1)?.as_usize();
@@ -941,8 +938,24 @@ impl<'a> CircuitInputStateRef<'a> {
         // If current call has caller.
         if let Ok(caller) = self.caller_mut() {
             caller.last_callee_id = call.call_id;
-            caller.last_callee_return_data_length = return_data_length;
-            caller.last_callee_return_data_offset = return_data_offset;
+            // EIP-211 CREATE/CREATE2 call successful case should set RETURNDATASIZE = 0
+            if step.op == OpcodeId::RETURN && call.is_create() && call.is_success {
+                caller.last_callee_return_data_length = 0u64;
+                caller.last_callee_return_data_offset = 0u64;
+            } else {
+                caller.last_callee_return_data_length = return_data_length;
+                caller.last_callee_return_data_offset = return_data_offset;
+            }
+        }
+
+        // If current call has caller_ctx (has caller)
+        // EIP-211 CREATE/CREATE2 call successful case should set RETURNDATASIZE = 0
+        if let Ok(caller_ctx) = self.caller_ctx_mut() && (
+           step.op == OpcodeId::RETURN &&
+           call.is_create() &&
+           call.is_success
+        ) {
+            caller_ctx.return_data.truncate(0);
         }
 
         self.tx_ctx.pop_call_ctx();
@@ -972,8 +985,6 @@ impl<'a> CircuitInputStateRef<'a> {
 
         let [last_callee_return_data_offset, last_callee_return_data_length] = match geth_step.op {
             OpcodeId::STOP => [Word::zero(); 2],
-            // EIP-211 CREATE/CREATE2 call successful case should set RETURNDATASIZE = 0
-            OpcodeId::RETURN if call.is_create() && call.is_success => [Word::zero(); 2],
             OpcodeId::REVERT | OpcodeId::RETURN => {
                 let offset = geth_step.stack.nth_last(0)?;
                 let length = geth_step.stack.nth_last(1)?;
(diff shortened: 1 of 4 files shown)


zkevm-circuits/src/evm_circuit/execution/return_revert.rs
```diff
@@ -97,6 +97,15 @@ impl<F: Field> ExecutionGadget<F> for ReturnRevertGadget<F> {
             is_create.clone() * is_success.expr() * not::expr(copy_rw_increase_is_zero.expr());
         let (caller_id, address, reversion_info, code_hash) =
             cb.condition(is_contract_deployment.clone(), |cb| {
+                // EIP-211 RETURNDATASIZE should be 0 for CREATE/CREATE(2)
+                for (field_tag, value) in [
+                    (CallContextFieldTag::LastCalleeReturnDataOffset, 0.expr()),
+                    (CallContextFieldTag::LastCalleeReturnDataLength, 0.expr()),
+                ] {
+                    // let cell = self.query_cell_with_type(phase);
```
han0110, line 105: leftover?

han0110 (commented):
LGTM! Nice fix!

han0110 (approved):

zkevm-circuits/src/evm_circuit/execution/return_revert.rs
hero78119: y! just clean it and thanks for the catch up!

lispc:
suggest add a testcase for this?

lispc (changes requested):
1. the case A/B/C is not mutual exclusive. I remember the LastCalleeReturnDataOffset is already set inside "handle_restore_context" in case C branch.  So even though this PR is correct(not sure), the LastCalleeReturnDataOffset/LastCalleeReturnDataLength is set twice? Not efficient.
2. the tests "evm_circuit::execution::return_revert::test" works well before this PR, and works well after this PR. So not least it is not obvious that something is fixed by this PR. I suggest either (A) add a new testcase, where RETURNDATASIZE opcode is after a successful CREATE (2) modify current test case, put a RETURNDATASIZE after the successful CREATE
3. in fact, i am afraid this RP is not correct. LastCalleeReturnDataOffset/LastCalleeReturnDataLength set twice, and the second time it is set to an incorrect value.

han0110:
> 3. in fact, i am afraid this RP is not correct. LastCalleeReturnDataOffset/LastCalleeReturnDataLength set twice, and the second time it is set to an incorrect value.

You are right, didn't notice it's set twice. Perhaps we could move the EIP-211 logics into `handle_restore_context`? Also the `handle_return` seems also need to be applied with the rule.

Second that adding some tests for these cases will be nice for this PR.

hero78119:
Y, working on it and will try to add test to cover those edge case, thanks for catching up the problem and suggestion!

hero78119 (commented):
Hi @lispc @han0110 , should be handle correctly now. please kindly help review again, thank you!

Besides, adding unittest with CREATE/CREATE2 + RETURNDATASIZE/RETURNDATACOPY bytecode to assure
1. RETURNDATASIZE assure `handle_restore_context ` handle correctly.
2. RETURNDATACOPY assure `handle_return` assign value correctly.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -940,6 +940,21 @@ impl<'a> CircuitInputStateRef<'a> {
             caller.last_callee_id = call.call_id;
             caller.last_callee_return_data_length = return_data_length;
             caller.last_callee_return_data_offset = return_data_offset;
+            // EIP-211 CREATE/CREATE2 call successful case should set RETURNDATASIZE = 0
+            if step.op == OpcodeId::RETURN && call.is_create() && call.is_success {
+                caller.last_callee_return_data_length = 0u64;
+                caller.last_callee_return_data_offset = 0u64;
+            }
+        }
+
```
lispc, line 949: what about 

```
           if step.op == OpcodeId::RETURN && call.is_create() && call.is_success {
                caller.last_callee_return_data_length = 0u64;
                caller.last_callee_return_data_offset = 0u64;
            } else {
                caller.last_callee_return_data_length = return_data_length;
                caller.last_callee_return_data_offset = return_data_offset;
            }
```


more clear.

lispc (approved):

lispc:
need to polish the "XXX: please pay attention to few hardcode number" before merging

bus-mapping/src/circuit_input_builder/input_state_ref.rs
hero78119: 👍 all done

hero78119:
> need to polish the "XXX: please pay attention to few hardcode number" before merging

y, i removed this comments entirely

zkevm-circuits/src/evm_circuit/execution/return_revert.rs
```diff
@@ -571,4 +576,72 @@ mod test {
...
+            .enumerate()
+            .filter(|(_, s)| s.op == OpcodeId::RETURN)
+            .flat_map(|(index, _)| block.geth_traces[0].struct_logs.get(index + 1))
+            .flat_map(|s| s.stack.nth_last(0)) // contract addr on stack top
+            .collect_vec();
+        assert!(created_contract_addr.len() == 2); // both contract addr exist
+        created_contract_addr
+            .iter()
+            .for_each(|addr| assert!(addr > &U256::zero()));
+
```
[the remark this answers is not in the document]
hero78119: y it make senses, added

han0110 (approved):
LGTM!

Issue #1118, opened 2023-01-30
more account lookup consistency issues


lispc:
(wip, write some observations, still debugging and fixing)

1. in current begin_tx, we first do transfer than do codehash reading for non-exist check. Is this correct? If transfer.value is non zero, then it will make a non existed account to become an existed account so codehash reading will be incorrect?
2. current Account::is_empty relys on 'store_hashmap.is_empty()'. But in evm, for storage, non-exsit is same with 0. So in fact we need to check 'store_hashmap.values().all( | value | value.is_zero())'. Or a better solution, the the rpc-statedb-building phase, just skip inserting 0 storage values inside statedb
3. when the dst of a transfer is non exist, after this transfer, the account no longer is non exist. Then, from state circuit side, prev_row is Account::CodeHash { is_write: false, prev: 0, value: 0}, next_row is Account::CodeHash { is_write: false, prev: keccak(nil), value: keccak(nil) }. State circuit will be broken..

ed255:
>     1. in current begin_tx, we first do transfer than do codehash reading for non-exist check. Is this correct? 

This is the current implementation, yes.  But this is not how it should be!  We should read the codehash first before touching possibly non-existing accounts.  I added a check to verify this behavior in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1092 but obviously it's being skipped here. After a quick check I realized that the check was added in `push_op` but all writes to accounts are done via `push_op_reversible`!  So we should add the check to `push_op_reversible` as well.

> If transfer.value is non zero, then it will make a non existed account to become an existed account so codehash reading will be incorrect? ( i fixed this by changing transfer and codehash reading order)

So there are 2 things to solve here:
1. First check code_hash and determine if account exists or not
2. If account doesn't exist and transfer value > 0, we should write the code_hash to `EMPTY_HASH` before applying the transfer (setting the code_hash to `EMPTY_HASH` will mark the account as existing for the following steps).

[...]

PR #1286, opened 2023-03-06
Fix: account lookup various inconsistencies

ed255:
### Description

Feature: extend `check_rw_lookup` with more checks
Fix several account lookups inconsistencies so that these rules are followed:
For Execution Steps that may create an account via transfer (this is BeginTx to empty account,  call{CALL, CALLCODE}):
- 1. Read callee code_hash to determine `is_empty = code_hash == 0`
- 2. if `balance > 0 & is_empty` then `write calle.code_hash 0 -> EMPTY_CODE_HASH` to create the callee account
- 3. if `balance > 0` then perform transfer

### Issue Link

- https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1118
- Investigate if this is now resolved https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1132

### Type of change

### Contents

- github workflows:
    - Set `RUST_MIN_STACK=16777216` env var to avoid stack overflows that cause some tests to get killed.
- bus-mapping:
    - Apply account lookup validity checks also via `push_op_reversible`
    - Remove the `rw` argument from `push_op_reversible` and assume it to be `RW::WRITE`
    - Remove the `value_prev` in `account_read` and assume it to be `value` (a read should have `value_prev = value`)
    - Add fields to ExecStep to help debugging and sanity checking: `reversible_write_counter_delta` and `copy_rw_counter_delta`.
    - Update Transfer assignment logic to conditionally create the receiver account, and to skip the transfer if the value is 0.
    - Reorder rw operations in BeginTx, Create to be consistent
    - Remove unused rw operations in ReturnRevert
- zkevm-circuits - EVM Circuit:
    - support conditional multiple reversible rw ops.  Previously the counter for reversible rw ops was a static expression, so the opcode could only do a static number of reversible operations.
    - Update TransferGadget logic to conditionally create the receiver account, and to skip the transfer if the value is 0.
    - Reorder rw operations in BeginTx to be consistent
    - Extend `check_rw_lookup` to also check the following:
        - reversion rw operations
        - number of rw operations from bus-mapping in the step equal the number ot rw lookup assignments + copy rws
    - In BeginTx, track the rw_deltas done by the TransferGadget via the Gadget helper functions
    - Fix: ReturnRevert, remove 3 extra rw_delta for case C that were not used by the ExecStep.
    - Increase `N_PHASE2_COLUMNS` from 3 to 4 to accommodate new changes
    - Introduce `StepRws` utility struct to help read rw operations from steps sequentially
    - Support nested conditions in ConstraintBuilder


bus-mapping/src/evm/opcodes.rs
```diff
@@ -425,6 +416,25 @@ pub fn gen_begin_tx_ops(state: &mut CircuitInputStateRef) -> Result<ExecStep, Er
...
+        );
+    }
+
+    // Transfer with fee
+    state.transfer_with_fee(
+        &mut exec_step,
+        call.caller_address,
+        call.address,
+        callee_exists,
+        call.is_create(),
```
lispc, line 434: i think even though is_create == true, must_create may not be true?   A user can transfer eth to a pre-calculated address before he really deploy contract at that address

lispc (approved):
great!

bus-mapping/src/evm/opcodes/callop.rs
```diff
@@ -156,16 +154,17 @@ impl<const N_ARGS: usize> Opcode for CallOpcode<N_ARGS> {
             call.caller_address,
             AccountField::Balance,
             caller_balance,
-            caller_balance,
         );
 
-        // Transfer value only for CALL opcode, insufficient_balance = false
-        // and value > 0.
-        if call.kind == CallKind::Call && !insufficient_balance && !call.value.is_zero() {
+        // TODO: What about transfer for CALLCODE?
```
han0110, line 159: When encountering `CALLCODE` the account should be deployed or deploying itself, for the the first case the account exists, and the latter the account is also created after this PR, so it seems correct to only do the transfer when it's `CALL`.

bus-mapping/src/evm/opcodes.rs
han0110: Kimi also thought about this case when implementing `CREATE`, and we discussed yesterday then decide to update the account `code_hash` to `EMPTH_HASH` no matter what, but let the prover witness the previous value, and constrain it shuold be either `0` or `EMPTH_HASH`

(or in another way we could ask prover the witness `callee_exists` to select `EMPTH_HASH` or `0` as previous value).

han0110 (approved):
LGTM!

lispc:
what about merging this PR now? and solve the "deploy at existed address" issue later?  Since it is a already big PR and brings some breaking and useful changes. Merging it early sounds better overall.  @han0110

han0110:
> what about merging this PR now? and solve the "deploy at existed address" issue later?

totally agree!

lispc:
i noticed the "must_create" is not consistent in bus-mapping(tx.is_create) VS circuit(tx.is_create || callee_not_exists) ?

han0110:
yeah it seems so

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/fc7b3d546cb00639b7ffe9328a1dea5eeb1405f0/bus-mapping/src/evm/opcodes.rs#L429-L437

vs 

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/fc7b3d546cb00639b7ffe9328a1dea5eeb1405f0/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs#L182-L191

Issue #1287, opened 2023-03-07
bug: CI  test with invalid memory reference


DreamWuGit:
### Describe the bug

found this issue in PR #1265 's  CI https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/4349281702/jobs/7598772747
```
Caused by:
  process didn't exit successfully: `/github/runner/_work/zkevm-circuits/zkevm-circuits/target/release/deps/testool-9af89d4775d0c804` (signal: 11, SIGSEGV: invalid memory reference)
Error: The process '/home/ubuntu/.cargo/bin/cargo' failed with exit code 101
```

Managed to resolve it in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1265/commits/edc24ab251b35c5bc5113555b35e4ce64ab13cf6 
by change `EvmCircuitCached` to `EvmCircuit`。

lispc:
@ed255 the Cached EvmCircuit is a bit strange..

lispc:
oh it seems related to RUST_MIN_STACK: 16777216..

ChihChengLiang:
`RUST_MIN_STACK: 16777216` is introduced in #1286. cc @ed255

ed255:
Ah yes, I've observed this previously.  Whenever this happens `signal: 11, SIGSEGV: invalid memory reference` running our code, it's most likely a stack overflow.  After reading some Rust documentation I found that Rust threads start with 2 MB of stack by default, and this is sometimes not enough for us.  Sometimes I've encountered this issue when running the tests in the test profile instead of release profile.  Recently I encountered it in release profile while developing the PR that CC linked https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1286 so I added an env var to increase the stack size in github actions.
I didn't find a better way to set the default stack size other than using the env var.

My guess is that the big stack requirements comes from having the `ExecutionConfig` from the EVM Circuit type so big.  This type is held under a Box in the ` EvmCircuitConfig` to try to mitigate this problem.  But maybe it's when this value is created that it's held in the stack and can cause the overflow.
Notice that the size of `ExecutionConfig` keeps growing with every execution state that we add, and with every new gadget that we add to an existing execution state.

ed255:
I wish we could find a way to resolve the massive stack usage, so that we don't have to depend on setting a bigger stack via `16777216`!

lispc:
[...]

PR #1294, opened 2023-03-08
fix: reduce stack consumption in EVM Circuit

ed255:
### Description

In EVM Circuit's ExecutionConfig, store the gadget for every execution state in a Box to significantly reduce the size of ExecutionConfig which inevitably needs to live in the stack at some points.  This PR removes the extended stack setting `RUST_MIN_STACK: 16777216` introduced in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1286

Let's see if this works!

### Issue Link

Related to https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1287

### Type of change

### Rationale

Even though setting the env var `RUST_MIN_STACK=16777216` fixes the stack overflow we encountered before, I think it's better to resolve this by avoiding consuming so much stack that it overflows on its default size.  This is because setting this env var can be annoying and it's easy to forget it in many scenarios.


hero78119:
LGTM! Something interested is should we also introduce some way to profile it, and maybe integrate it into `make stack_profiling` part of CI build ?
this [post on SO](https://stackoverflow.com/questions/38929524/how-to-measure-a-functions-stack-usage-in-rust) share an interesting asm although hacky but looks controlable overhead comparing with external tool.

ed255:
> LGTM! Something interested is should we also introduce some way to profile it, and maybe integrate it into `make stack_profiling` part of CI build ? this [post on SO](https://stackoverflow.com/questions/38929524/how-to-measure-a-functions-stack-usage-in-rust) share an interesting although hacky but looks controlable overhead comparing with external tool.

From the post you linked, this tool looks quite interesting https://crates.io/crates/cargo-call-stack !  I think it's worth exploring :)

han0110 (approved):
LGTM!

lispc (approved):

hero78119:
> > LGTM! Something interested is should we also introduce some way to profile it, and maybe integrate it into `make stack_profiling` part of CI build ? this [post on SO](https://stackoverflow.com/questions/38929524/how-to-measure-a-functions-stack-usage-in-rust) share an interesting although hacky but looks controlable overhead comparing with external tool.
> 
> From the post you linked, this tool looks quite interesting https://crates.io/crates/cargo-call-stack ! I think it's worth exploring :)

Y, just the `cargo-call-stack` produce dot diagram.  The SO shared some assembly code way + `Tick!` way looks lightweight   We can explore more if the stack problem still there after this fix :)

PR #1290, opened 2023-03-07
add event trigger: workflow_dispatch

AronisAt79:
### Description
With this PR, we enable on demand execution of the integration-tests workflow, by adding the workflow_dispatch event trigger. 

### Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Rationale
Adding the possibility to manually trigger the workflow, acts as an enabler for integrations tests with real prover. Due to high resource requirements and extended duration of integration tests with real prover, it makes sense to not run the wf on each push. Along with scheduled executions (ie once per week), we provide the option to manually trigger a parametrized wf execution (select development branch, select type of prover etc).

* Scheduler and wf parametrization is not part of this PR

```
This PR contains:
- Modification of integration.yml, add the workflow_dispatch event trigger 
```

PR #1297, opened 2023-03-08
add scheduler, add conditional parameter provertype, add workflow_dis…

AronisAt79:
### Description
As a follow up to #1290, we add required input choices for manual trigger (workflow_dispatch), as a means to define whether we want to invoke mock or real prover and select target branch.
We introduce environment variable [provertype] 
We also add a scheduler to trigger integration tests every Sunday at 01:50AM. Scheduled job will default to real_prover.
Consecutiveness job is also now enforced on scheduled events and workflow_dispatch

### Type of change

- [ X] New feature (non-breaking change which adds functionality)

### Contents

- Changed File: .github/workflows/integration.yml

### Rationale
Overall, we want integration tests to be invoked on 4 distinct occasions:
1. Ready for Review PRs, labelled as 'trigger-integration-tests'
2. On branch push to main
3. On a predefined interval (once a week on offline hours for now)
4. On demand

1 and 2 are always executed with mock object to save time and cloud resources
3 is always executed with real_prover
4 allows selection of both prover type and branch to execute integration tests against. 

### Design choices

Due to gh actions not supporting ternary operations on environment variables
for example:
    if github.event == pull_request:
        env.provertype="mock_prover"
    elif github.event == schedule:
        env.provertype="real_prover"

we introduce new job steps to conditionally inject ${{ env.provertype }} to $GITHUB_ENV, for example:
    - name: Set provertype to workflow_dispatch choice
        run: |
          echo "provertype=${{ inputs.provertype }}" >> $GITHUB_ENV
        if: github.event_name == 'workflow_dispatch'

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,16 +1,30 @@
 name: Integration Tests
 
 on:
+  schedule:
+    - cron: '50 1 * * SUN'
   pull_request:
     types: [synchronize, labeled]
   push:
     branches:
       - main
-  workflow_dispatch:      
+  workflow_dispatch:    
+    inputs:
+      provertype:
+        description: 'invoke real vs mock prover'
+        required: true
+        default: 'mock_prover'
+        type: choice
+        options:
+        - real_prover
+        - mock_prover  
 
 jobs:
   consecutiveness:
-    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'trigger-integration-tests')
+    if: github.event_name == 'push' ||
+        github.event_name == 'workflow_dispatch' ||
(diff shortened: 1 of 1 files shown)


AronisAt79:
NOTE: Integrations tests with real_prover will still most probably fail atm , due to insufficient system resources. A new PR will follow to add a cloud host with sufficient RAM.

AronisAt79:
> As I user what would I need to do to trigger the real prover?

Go to actions and on the "All workflows" bar on the left, select Integration Tests. Then you get the option to manually trigger the wf porovidinf prover type and target branch as input variables.

image

leolara (approved):
LGTM!! 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

Issue #1298, opened 2023-03-09
incorrect ExecutionState::halts_in_exception


lispc:
### Describe the bug

ErrorDepth / ErrorInsufficientBalance / ErrorContractAddressCollision, is not halt? They did nothing

han0110:
I think this issue is similar to this one https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/985 right?

For `ErrorDepth` and `ErrorInsufficientBalance`, they indeed don't halt for the caller, and we could remove these state and handle them in the expected state like:

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/ad4a0744b04703ea87a6f8bf98ddf0bcdbbf603f/zkevm-circuits/src/evm_circuit/execution/callop.rs#L265-L288

For `ErrorContractAddressCollision` we could treat it as exception since it happens when it's returning to caller's context and the result will be the same as other exceptions (make sure `is_success = 0` and revert all state changes)?

PR #1299, opened 2023-03-09
Fix `ExecutionState::halts_in_exception`

han0110:
### Description

This PR aims to remove `ExecutionState::{ErrorDepth,InsufficientBalance
,ErrorContractAddressCollision}` from `ExecutionState::halts_in_exception` and handle `InsufficientBalance` properly by `CallOpGadget`

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1298

### Type of change

### Contents

Some exception states happen right in the moment when we are doing `*CALL*` or `CREATE*`, including:

- `ErrorDepth`
- `ErrorInsufficientBalance`
- `ErrorContractAddressCollision`

Which will only be treated as failure of sub-call, and won't halt current call.

```mermaid
stateDiagram-v2
    state fork <<fork>>
    state join <<join>>
    state NextCall {
        ...* --> join
    }
    state CurrentCall {
        ... --> fork
        fork --> CALL|CREATE
        fork --> ErrorDepth
        fork --> ErrorInsufficientBalance
        fork --> ErrorContractAddressCollision
        CALL|CREATE --> NextCall
        ErrorDepth --> join
        ErrorInsufficientBalance --> join
        ErrorContractAddressCollision --> join
        join --> NextStep
    }
```

However, if we want to handle these as a separate state we'd have many constraint duplicated (e.g. gas cost calculation), so `ErrorInsufficientBalance` is merged into `CALL` in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/998 without too much overhead. For the other 2 states we could apply the same strategy to avoid confusion.


zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -139,9 +138,6 @@ impl ExecutionState {
             Self::ErrorInvalidOpcode
                 | Self::ErrorStack
                 | Self::ErrorWriteProtection
-                | Self::ErrorDepth
-                | Self::ErrorInsufficientBalance
-                | Self::ErrorContractAddressCollision
```
han0110, line 144: @lispc Will this make thing worse if we remove them before implementing them in `CALL` or `CREATE`?

hero78119 (commented):
For the error handling just have different thoughts.
If an non-halt error is tightly couple with a state, how about
1. in `ExecError`, we still keep this error type, or maybe we define another new enum type `ExecNonHaltError`
2. we still try the best to recognize the error in `get_step_err`
3. bus-mapping `fn_gen_error_state_associated_ops` use normal state op to `gen_associated_ops`
4. still keep the ExeError -> ExecutionState mapping
5. in circuit gadget, use normal state gadget to process this error
6. make normal op support this error

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -299,8 +299,6 @@ pub(crate) struct ExecutionConfig<F> {
     error_oog_self_destruct:
         Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorOutOfGasSELFDESTRUCT }>>,
     error_oog_code_store: Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorOutOfGasCodeStore }>>,
-    error_insufficient_balance:
```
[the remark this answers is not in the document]
han0110: Make sense, updated in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1299/commits/a12fb95864603c1628f31907a338c3c13314ecb2

zkevm-circuits/src/evm_circuit/step.rs
hero78119: `ErrorDepth`, `ErrorContractAddressCollision`, `ErrorInsufficientBalance` these few states only use along with `DummyGadget`, 
 and these states never touch `halts_in_exception` because halts_in_exception only use in `RestoreContextGadget`. Therefore its constraints never activated. Maybe we open another issue to tracking the implemetatoin~?

han0110: Make sense, for `ErrorContractAddressCollision` we got https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1309 and for `ErrorDepth` we have https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1312 now.

hero78119 (approved):
LGTM 👍

PR #1311, opened 2023-03-10
bus-mapping: more robust access parsing

lispc:
### Description

When building AccessSet, we read stack input of opcode(eg: read last element of stack when processing  BALANCE opcode). But this method will encounter error when dealing with some ill-formed bytecode(BALANCE when empty stack). So in this PR, when access parsing for a single step encounters errors, the error is printed, the access parsing procedure will continue instead of being interrupted and return.

### Issue Link

[_link issue here_]

### Type of change

### Contents

- [_item_]

### Rationale

### How Has This Been Tested?

"StackUnderFlowContractCreation_d0_g0_v0" in testool can pass with this PR

<hr>

## How to fill a PR description 

Please give a concise description of your PR.

The target readers could be future developers, reviewers, and auditors. By reading your description, they should easily understand the changes proposed in this pull request.

MUST: Reference the issue to resolve

### Single responsability

Is RECOMMENDED to create single responsibility commits, but not mandatory.

Anyway, you MUST enumerate the changes in a unitary way, e.g.

```
This PR contains:
- Cleanup of xxxx, yyyy
- Changed xxxx to yyyy in order to bla bla
- Added xxxx function to ...
- Refactored ....
```

### Design choices

RECOMMENDED to:
- What types of design choices did you face?
- What decisions you have made?
- Any valuable information that could help reviewers to think critically

### DIFF
--- a/bus-mapping/src/circuit_input_builder/access.rs
+++ b/bus-mapping/src/circuit_input_builder/access.rs
@@ -3,6 +3,9 @@ use eth_types::{evm_types::OpcodeId, Address, GethExecStep, GethExecTrace, ToAdd
 use ethers_core::utils::get_contract_address;
 use std::collections::{hash_map::Entry, HashMap, HashSet};
 
+use AccessValue::{Account, Code, Storage};
+use RW::{READ, WRITE};
+
 /// State and Code Access with "keys/index" used in the access operation.
 #[derive(Debug, PartialEq, Eq)]
 pub enum AccessValue {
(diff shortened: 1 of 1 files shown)


bus-mapping/src/circuit_input_builder/access.rs
```diff
@@ -145,117 +145,124 @@ pub fn gen_state_access_trace<TX>(
...
+                        call_stack.push((address, CodeSource::Address(address)));
+                    }
                 }
+                _ => {}
             }
-            _ => {}
+            Ok(())
+        })();
+        if let Err(e) = result {
+            log::warn!("err when parsing access: {:?}, step {:?}", e, step);
```
hero78119, line 263: Wdyt if we log + break here earlier to skip all the left trace parsing? Although it wont happened on a `normal` geth trace since this op should be the last op, but error handling perspective terminate here make more sense

hero78119 (commented):
LTGM! Just raise a minor question

ed255 (approved):
LGTM! Thanks for submitting this fix!

bus-mapping/src/circuit_input_builder/access.rs
hero78119: I c, just realized stack underflow wont stop depth - 1 execution

Issue #966, opened 2022-12-06
Super circuit integration tests should use actual prover


leolara:
The super circuit was added after this task was started https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/789 so we should give it support for super circuit.

ed255:
A big part of this task is already complete.  The missing bits are:
- [ ] https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/966
- [ ] https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/967
- [ ] https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/968
- [ ] Get all tests with real prover to pass (some unidentified fixes required)

aguzmant103:
Needs Blocked by https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/967

PR #1315, opened 2023-03-10
Fix integration tests with real prover

davidnevadoc:
### Description

Some integration tests were failing when using the real prover but not when using the mock prover.
In particular, the keccak tests were the ones failing due the bug described in  issue https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1016 

### Issue Link
Solves https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1016
(Last task) Solves:  https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/966

### Type of change

### Contents

This PR applies the same changes made in this PR https://github.com/scroll-tech/zkevm-circuits/pull/308, plus some minor parameter adjustments.

### How Has This Been Tested?

This has been tested by running the integration tests using the mock prover. That is, with setup and genda don running:
`./run.sh --steps tests --tests circutis::real_prover`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,27 +1,93 @@
 # zkEVM contribution guidelines
 
-Thank you for investing your time in contributing to our project! 
+Thank you for investing your time in contributing to our project!
 
-Before starting, we encourage you to take a look to the following documentation as it might help you in your first contribution adventure:
-- https://github.com/privacy-scaling-explorations/zkevm-docs
-- https://github.com/privacy-scaling-explorations/zkevm-specs
+# How you can Contribute
+
+First take a look at our [specs](https://github.com/privacy-scaling-explorations/zkevm-specs), [auditing videos](https://www.youtube.com/watch?v=HhHTho2QZa4), and [zkevm docs](https://github.com/privacy-scaling-explorations/zkevm-docs). It’s very important that you understand at a high-level how the ZK-EVM architecture and all of its components work.
+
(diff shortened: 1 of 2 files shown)


zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
```diff
@@ -11,7 +11,7 @@ use std::{env::var, vec};
 
 pub(crate) fn get_num_rows_per_round() -> usize {
     var("KECCAK_ROWS")
-        .unwrap_or_else(|_| "5".to_string())
```
leolara, line 14: @davidnevadoc why this increase?

zkevm-circuits/src/keccak_circuit/keccak_packed_multi.rs
```diff
@@ -11,7 +11,7 @@ use std::{env::var, vec};
...
+        .unwrap_or_else(|_| "12".to_string())
```
leolara, line 14: @davidnevadoc why this increase?

lispc: as we discussed with Han previously，circuit being wider is friendly for super circuit, But the more columns the more  proving cost of aggregation circuit. So we did not choose minimum possible value But a bit larger

leolara: Ok

PR #1316, opened 2023-03-11
Abstract code hash function into CodeDB

z2trillion:
### Description

Abstract code hash function into CodeDB to clean up code and so that implementing non-keccak code hashes will be easier.

### Type of change

### Contents

Add `empty_code_hash` and `hash` methods to `CodeDB`
Replace existing uses of `EMPTY_HASH` and `H256(keecak256(...))` with the above two, respectively. 
Remove `EMPTY_HASH` and `EMPTY_HASH_LE` from keccak256 lib and add `EMPTY_CODE_HASH_LE` to `state_db`.

### How Has This Been Tested?

Existing tests cover this change and still pass.


lispc:
@han0110 @ed255 
more background on this PR: although a strict type-1 zkevm don't need this abstraction, while type-2 / type-2.5 zkevm needs this abstraction. For example, for scroll zkevm we use keccak/poseidon dual code hash, then without this PR, it will be heavy burden to maintain our diffs long term.     In the future i think it will be great for more teams to reuse the "zkevm-circuits" to developer their own l2, they may use codehash like poseidon/rescue/others.     Within this type of abstraction, it will bring much benifit.

han0110 (approved):
LGTM! It makes sense to gather cryptographic primitive dependency into single place for easy transition to another if necessary. Nice refactor!

PR #1319, opened 2023-03-14
Integration tests with ephemeral runners via GitHub ops

ntampakas:
### Description

This PR introduces github-ops functionality to zkevm-circuits/integration-tests, allowing execution on ephemeral AWS hosted runners. Execution mode is parameterized based on event trigger type and prover type selection.    

### Type of change

### Contents

- .github/workflows/integration.yml
- integration-tests/run.sh
- integration-tests/README.md

```
This PR contains:
- A new job that initializes prover and instance type workflow variables 
- run.sh updated to Compose V2
- Updated README.md file to reflect the updated docker compose commands
```

### Design choices
As per new integration tests execution mode, we run the tests manually, on schedule or on specific GitHub events (PR, push to main). Depending on workflow setup, test is executed either with real or mock prover, thus different AWS instance is selected to accommodate increased resource requirements for real prover. For this reason, we introduced a new job that initializes instance type and prover type variables as job outputs to be used for subsequent jobs (integration-tests).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -25,11 +25,30 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  set-instance-type:
+    if: github.event.pull_request.draft == false
+
+    runs-on: ubuntu-latest
+    outputs:
+      output1: ${{ steps.set-instance-type.outputs.type }}
+    steps:
+      - id: set-instance-type
+        name: Select instance type
+        run: |
+          if [ "${{ github.event.inputs.provertype }}" = "mock_prover" ] || [ -z ${{ github.event.inputs.provertype }} ]; then
+            echo "type=c5.9xlarge" >> "$GITHUB_OUTPUT"
+          elif [ "${{ github.event.inputs.provertype }}" = "real_prover" ]; then
+            echo "type=c5.24xlarge" >> "$GITHUB_OUTPUT"
+          else
+            exit 1
+          fi
+
   integration-tests:
     if: github.event.pull_request.draft == false
 
     name: Integration Tests
-    runs-on: ["${{github.run_id}}", self-hosted, x1e.4xlarge]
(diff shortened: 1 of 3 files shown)


leolara (approved):

.github/workflows/integration.yml
```diff
@@ -97,14 +82,20 @@ jobs:
...
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      # Run an initial build in a separate step to split the build time from execution time
+      # Run an initial build in a separate step to split the build time from execution time    
       - name: Build bins
         run: cargo build --bin gen_blockchain_data
       - name: Build tests
         run: for testname in rpc circuit_input_builder circuits; do cargo test --profile release --test $testname --features $testname --no-run; done
       - run: ./run.sh --steps "setup"
+        working-directory: integration-tests
```
ed255, line 91: I noticed that this directive was previously specified in "defaults". I think it should also be applied for "Build bins" and "Build tests".  Maybe it also works from the root of the repository, but I think the "Build tests" step may build much more than necessary (so it will take more time than necessary)

ed255 (approved):
LGTM, nevertheless I've added a note about the `working-directory` directive.  Please take a look :)

.github/workflows/integration.yml
ntampakas: working-directory was removed from the default section, because during the workflow execution we received a "No such file or directory" error message. With current workflows a generic image is used to deploy the runner, via github-ops, therefore at this specific execution point, the expected dir structure is not there and the default/working-directory fails. As a result, we need to set working directory after checkout step, for each separate run directive of the job. 

"Build bins" and "Build tests" work fine from the default working repo directory. Of course if there is a reason that we 're missing we can adapt.

ChihChengLiang (approved):
LGTM

PR #1320, opened 2023-03-15
Fix to enable `gas_is_u64` for failed Calls

silathdiir:
### Description

Found by `testool` cases (`ABAcalls1_d0_g0_v0`).

Reference [go-ethereum callGas function](https://github.com/ethereum/go-ethereum/blob/master/core/vm/gas.go#L44), the returned gas (if input is Unit64 overflow) could also cause out of gas error.

### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

Add unit-test case `test_oog_call_with_overflow_gas` (failed without this fix).


hero78119 (approved):
LGTM!

PR #1326, opened 2023-03-20
remove create for address collision detect

DreamWuGit:
### Description
remove create for address collision detect
### Issue Link
todo

### Type of change

### Rationale

for create op code and tx deploy case(eventually use `create` method), creating contract address is derived from kecakk(caller_address, caller_nonce), when call `create`, the caller_once increase by 1 so never meet address collision issue.

This PR contains:
-  Added trace test to show `create` won't get  address collision error
- remove `create` when detecting address collision error

### DIFF
--- a/bus-mapping/src/circuit_input_builder/tracer_tests.rs
+++ b/bus-mapping/src/circuit_input_builder/tracer_tests.rs
@@ -482,6 +482,142 @@ fn tracer_err_address_collision() {
     );
 }
 
+#[test]
+fn tracer_create_without_collision() {
+    // We do CREATE twice with the same parameters, with a code_creater
+    // that outputs the same, which will lead to the different new
+    // contract address.
+    let code_creator = bytecode! {
+        PUSH1(0x00) // value
+        PUSH1(0x00) // offset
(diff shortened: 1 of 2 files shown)


bus-mapping/src/circuit_input_builder/tracer_tests.rs
```diff
@@ -482,6 +482,142 @@ fn tracer_err_address_collision() {
     );
 }
 
+#[test]
+fn tracer_create_without_collision() {
```
hero78119, line 486: `tracer_create_collision_free` more precise

hero78119 (approved):
Thanks for catch up this! just nitpick on unittest function naming, other LGTM

bus-mapping/src/circuit_input_builder/tracer_tests.rs
DreamWuGit: updated in [c5dcce4](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1326/commits/c5dcce4fb6e4c8659c6e56696603925d4203c908) thanks!

PR #1331, opened 2023-03-21
feat: SuperCircuit configuration flags

pinkiebell:
### Description

Add compile-time configuration flags to selectively enable sub circuits / block,mpt table.
Due to the design of the halo2 api we can't do this at runtime.

### Type of change

### Rationale

Selectively enabling parts of the super circuit configuration comes in handy when testing and debugging and makes the downstream integration a bit more streamlined. The sub circuits can't be used directly so that we need a dummy test circuit for each of the sub circuits but exporting them publicly increases the maintenance factor and increases the likelihood of introducing breaking changes for downstream crates. Hence, allow the SuperCircuit to be configured to avoid exporting test circuits for each of the sub circuits.


ed255:
I find this proposal very interesting!  I have some questions:
- Is this feature intended to be used with the MockProver? Real Prover? Both?
- I see that the configuration of the SuperCircuit is untouched.  This means that even with disabled parts, the columns and gates are still created.  Did you think about removing this based on the flags?
- You mention that halo2 doesn't support runtime configuration, but following the previous question: I didn't see any usage of checking the flags in this PR happening in the configuration.  So what's the rationale of using the const generic instead of adding a struct with boolean fields to the `SuperCircuit` type?

On a different note, I think we should add a way to support runtime configuration of circuits in our halo2 fork, I think we will benefit greatly from that.

pinkiebell:
> I find this proposal very interesting! I have some questions:
> 
> * Is this feature intended to be used with the MockProver? Real Prover? Both?

This envision that this feature is mainly used in testing with MockProver and in zkevm-chain. However, I think using that on any setup can replace all the dummy circuit setups we have in place.

> * I see that the configuration of the SuperCircuit is untouched.  This means that even with disabled parts, the columns and gates are still created.  Did you think about removing this based on the flags?

Ah yes. I had the feeling I was missing something the whole time but didn't grok it at the time of writing 😄 

> * You mention that halo2 doesn't support runtime configuration, but following the previous question: I didn't see any usage of checking the flags in this PR happening in the configuration.  So what's the rationale of using the const generic instead of adding a struct with boolean fields to the `SuperCircuit` type?

I will add this to the configuration too. 

> On a different note, I think we should add a way to support runtime configuration of circuits in our halo2 fork, I think we will benefit greatly from that.

Yes. This would greatly improve developer experience and binary size / compilation times 🙏

pinkiebell:
@ed255 I wrapped the members of `SuperCircuitConfig` with `Option`. I did like to know what you think about this and if we go forward with that approach, shall I do the same with the SuperCircuit struct? It would add a bit more 'noise' to the super circuit and we only clone some data in `fn new_from_block` anyway, might not be worth it - fyi

zkevm-circuits/src/super_circuit.rs
```diff
@@ -334,28 +553,49 @@ impl<F: Field, const MAX_TXS: usize, const MAX_CALLDATA: usize, const MOCK_RANDO
...
-            .synthesize_sub(&config.state_circuit, challenges, layouter)?;
-        self.copy_circuit
-            .synthesize_sub(&config.copy_circuit, challenges, layouter)?;
-        self.exp_circuit
-            .synthesize_sub(&config.exp_circuit, challenges, layouter)?;
-        self.evm_circuit
-            .synthesize_sub(&config.evm_circuit, challenges, layouter)?;
-        self.pi_circuit
-            .synthesize_sub(&config.pi_circuit, challenges, layouter)?;
+        if let Some(circuit_config) = &config.keccak_circuit {
```
ed255, line 556: I think this needs to handle more conditions.
We have a list of circuits and a list of tables.
Some circuits have dependencies on tables, and you encoded that in `SuperCircuitConfig::new` with flag checks for each table construction, to make sure the tables exist whenever some circuit that uses them is enabled.
Then whenever a table is required by a circuit, the table should be assigned.  If the circuit associated with the table is enabled, then the circuit itself has the assignment code which is in `sub_circuit.synthesize_sub`.  But when that circuit is not enabled but yet the table is needed, we still need to assign it.  For that we have methods to assign the table outside of it's corresponding circuit.  Here's an example of how the Bytecode test circuit assigns the keccak table:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/3aa36a2ca9360723a5eb87816a6b02655598d2d5/zkevm-circuits/src/bytecode_circuit/dev.rs#L49-L53

So I think each `if` here requires an else.
For example in the keccack circuit the else should do:
```rust
} else if is_enabled(
    flags,
    SUPER_CIRCUIT_FLAG_TX
    | SUPER_CIRCUIT_FLAG_BYTECODE
    | SUPER_CIRCUIT_FLAG_EVM,
    ) {
        config.keccak_table.dev_load(
            &mut layouter,
            self.bytecodes.iter().map(|b| &b.bytes),
            &challenges,
        )?;
}
```

ed255:
> @ed255 I wrapped the members of `SuperCircuitConfig` with `Option`. I did like to know what you think about this and if we go forward with that approach, shall I do the same with the SuperCircuit struct? It would add a bit more 'noise' to the super circuit and we only clone some data in `fn new_from_block` anyway, might not be worth it - fyi

I can't think of any other way to approach this other than having an Option for each field in the `SuperCircuitConfig`, so this approach seems good to me!

I'm not sure if the same is needed for the `SuperCircuit`: that struct only contains witness data in general, and as you say, we only save a few clones in testing and bench environments, where the bottleneck is in other areas right now.  So I would leave that as is.

Please take a look at my comment https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1331#discussion_r1170044032 it may be related to the failing test!

ChihChengLiang:
Close this one due to inactivity. @pinkiebell feel free to reopen.

Issue #1332, opened 2023-03-21
Soundness bug in Bytecode circuit 1 (missing state transition constraint)


ed255:
The Bytecode circuit is missing the constraint that forces the a "header" state after the last "byte" state in a bytecode.  This allows an attacker to insert bytes after the last real byte.  If we insert a "byte" state after the last bytecode "byte", we reach a situation in which we can only have "byte" states until the end of the circuit.  In theory this should not be possible because we are constraining the last state to be "header", but there's another bug here!  `q_enable` is used in the constraint to force the last row to be "header", and `q_enable = 0` in the last row.

Here's an example of the attack:

For a bytecode circuit of size 6 which contains 1 bytecode of length 3 it looks like this:

| Tag    | Index | Value | CodeHash |
| ------ | ----- | ----- | -------- |
| Header | 0     | 3     | 0        |
| Byte   | 0     | a     | C1       |
| Byte   | 1     | b     | C1       |
| Byte   | 2     | c     | C1       |
| Header | 0     | 0     | 0        |
| Header | 0     | 0     | 0        |

But we can do this assignment and the constraints will work

| Tag    | Index | Value | CodeHash |
| ------ | ----- | ----- | -------- |
| Header | 0     | 3     | 0        |
| Byte   | 0     | a     | C1       |
| Byte   | 1     | b     | C1       |
| Byte   | 2     | c     | C1       |
| Byte   | 3     | d     | C1       |
| Byte   | 4     | e     | C1       |

We're introducing bytes 3 and 4 and we can use any value for those.

# Summary

## Bug 1

The transition Byte -> Header is not constrained to be true when the appropriate condition appears
- Solution: Add a constraint that forces the next state to be "Header" when the current state is "Byte" and index == length - 1

## Bug 2

The last row has `q_enable=0`, which effectively disables all polynomial constraints applied in that row (including the one that forces the row to be a "Header" state).
- Solution: Set `q_enable=1` to all rows of the circuit, including the last one

# Demo

This commit https://github.com/privacy-scaling-explorations/zkevm-circuits/compare/main...bug/bytecode_soundness1 includes a test and some helper code that overwrites the assignment to show how bytes can be added after the "real" bytecode, and all the constraints still pass.

rrzhang139:
Would like to give this a try! Btw, there is a typo in Bug 2. I think you meant to say q_enable=0 on the last row. Will first change the specs!

ed255:
[...]

PR #1333, opened 2023-03-24
Bytecode soundness patch on last byte

rrzhang139:
### Description

Fixes bug soundness in bytecode circuit, where an attacker can insert as many bytes after the last byte (row where `q_last`=1). 

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1332

### Type of change

### Contents

- Bug fix 1: Add a constraint that forces the next state to be "Header" when the current state is "Byte" and `index == length - 1`
- Bug fix 2: set `q_enable=1` for the row where `q_last=1`

### Rationale

This would patch the possibility of an attacker inserting false bytes that extend beyond the correct bytecode. 

### How Has This Been Tested?

Created a test called `bytecode_soundness_bug_1` that overwrites the last rows with`tag=Byte`. 
`cargo test --features test --package zkevm-circuits --lib -- bytecode_circuit::circuit::tests::bytecode_soundness_bug_1 --exact --nocapture`
Also ran other unit tests to check correctness was maintained.


rrzhang139:
@ed255 pushed some changes to fix bug 1 and bug 2. unit tests don't pass because I still have your overwrite test which should fail now.

circuit-benchmarks/src/bench_params.rs
```diff
@@ -0,0 +1 @@
+pub(crate) const DEGREE: usize = 19;
```
ed255, line 1: This new file seems unrelated to this PR

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -467,11 +509,56 @@ impl<F: Field> BytecodeCircuitConfig<F> {
...
                         &mut region,
                         &push_data_left_is_zero_chip,
+                        &index_length_diff_is_zero_chip,
                         empty_hash,
                         idx,
                         last_row_offset,
                     )?;
                 }
+
+                // Overwrite
```
ed255, line 519:
```suggestion
                // Overwrite the witness assignment by using the values in the `overwrite` parameter.  This is used to explicitly set intermediate witness values for negative tests.
```
I think extending this comment will add more clarity on what's going on here

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -1094,4 +1212,38 @@ mod tests {
             test_bytecode_circuit_unrolled::<Fr>(k, vec![invalid], false);
         }
     }
+
+    use halo2_proofs::dev::MockProver;
+
+    #[test]
+    fn bytecode_soundness_bug_1() {
```
ed255, line 1219:
```suggestion
    #[test]
    #[should_panic]
    fn bytecode_soundness_bug_1() {
```

See https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic

This allows us to keep this negative test and still pass the CI

ed255 (commented):
LGTM overall!  Thanks for submitting the fix to this soundness bug.
My main comment is about tagging the bug test as `should_panic` so that we can keep it and pass the CI.  See my other two small comments as well.

I'll approve the PR once the CI passes

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -467,11 +509,56 @@ impl<F: Field> BytecodeCircuitConfig<F> {
...
+                    ] {
+                        region.assign_advice(
+                            || format!("assign {} {}", name, offset),
+                            column,
+                            offset,
+                            || Value::known(value),
+                        )?;
+                    }
+
+                    if row.tag == F::one() {
```
leolara, line 537: If we are overriding, shouldn't we also override the value_rlc or leave it alone alone as it was before?

Perhaps we could implement a system that allows to choose which columns we override?

leolara:
Perhaps we take this opportunity to alsof fix: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1347

leolara (changes requested):

circuit-benchmarks/src/bench_params.rs
leolara: I agree this change should not be included, please

zkevm-circuits/src/bytecode_circuit/circuit.rs
```diff
@@ -342,6 +359,24 @@ impl<F: Field> SubCircuitConfig<F> for BytecodeCircuitConfig<F> {
             ]))
         });
 
+        // bug 1 fix
+        // When cur.tag == Byte and cur.index + 1 == cur.length ->
+        // assert next.tag == Header
+        meta.create_gate("cur.tag == Byte and cur.index + 1 == cur.length", |meta| {
+            let mut cb = BaseConstraintBuilder::default();
+
+            cb.require_zero(
```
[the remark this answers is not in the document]
ed255: I think I don't understand this proposal.
Would this gate change following your proposal?  Maybe you can make a diff to show how it would look?

zkevm-circuits/src/bytecode_circuit/circuit.rs
ed255: > If we are overriding, shouldn't we also override the value_rlc or leave it alone alone as it was before?

I wrote this overwrite code originally.  The reason why `value_rlc` is not included in the overwrite values is because it can't be precomputed as it depends on the challange, which is only known at phase 2.  So the lines below compute a correct value_rlc during assignment.

> Perhaps we could implement a system that allows to choose which columns we override?

This is the approach taken in the state circuit: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/0e76d2b25e62c22f75b26a64d9d7ea78bb3fede0/zkevm-circuits/src/state_circuit.rs#L500
Nevertheless I'm not sure if it would be useful here. (I think it would make the negative tests more complex)

zkevm-circuits/src/bytecode_circuit/circuit.rs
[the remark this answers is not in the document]
leolara: I don't know how to add the diff.

So this gate would not exists and we would add `cb.require_zero(and::expr(vec![is_next_byte(meta), index_length_diff_is_zero.clone().is_zero_expression])` to the gate annotated with "Byte Row", in main line starting in 177.

So if you are in a Byte row: next row to be byte AND index_length_diff == 0 must be false.

I think that would follow more the rest of the code structure.

zkevm-circuits/src/bytecode_circuit/circuit.rs
leolara: Ok

leolara:
I guess we need to update the spec?

zkevm-circuits/src/bytecode_circuit/circuit.rs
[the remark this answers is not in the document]
ed255: Thanks for the diff!
So the original constraint was formed like: If we're in this tag and this condition, then next tag must be this
The new one is formed like: if we're in this tag, then this next tag can't happen with this condition.

They are equivalent, but I think the original one is a bit more clear?  I would leave @rrzhang139 to choose the way forward.

rrzhang139:
@ed255 Should I perhaps remove the override code and the bytecode_soundness_bug_1() test? Or is there a better way to test overriding without changing the function signature? Seems unnecessary to have it just for one test.

ed255:
> @ed255 Should I perhaps remove the override code and the bytecode_soundness_bug_1() test? Or is there a better way to test overriding without changing the function signature? Seems unnecessary to have it just for one test.

I don't follow.  Are you referring to the `assign` and `assign_internal` function signatures?  I think it's not too bad to leave the override code and the bug test, specially because the `overwrite` parameter is not included in the `BytecodeCircuit` constructor.  Let me know if you're referring so something else.

leolara:
We will use override for other tests in the future. But I insist that keeping the "bug 1" or "bug 2" in the code will not bring clarity as nobody will know what these number mean in a few months.

ed255 (approved):
LGTM!

leolara (approved):

rrzhang139:
@ed255 @leolara can I get this merged now?

leolara:
@rrzhang139 yes, you can merge with two approvals 👍

rrzhang139:
@leolara I can't merge because I don't have write access on this repo

leolara:
@rrzhang139 sorry I never know who can

Issue #1302, opened 2023-03-09
incorrect optimization of yul compiler in testool


lispc:
### Describe the bug

for the "tests/src/GeneralStateTestsFiller/stCreate2/Create2OOGFromCallRefundsFiller.yml" test case,

yul compiler output sstore(1,1) as `PUSH1 1; DUP1; SSTORE`, but the create2 address inside yml expect `PUSH1 1; PUSH1 1; SSTORE`..  

I am investigate how to disable this optimization

lispc:
the optimization is introduced in https://github.com/ethereum/solidity/releases/tag/v0.8.7

lispc:
this line https://github.com/ethereum/solidity/blob/abaa5c0eb321aab4cd09617598696172378a4b83/libyul/AssemblyStack.cpp#L291

introduced in https://github.com/ethereum/solidity/commit/847e30e6ffb2e0b82a0888609184fd7db9f208a4

lispc:
I will make a PR to update dockerfile and codehash.txt soon

PR #1341, opened 2023-03-30
disable optimization for solc compiling

lightsing:
fix https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1302


lispc:
Patch diff  can be deleted

testool/src/compiler.rs
```diff
@@ -71,6 +72,139 @@ pub struct Compiler {
...
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SourceLocation {
+    end: i32,
+    file: String,
+    start: i32,
+}
+
 impl Compiler {
```
leolara, line 208: Why is the compiler impl so far from the compiler struct, if this compiler depends now in the new structs, perhaps we should move struct Compiler down, right before impl Compiler?

leolara (changes requested):
Great adding more abstraction to this, but I have one question.

testool/src/compiler.rs
lightsing: solved

lispc (approved):

leolara (approved):

PR #1346, opened 2023-03-30
ci/test-features: add dummy complete step for branch rules

pinkiebell:
### Description

Add a single dummy step that can be easily added to the branch protection rules.
Without this change any matrix combination has to be added to branch protection rules independently.


ChihChengLiang:
Merging in case other PR got blocked.

PR #1353, opened 2023-04-12
Witness manipulation moved from the circuit to the witness generator

miha-stopar:
The goal of this PR is to modify witness descriptors from `[][]byte` to `Vec<Node>` where `Node` matches the struct in the MPT circuit. Also, the witness generator code will be modified to use the same structs as used in the MPT circuit.

The corresponding witness generator PR: https://github.com/privacy-scaling-explorations/mpt-witness-generator/pull/2


zkevm-circuits/src/mpt_circuit/witness_row.rs
```diff
@@ -245,8 +245,7 @@ impl<F: Field> MptWitnessRow<F> {
 
 // TODO(Brecht): Do all of this on the MPT proof generation side
```
miha-stopar, line 246: @Brechtpd: Now the RLP bytes are added to the rows where they were not available previously (like extension C row) and some streams are shifted to the left (like account codehash row). I will remove branch children C part and move modified node into branch init, but I am not sure if you meant by “all of this” the whole function? Because for example `rlp_bytes` are now stored at the beggining of the row, but the function sets `row.rlp_bytes` and then discards RLP bytes and shifts the remaning bytes to the left. Is this meant to stay as it is or did you mean to move the RLP bytes to some other place in the witness generator?

Brechtpd: I think ideally the witness generator generates some json file that stores the `Vec<Node>` data returned by this function, so the only thing we have to do on the circuit side is to deserialize the json and we're good to go. The witness is currently set directly from this data, the benefit is that it's easier to change the the circuit if the data is stored in a more high level format than the current values directly expected by the circuit (but which gets inconvenient when changing how things work in the circuit).

miha-stopar:
This PR is obsolete, these changes have been later merged into mpt-2 branch and are now already in master.

Issue #916, opened 2022-11-18
LtChip is missing byte range checks


ed255:
The `LtChip` defined in `gadgets/src/less_than.rs` uses this constrain:
`lhs - rhs = diff - lt * 2.pow(8*N)`
where `diff = from_bytes(diff_bytes)` and `diff_bytes: [Column<Advice>; N],`

This constrain is only useful if `diff` is constrained to be between `0` and `2^{8*N}`.  This can be achieved by constraining each element in `diff_bytes` to be between `0` and `2^8`.  Nevertheless this is not constrained.  There's a note in the struct field documentation saying this:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4cfccfa6c3b251284ff61eeb907d548d59206753/gadgets/src/less_than.rs#L33-L35
But I believe this is quite dangerous, because it's very easy to forget!

I've searched for usages of `LtChip` and I found it's used only in the `CopyCircuit`: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4cfccfa6c3b251284ff61eeb907d548d59206753/zkevm-circuits/src/copy_circuit.rs#L69
And I didn't see any byte range constraint on the elements in `addr_lt_addr_end.diff`, which is a bug.

I think we must avoid leaving gadgets with missing constraints and expecting the chip that uses the gadgets to add them (unless we had a way to automatically check that we didn't miss these constraints)

enricobottazzi:
Hi @ed255, I'm using this chip inside my circuit and I'm interested in fixing the bug you mentioned here. 

Can you please provide some high-level direction on how you would suggest fixing it? 

My intuition would be to use a lookup table of 2^8 bits. I'm not sure if you already have such a lookup table implemented somewhere in your circuits that can be called here.

ed255:
Resolved via https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1364

PR #1364, opened 2023-04-18
Add Byte Range Checks in `LtChip`

enricobottazzi:
### Description

Add Byte Range Checks in `LtChip`. This PR modifies the `LtChip`:

- Added Fixed Column `u8` to the configuration of the chip 
- Added a lookup constraint between each `diff` advice column and the `u8` column
- Added `load` function that assigns each integer from 0 to 255 (8bits) to the `u8` column

Furthermore, the following have been applied to the tests: 

- In `less_than` testing, the circuit now takes `k = 9` to support a higher number of rows
- In `copy_circuit` testing, the `assign_copy_events` function now calls `load` on the `lt_chip` to load the `u8` column
 
### Issue Link

Related to #916 

### Type of change

### How Has This Been Tested?

`make test-all`


gadgets/src/less_than.rs
```diff
@@ -61,6 +65,7 @@ impl<F: Field, const N_BYTES: usize> LtChip<F, N_BYTES> {
         let lt = meta.advice_column();
         let diff = [(); N_BYTES].map(|_| meta.advice_column());
         let range = pow_of_two(N_BYTES * 8);
+        let u8 = meta.fixed_column();
```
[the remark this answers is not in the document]
enricobottazzi: It looks like [the first](https://github.com/privacy-scaling-explorations/halo2/blob/fcdd5b910d92ab278d679cf8aa176c885455aa6c/halo2_proofs/src/plonk/circuit.rs#L1891) API returns a `Column<Fixed>` while [the second](https://github.com/privacy-scaling-explorations/halo2/blob/fcdd5b910d92ab278d679cf8aa176c885455aa6c/halo2_proofs/src/plonk/circuit.rs#L1856) returns a`TableColumn`, which is a wrapper around a fixed column.

From the [Halo2 Discord](https://discord.com/channels/995022112811135148/1005560452446683186/1085076854874054806) I read that "halo2 fills in TableColumns with a specified default value". It would be interesting to understand the differences between the two further.

I also saw the range check [being performed as you suggest](https://github.com/icemelon/halo2-examples/blob/master/src/range_check/example2/table.rs), but eventually I opted for this syntax to maintain coherence with [the example](https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/916#issuecomment-1508209602) that @ed255 linked in the issue.

gadgets/src/less_than.rs
```diff
@@ -81,7 +86,21 @@ impl<F: Field, const N_BYTES: usize> LtChip<F, N_BYTES> {
...
 
-        LtConfig { lt, diff, range }
+        meta.annotate_lookup_any_column(u8, || "LOOKUP_u8");
+
+        diff[0..N_BYTES].iter().for_each(|column| {
+            meta.lookup_any("range check lookup", |meta| {
+                let exp = meta.query_advice(*column, Rotation::cur());
+                vec![(exp, meta.query_fixed(u8, Rotation::cur()))]
+            });
+        });
```
ChihChengLiang, line 96: I have two suggestions here:

1. What do you think we merge them into one gate? It looks more readable to me if one purpose is achieved using only one gate.
2. It looks like the term "exp" is copied from the exp circuit. We can rename it to u8-related terms.

```suggestion
        meta.lookup_any("range check for u8", |meta| {
            diff[0..N_BYTES]
                .iter()
                .map(|column| {
                    let u8_cell = meta.query_advice(*column, Rotation::cur());
                    let u8_range = meta.query_fixed(u8, Rotation::cur());
                    (u8_cell, u8_range)
                })
                .collect()
        });
```

gadgets/src/less_than.rs
```diff
@@ -157,7 +195,8 @@ mod test {
 
             // TODO: remove zk blinding factors in halo2 to restore the
             // correct k (without the extra + 2).
-            let k = usize::BITS - $values.len().leading_zeros() + 2;
+            // let k = usize::BITS - $values.len().leading_zeros() + 2;
```
ChihChengLiang, line 198: Leftover?
```suggestion
```

gadgets/src/less_than.rs
```diff
@@ -174,7 +213,8 @@ mod test {
...
```
ChihChengLiang, line 216:
```suggestion
```

ChihChengLiang (approved):
LGTM after all the nitpick feedback addressed.

gadgets/src/less_than.rs
enricobottazzi: I followed your advice as the two expressions seem equivalent to me. The only difference is that expression 1 creates a vector with a single tuple for each element in the iterator, while Expression 2 creates a tuple for each element and then collects all tuples into a collection.

Despite that, the version containing the commit with the updated syntax generates an error in the range check (as you can see below). Can you explain why?

ChihChengLiang: I'm so sorry. I made a mistake. Your previous implementation was correct. My suggestion was wrong. 
My proposed constraints would enforce all cells to be the same value.

Think about looking up the expression `A xor B = C`. You would program the gate to return `vec![(cell_a, table_col_a), (cell_b, table_col_b), (cell_c, table_col_c)]`. That means cells a, b, and c must be on the same row of the table columns.

Now say we have cells `u8_a` and `u8_b`. Writing two gates constrain them to be within u8 range, but `a` and `b` could differ. Writing one gate that returns a vector of two elements adds an additional constraint that a and b have to be equal.

enricobottazzi: No problem at all :) I reverted to the prev version.

I guess that the design of the columns in the `xor` example would be something like this:

| adv__1 | adv_2 | adv_3 | table_col_a | table_col_b | table_col_c |
| ------ | ----- | ----- | ----------- | ----------- | ----------- |
| cell_a     | cell_b     | cell_c     | 0           | 0           | 0           |
|        |       |       | 0           | 1           | 1           |
|        |       |       | 1           | 0           | 1           |
|        |       |       | 1           | 1           | 0           |

> That means cells a, b, and c must be on the same row of the table columns.

Why is that?

> Writing one gate that returns a vector of two elements adds an additional constraint that a and b have to be equal.

I still don't get why is that.  In the `xor` example, even though we created one big gate, we weren't constraining `cell_a`, `cell_b`, `cell_c` to be equal.

ChihChengLiang: When we do XOR, we are looking up for 
```
(a, b, c) ∈ {(0, 0, 0), (0, 1, 1), (1, 0, 1), (1, 1, 0)}
```
but not for
```
(a ∈ {0, 0, 1, 1}) ^(b ∈ {0, 1, 0, 1}) ^(c ∈ {0, 1, 1, 0})
```

`a=1, b=1, c=1` is an invalid XOR witness, but it is allowed in the second expression.

ChihChengLiang: I tried to create an example to reproduce the lookup to the same table column.

https://github.com/ChihChengLiang/halo2_examples/blob/bb94962a3a9a92394ff0b088fa431e8173f3ef57/src/main.rs#L128-L130

enricobottazzi: Thanks cc. The `xor` example is fully clear to me. 

So trying to apply this concept to the `range_check` example.

| adv__1 | lookup_u8 |
| ------ | ----- | 
| u8_a   | 0     |
| u8_b   | 1      |   
|            |    2   |  
|            |     ...  |  
|            |   255  |  

1.

```
(u8_a, u8_b) ∈ {0, 1, 2, ..., 255}
```

This gate constrains `u8_a` and `ub_a` to be in the range **and** to be equal.

2.

```
(u8_a ∈ {0, 1, 2, ..., 255}) ^ (u8_b ∈ {0, 1, 2, ..., 255})
```

This gate constraints `u8_a` and `ub_a` to be in the range.

Is that correct?

ChihChengLiang: Yes, that's correct.

PR #1371, opened 2023-04-25
Updated Chiquito lookup table grammar for bytecode circuit

qwang98:
## Description
- Use Chiquito tag version "v2023_04_24", which updated PLONKish lookup table syntax with auto annotation features
- Updated the bytecode circuit according to the new syntax

## Contents
This is an update to Leo's original pull request here: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1348


leolara:
@qwang98 there is some indentation that needs to be fixed before it can be merged 

https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/4794651096/jobs/8551575470?pr=1371

Merge after fixing that

PR #1372, opened 2023-04-25
[feat] add memory reconstruction for precompile

lightsing:
### Description

Added memory reconstruction for call to precompile contracts.

### Issue Link

No issue is associated with this pull request.

### Type of change

### Contents

- Added a new function `is_precompiled` to check if the address of the contract is a precompiled contract.
- Added a new function `execute_precompiled` to execute a precompiled contract.
- Added a new helper function `code_address` associated with `Call` struct to get the actual address of code execution.
- Reconstruct memory for call to precompile address.

### Rationale

The addition of support for call to precompiled contracts in bus-mapping allows no-memory exec trace.

### How Has This Been Tested?

This PR only includes the memory reconstruction part. It has been test locally via `bus_mapping::evm::opcodes::callop::tests::test_precompiled_call`.


bus-mapping/src/precompile.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub(crate) fn execute_precompiled(address: &Address, input: &[u8], gas: u64) -> (Vec<u8>, u64) {
+    let Some(Precompile::Standard(precompile_fn)) = Precompiles::latest()
+        .get(address.as_fixed_bytes())  else {
+        panic!("calling non-exist precompiled contract address")
+    };
+
+    match precompile_fn(input, gas) {
+        Ok((gas_cost, return_value)) => (return_value, gas_cost),
+        Err(_) => (vec![], gas),
```
lispc, line 21: log err here

lightsing: As anticipated, if an error arises due to invalid input data or insufficient gas, logging the error here will not assist in the debugging process.

lightsing:
the toolchain need to be upgraded to `nightly-2022-12-10` or higher.

bus-mapping/src/precompile.rs
```diff
@@ -0,0 +1,23 @@
+//! precompile helpers
+
+use eth_types::Address;
+use revm_precompile::{Precompile, Precompiles};
+
+/// Check if address is a precompiled or not.
+pub fn is_precompiled(address: &Address) -> bool {
+    Precompiles::latest()
```
ed255, line 8: Perhaps we could use `Precompiles::berlin()` here instead of `latest` so that it's obvious which which set of precompiles we have enabled (and to avoid surprises in the future if there are new precompiles and we update the revm dependency)

bus-mapping/src/precompile.rs
```diff
@@ -0,0 +1,23 @@
...
+        .get(address.as_fixed_bytes())
+        .is_some()
+}
+
+pub(crate) fn execute_precompiled(address: &Address, input: &[u8], gas: u64) -> (Vec<u8>, u64) {
+    let Some(Precompile::Standard(precompile_fn)) = Precompiles::latest()
```
ed255, line 14: Same as the last comment.

ed255 (approved):
LGTM!  Thanks for adding this!  I've added a small nitpick request.

PR #1378, opened 2023-05-01
Prune tx struct fields and rm witness Call

ChihChengLiang:
### Description

We tried to remove some unused tx structs.

### Issue Link

#528 

### Type of change

### Contents

- Remove zkevm-circuits/src/witness/call.rs and the `Call` struct
- Remove fields in circuit_input_builder Transaction struct. Reuse Geth type Transaction struct whenever possible
- Remove duplicated call data gas cost computation
- Add APIs to get "To" address so that we know when we want contract address or zero address.

### Rationale

We try to start the de-duplication from the low-hanging fruits. So I didn't remove the witness.rs Transaction struct in this PR.

### How Has This Been Tested?

Local build tests and quick tests


ed255 (approved):
LGTM!  Nice pruning!

KimiWu123:
Nice work!

bus-mapping/src/circuit_input_builder/execution.rs
```diff
@@ -123,6 +101,12 @@ pub enum ExecState {
     EndBlock,
 }
 
+impl Default for ExecState {
+    fn default() -> Self {
+        ExecState::Op(OpcodeId::INVALID(0))
+    }
```
lispc, line 107: i know this code is from the old ExecStep::default().  But INVALID(0) is quite anti intuition. I would suggest 

(1) change to OpcodeId::INVALID(0xef)
(2) change to OpcodeId::STOP ( it is 0x0). // i would prefer this 

```
ExecStep
```

lispc (approved):

PR #1390, opened 2023-05-04
Unified cell manager (integrated with EVM circuit)

leolara:
### Description

Create unified Cell Manager and integrate with EVM circuit (Keccak in other PR).

### Type of change

- Refactor

### Contents

- Unified Cell Manager.
- Strategy that mimics the previous EVM circuit cell manager.
- Integration with EVM circuit
- Removing EVM circuit specific Cell Manger

Not included, for another PR:

 - Keccak compatible Cell Manager strategy
 - Integration with keccak 

### Rationale

https://hackmd.io/_RIuqSZSTp-p8fZ4ohbyVg

### How Has This Been Tested?

Unit tests pass


zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,269 @@
...
+    poly::Rotation,
+};
+
+use crate::{
+    evm_circuit::{table::Table, util::CachedRegion},
+    util::query_expression,
+};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+/// CellType
```
ed255, line 17: I know that a lot of these types and methods have been just moved from `zkevm-circuits/src/evm_circuit/util.rs` and were `pub(crate)` and didn't have much documentation.

Now many types and functions here are `pub` and you added some documentation lines which is just the type/function name.  I think it would be a very good opportunity to have a better documentation of the CellManager!  One or two lines for each type/function would be great.

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,269 @@
...
+        }
+    }
+
+    pub fn expr<F: Field>(&self, meta: &mut ConstraintSystem<F>) -> Expression<F> {
+        query_expression(meta, |meta| meta.query_advice(self.advice, Rotation::cur()))
+    }
+}
+
+/// CellManagerStrategy
+pub(crate) trait CellManagerStrategy {
```
ed255, line 150: This trait should work for all circuits right?  But I see that `query_cell` uses an argument `CellManagerColumns`, which depends on `CellType`, which currently is specific to the EVM Circuit.  Do you think that `CellType` could work with the other circuits?  If not, do you have any idea on how to address this?

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,269 @@
...
+impl CellManagerColumns {
+    pub fn add_column(&mut self, cell_type: CellType, column: Column<Advice>) {
+        let idx = self.columns_list.len();
+        let cell_column = CellColumn::new(column, cell_type, idx);
+        self.columns_list.push(cell_column.clone());
+        if let Some(columns) = self.columns.get_mut(&cell_type) {
+            columns.push(cell_column);
+        } else {
+            self.columns.insert(cell_type, vec![cell_column]);
+        }
```
ed255, line 183: This is a nitpick, but the `HashMap` has a nice API to do update/insert: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry
So this could be:
```rust
self.columns.entry(&cell_type).and_modify(|columns| columns.push(cell_column)).or_insert(vec![cell_column]);
```

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,269 @@
...
+            None
+        }
+    }
+
+    pub fn columns(&self) -> Vec<CellColumn> {
+        self.columns_list.clone()
+    }
+
+    pub fn get_width(&self) -> usize {
+        self.columns.values().fold(0, |acc, g| acc + g.len())
```
ed255, line 207:
```suggestion
        self.columns_list.len()
```
This should be the same right?

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -91,29 +128,212 @@ pub(crate) struct CellManager<F> {
...
+        count: usize,
+    ) -> Vec<Cell<F>> {
+        let mut acc = vec![];
+
+        for _ in 0..count {
+            let cell = self.query_cell(meta, cell_type);
+            acc.push(cell);
+        }
+
+        acc
```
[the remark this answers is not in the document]
ed255: This should be even simpler:
```rust
(0..count).map(|_| self.query_cell(meta, cell_type)).collect()
```

zkevm-circuits/src/util/cell_manager_strategy.rs
```diff
@@ -0,0 +1,244 @@
...
+        if let Some(v) = self.0.get_mut(&cell_type) {
+            v.push(advice);
+        } else {
+            self.0.insert(cell_type, vec![advice]);
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct CMFixedWidthStrategy {
```
ed255, line 22: I think this file should be documented via comments.  Maybe you can skip `CMFixedHeigthStrategy` because it's unfinished, but `CMFixedWidthStrategy` contains many things that I think should be documented like:
- What is `perm_substitution`?
- What is `next`?
- How does this strategy work?

zkevm-circuits/src/util/cell_manager_strategy.rs
```diff
@@ -0,0 +1,244 @@
...
+        let column = columns
+            .get_column(cell_type, column_idx)
+            .expect("column not found");
+
+        let cell = Cell::new_from_cs(meta, column.advice, column.idx, self.height_offset + row);
+
+        column_idx += 1;
+        if column_idx >= columns.get_cell_type_width(cell_type) {
+            column_idx = 0;
+            row += 1;
```
ed255, line 112: Maybe I have missed it: there should be a parameter for max height, so the number of rows doesn't grow indefinitely.  Where is it?

zkevm-circuits/src/util/cell_manager_strategy.rs
```diff
@@ -0,0 +1,244 @@
...
+                    next.1
+                } else {
+                    next.1 + 1
+                }
+            })
+            .max()
+            .unwrap_or(0)
+    }
+
+    type Stats = BTreeMap<CellType, (usize, usize, usize)>;
```
ed255, line 135: I suggest moving this to the line after `impl CellManagerStrategy for CMFixedWidthStrategy {` to be consistent with the codebase.

ed255 (commented):
Overall looks good!  I've left a few comments.  The code suggestions are nitpicks so feel free to ignore them if you don't like my suggestions :P

My main request is to add comments to the CellManager, both the new and old code, so that it's better documented.

zkevm-circuits/src/util/cell_manager_strategy.rs
ed255: For reference, this is how the old implementation dealt with this situation: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/73b2aebca9400ac8b31c9394fb726690dbfe0b74/zkevm-circuits/src/evm_circuit/util.rs#L440

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -537,9 +539,13 @@ pub(crate) struct DynamicSelectorHalf<F> {
 }
 
 impl<F: Field> DynamicSelectorHalf<F> {
-    pub(crate) fn new(cell_manager: &mut CellManager<F>, count: usize) -> Self {
-        let target_pairs = cell_manager.query_cells(CellType::StoragePhase1, (count + 1) / 2);
-        let target_odd = cell_manager.query_cell(CellType::StoragePhase1);
+    pub(crate) fn new(
+        meta: &mut ConstraintSystem<F>,
+        cell_manager: &mut CellManager<CMFixedWidthStrategy>,
```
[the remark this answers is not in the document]
leolara: Previously meta was borrowed by the constructor and Cell were pre-queried, this is more efficient and make the code simpler, I think. Now the CB has meta and the circuits call this through CB, so does not need to change much code.

zkevm-circuits/src/util/cell_manager.rs
leolara: So, the idea is that the keccak circuit will have a default CellType, but in general the CellType concept is necessary to make it general.

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,119 @@
...
     fn get_height(&self) -> usize;
 }
+
+#[derive(Default)]
+pub(crate) struct CellManagerColumns {
+    columns: HashMap<CellType, Vec<Column<Advice>>>,
+}
+
+impl CellManagerColumns {
+    pub fn add_column(&mut self, cell_type: CellType, column: Column<Advice>) {
```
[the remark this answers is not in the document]
leolara: Because with the cell manager a column is used by different signals, this would not be useful. We need to create a MockProver that is Cell or signal aware to put annotations that would make sense. We talk about this with @ed255

zkevm-circuits/src/util/cell_manager.rs
leolara: Great!! Didn't know. I think it is good to say this things to spread the knowledge, not nitpick

zkevm-circuits/src/util/cell_manager.rs
leolara: Yes, at the beggining there were no colums_list (btw, if you think of a better name...)

zkevm-circuits/src/util/cell_manager_strategy.rs
leolara: So, the max height it was necessary before, and was used only, for creating all the cells at the beggining in the constructor, now it is not necessary.

We can add a limit with assert if you think it is necessary.

zkevm-circuits/src/util/cell_manager_strategy.rs
leolara: Done

zkevm-circuits/src/util/cell_manager.rs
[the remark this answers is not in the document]
leolara: Done

zkevm-circuits/src/util/cell_manager_strategy.rs
leolara: I have added doc comments.

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -14,50 +19,76 @@ pub(crate) enum CellType {
...
+#[derive(Clone, Debug)]
 /// Cell
-pub struct Cell<F> {
-    _p: PhantomData<F>,
+pub(crate) struct Cell<F> {
+    pub(crate) expression: Expression<F>,
+    pub(crate) column_expression: Expression<F>,
+    pub(crate) column: Column<Advice>,
+    pub(crate) column_idx: usize,
+    pub(crate) rotation: i32,
```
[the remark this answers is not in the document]
leolara: rotation in a query an offset relative to the 0 rotation. So for each polynomial constrain, it has to be 0 for every row. The polinomial constrain contains queries of the form (column, rotation) that act as variables, where rotation 0 means the current row, +1 the next one, -1 the previous one and so on...

It is just like excel 😆

ChihChengLiang:
I removed my review request since we have already two reviewers on this PR.

leolara:
Sorry I just read @ed255 mention to document the old code, which code do you mean?

zkevm-circuits/src/util/cell_manager_strategy.rs
ed255: I think we should add this parameter and corresponding assertion for safety.  Extra rotations have some cost (even if it's small), and we should control how high is the highest ExecStep.  With this maximum value we will be aware of changes that require increasing the max high step, because the corresponding PR will have to update this parameter.  And to start we can use the same max value as before.

ed255 (approved):
LGTM!  Sorry for the delay on continuing the review for this PR!  Nice work!

I've left one request to add an assert to check for a maximum height https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1390#discussion_r1238591211  I'm approving the PR already so that you can merge it once you apply this change! (or if you disagree with the request, let me know!)

zkevm-circuits/src/util/cell_manager.rs
```diff
@@ -1,8 +1,286 @@
...
+/// Cell is a (column, rotation) pair that has been placed and queried by the Cell Manager.
+pub(crate) struct Cell<F> {
+    pub(crate) expression: Expression<F>,
+    pub(crate) column_expression: Expression<F>,
+    pub(crate) column: Column<Advice>,
+    pub(crate) column_idx: usize,
+    pub(crate) rotation: usize,
+
+    // DEPRECATED: the same value as column_idx
+    pub(crate) cell_column_index: usize,
```
ChihChengLiang, line 66: explicitly mark it deprecated?
```suggestion
    #[deprecated(note="the same value as column_idx")]
    pub(crate) cell_column_index: usize,
```

ChihChengLiang (approved):
LGTM.

zkevm-circuits/src/util/cell_manager.rs
leolara: Done

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -29,62 +23,9 @@ pub(crate) mod memory_gadget;
...
-                )
-            },
-            self.column,
-            offset + self.rotation,
-            || value,
-        )
-    }
-}
+// Removing this would require to edit almost all gadget
+pub(crate) use crate::util::cell_manager::{Cell, CellType};
```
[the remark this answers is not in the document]
leolara: Done

leolara:
@ed255 I have added the max_height to the FixedWitdth strategy, now it will panic if the max_height is reached.

leolara:
@ChihChengLiang so after marking them as deprecated, clippy does not pass, because emits an error in each use

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
ChihChengLiang, line 1: In line 83, if we replace cell_column_index with column_idx, we can remove all the use of `cell_column_index` and the field itself completely. So no deprecated field will be referenced in the codebase.
```diff
-cb.curr.cell_manager.columns()[memory_offset.cell_column_index].cell_type
+cb.curr.cell_manager.columns()[memory_offset.column_idx].cell_type
```

ChihChengLiang (commented):
Hi Leo,
I found that we can remove `cell_column_index` completely.
If that doesn't work, feel free to remove the deprecation macro.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
leolara: ok, that seems to work. I am confused why the deprecated tag makes clippy fail for the field but not for the reexport.

ChihChengLiang: The Clippy failed because we disallow warnings in the CI with the `-D warning`.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/9e974d94e8491b482fc6293d4ab16b979cf346a2/.github/workflows/lints.yml#L66

I experimented with it locally in the Makefile. If we remove the `-D warning` flag, Clippy is happy. Otherwise, it would be reported as an error.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/9e974d94e8491b482fc6293d4ab16b979cf346a2/Makefile#L15

leolara: Yes, my surprise is that it does not fail with the deprecated for the reexport, just for the field

Issue #1388, opened 2023-05-03
Write Word utilities, types and abstractions for word lo/hi


ed255:
Before refactoring all circuits to work with word lo/hi we need to create new utilities, types and abstractions for all circuits to deal with Words in a clean way.

This is a non-exhaustive list of topics to cover:
- A type to encapsulate 2 columns that represent the word_lo and word_hi
- A type to encapsulate 2 Cells that represent the word_lo and word_hi
- ConstraintBuilder functions / Gadget to generate constraints like
  - constraint two words are equal
  - test if two words are equal
  - constraint a word is zero
  - test if a word is zero
  - test if a word is not zero
  - test if a word is smaller than another one
  - etc.
- Helper utilities to go from:
  - word to bytes
  - bytes to word
  - word to limbs
  - limbs to word

Some months ago Haichen started looking into this, we can use his previous unfinished work as inspiration:
  - https://github.com/scroll-tech/zkevm-circuits/commit/e2f73af7956f41f3875901241c27be798a9a60ff
  - https://github.com/scroll-tech/zkevm-circuits/commit/f3eb9e8b9040bc1a69d82db19f9acae83f388138

hero78119:
### Rationale

Defined `WordN`, where N represent number of limbs, where the word size is EVM word 256bits. For instance, `Word4` means 4 limbs, each limb 64 bits. Giving 2 word word_n1, word_n2, and n1 % n2 = 0, There are util function to convert n1 to n2. Words related utility are defined in new file `word.rc`

constraint builder also introduce few util apis
- util functions to support query `WordN` cells. For N=32, limb cell integrate with byte lookup. For N < 32, need caller to have range check carefully.  
- word equality check
- separate read/write api to hint which need careful range check

#### Range check strategy
State circuit will assure read value match with last write. Therefore, for lookup table except `stack.pop`, such as txtable, blocktable, callcontext, we need to assure value `write` got proper range check carefully. For read part, range check can be skip by purpose.

Special note for `stack.pop`, since some arithmetcis operation might need special range for operand from stack, and stack can be any value. Therefore, some case read part also need range check carefully, unless it's just some equality check, like `CmpGadget`, where the value range is not matter, such kind of operation can skip range check.

> TODO Any other case need range check during read?


### Pending tasks
- initial version of util function
[...]

PR #1394, opened 2023-05-08
word hi/lo utilities (initial version)

hero78119:
### Description

[_PR description_]

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1388

### Type of change

### Contents
- [x] define general type `Word` and `WordLimbs` along with util function to replace RandomLinearCombination in `word.rs`
- [x] evm util function/common gadgets switch to new Word type 
- [ ] (Ongoing, partially done) EVM circuit switch to new Word type

Pending List 

### Rationale

Defined `WordN`, where N represent number of limbs, where the word size is EVM word 256bits. For instance, `Word4` means 4 limbs, each limb 64 bits. Giving 2 word word_n1, word_n2, and n1 % n2 = 0, There are util function to convert n1 to n2. Words related utility are defined in new file `word.rc`

constraint builder also introduce few util apis
- util functions to support query `WordN` cells. For N=32, limb cell integrate with byte lookup. For N < 32, need caller to have range check carefully.  
- word equality check
- separate read/write api to hint which need careful range check

#### Range check strategy
State circuit will assure read value match with last write. Therefore, for lookup table except `stack.pop`, such as txtable, blocktable, callcontext, we need to assure value `write` got proper range check carefully. For read part, range check can be skip by purpose.

Special note for `stack.pop`, since some arithmetcis operation might need special range for operand from stack, and stack can be any value. Therefore, some case read part also need range check carefully, unless it's just some equality check, like `CmpGadget`, where the value range is not matter, such kind of operation can skip range check.

> TODO Any other case need range check during read?

### Pending tasks
- [ ] evm circuit switch to word type. It will be finished in another evm circuit pr. In this pr have incomplete evm circuit work,  just to verify the usage of utility functions.
- [ ] optimise cell comsumption, e.g. U64 type, such as `gas` type which just need 8 bytes cells range check. For example, refer `CommonCallGadget` gas type.
- [ ] support byte16 range check 
- [ ] review common_gadget all range check.
- [ ] review reversible info range check
- [ ] review word::from_lo_uncheck range check
- [ ] optimise range check for lookup read, see comment https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394#discussion_r1187395854
- [ ] replace `Word<Expression<F>>` with generic `WordExpr<F>`  in lookup so can save `to_word()` conversion in most of gadget
- [ ] static assertion on const generic check
- [ ] review memory to assure if length is 0 then offset can be larger than u64

### How Has This Been Tested?

- [ ] compile pass first
- [ ] all unittest pass
<hr>


zkevm-circuits/src/evm_circuit/execution/add_sub.rs
```diff
@@ -50,9 +50,18 @@ impl<F: Field> ExecutionGadget<F> for AddSubGadget<F> {
...
+            is_sub.expr().0,
+            c.expr().to_word(),
+            a.expr().to_word(),
+        ));
+        cb.stack_pop(b.expr().to_word());
+        cb.stack_push(Word::select(
+            is_sub.expr().0,
+            a.expr().to_word(),
+            c.expr().to_word(),
+        ));
```
ed255, line 64: This opcode does POP on a and b, and PUSH on c.
This means that `a` and `b` are read and `c` is written.
The StateCircuit guarantees that any read is consistent with the previous write.  From this we can optimize out the range checks on the reads, and only leave the range checks on the writes!

So we can directly read `a` and `b` from the stack and assume their corresponding `lo,hi` are 128 bits each.  From this we create `c` and prove that `c.lo` and `c.hi` are 128 bit each, and then write `c` to the stack.

In this case the native format for the `AddWordsGadget` is `lo, hi` so I think there's no need to express `a` and `b` as 8 bit limbs at any time.  `AddWordsGadget` doesn't do any range check on `sum.lo, sum.hi` so those ones do need to be split into limbs to guarantee that `sum.lo, sum.hi` are 128 bit each.

Note that this is an optimization, so we may apply it in the future.  I see that you also implemented `cb.query_word` so, so this is just a comment :)

zkevm-circuits/src/evm_circuit/table.rs
```diff
@@ -311,13 +313,17 @@ impl<F: Field> Lookup<F> {
                 field_tag,
                 index,
                 value,
-            } => vec![id.clone(), field_tag.clone(), index.clone(), value.clone()],
+            } => [
+                vec![id.clone(), field_tag.clone(), index.clone()],
+                value.limbs.to_vec().clone(),
+            ]
+            .concat(),
```
ed255, line 320: Another option for this would be:
```rust
vec![id.clone(), field_tag.clone(), index.clone(), value.lo().clone(), value.hi().clone()]
```
What do you think about this form?

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -499,10 +499,162 @@ impl<F: FieldExt, const N: usize> Expr<F> for RandomLinearCombination<F, N> {
...
+}
+
+impl<F: FieldExt, const N: usize> WordCells<F, N> for WordLimbs<Cell<F>, N> {
+    fn assign<const N1: usize>(
+        &self,
+        region: &mut CachedRegion<'_, '_, F>,
+        offset: usize,
+        bytes: Option<[u8; N1]>,
+    ) -> Result<Vec<AssignedCell<F, F>>, Error> {
+        assert_eq!(N1 % N, 0); // assure N|N1
```
ed255, line 551: Maybe we can use https://docs.rs/static_assertions/latest/static_assertions/ for these assertions that should be performed at compile time?

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,8 +399,39 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
+    // default query_word is 2 limbs
+    pub(crate) fn query_word<const N: usize>(&mut self) -> Word<Cell<F>> {
+        Word::new(
+            self.query_cells(CellType::StoragePhase1, N)
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    pub(crate) fn query_word4<const N: usize>(&mut self) -> Word4<Cell<F>> {
```
ed255: Document that the resulting word is not range checked.  (each limb is not guaranteed to be 64 bits)

zkevm-circuits/src/evm_circuit/util/math_gadget/is_equal_word.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    pub(crate) fn assign(
+        &self,
+        region: &mut CachedRegion<'_, '_, F>,
+        offset: usize,
+        lhs_lo: F,
+        lhs_hi: F,
+        rhs_lo: F,
+        rhs_hi: F,
```
ed255, line 49: I think in both `assign` and `assign_value` we could have the arguments `lhs: Word<F>, rhs: Word<F>`, and in `assign_value` with `Word<Value<F>>`.  The same applies to `IsZeroWordGadget`

zkevm-circuits/src/evm_circuit/util/math_gadget/mul_add_words.rs
```diff
@@ -59,15 +63,15 @@ impl<F: Field> MulAddWordsGadget<F> {
 
         let mut a_limbs = vec![];
         let mut b_limbs = vec![];
-        for trunk in 0..4 {
-            let idx = (trunk * 8) as usize;
-            a_limbs.push(from_bytes::expr(&a.cells[idx..idx + 8]));
-            b_limbs.push(from_bytes::expr(&b.cells[idx..idx + 8]));
+        let word4_a: Word4<Expression<F>> = a.expr().to_wordlimbs();
```
ed255, line 66: This looks really nice!  I think having these gadgets take `Word32` and then converting to a different number of limbs as necessary makes the refactor look very clean!

ed255 (commented):
Overall the abstractions look really good to me!  I see that you focused a lot on the EVM Circuit.
For example, all the word types are in `zkevm-circuits/src/evm_circuit/util.rs` but they will also be useful outside of the `evm_circuit`.  I think a next step will be to focus on how these types can be used in other circuits :)

The other part that needs to be explored is the tables and lookups.  I see you changed `zkevm-circuits/src/table.rs` a bit.  I'm looking forward to how these abstractions can be used on a circuit that contains a table that the evm circuit looks up to.

ed255:
Also I think it's a very good idea for now to mechanically translate every word_rlc to Word32.  That should work all the time, and we can focus on some optimizations like https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394#discussion_r1187395854 in the future.  No need to worry about it now.

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
                 Transition::{Delta, To},
             },
             memory_gadget::{MemoryAddressGadget, MemoryExpansionGadget},
```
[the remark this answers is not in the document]
ed255: I think `Word::from_lo()` is still useful, but we should only use it when we can guarantee that the value is < 128 bits.  I think there are cases where we can guarantee that like in `ComparatorGadget`.

From the current PR, I think `MemoryAddressGadget` should take `Word32` instead of `[Cell<F>; N_BYTES_MEMORY_ADDRESS]` to keep equivalency to the previous code.

zkevm-circuits/src/evm_circuit/util.rs
hero78119: Yes I plan to use this. However, `static_assertion` this library do not work on const generic, see issue here
https://github.com/nvzqz/static-assertions-rs/issues/40
I will try the workaround later.

zkevm-circuits/src/evm_circuit/table.rs
hero78119: I like `value.lo().clone(), value.hi().clone()` style, just I was thinking if using `lo(), hi()` here then once change to different field we need to modify this part. Although it just small change though.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,8 +399,39 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
     }
 
-    pub(crate) fn query_word_rlc<const N: usize>(&mut self) -> RandomLinearCombination<F, N> {
-        RandomLinearCombination::<F, N>::new(self.query_bytes(), self.challenges.evm_word())
+    // TODO remove me
+    // pub(crate) fn query_word_rlc<const N: usize>(&mut self) -> RandomLinearCombination<F, N> {
+    //     RandomLinearCombination::<F, N>::new(self.query_bytes(), self.challenges.evm_word())
+    // }
+
+    // default query_word is 2 limbs
```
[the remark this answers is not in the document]
hero78119: I introduce a small tricks here, so `query_word` will always input a larger limbs, and internally it will constrains new word expression is equals.
Any thought on this design :) ?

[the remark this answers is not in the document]
ed255: Is this trick pushed in the PR branch?

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,8 +400,65 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
+    pub(crate) fn query_word<const N: usize, const N2: usize>(
+        &mut self,
+        word_constrians: WordLimbs<Cell<F>, N2>,
+    ) -> Word<Cell<F>> {
+        let new_word = self._query_word();
+        self.require_equal(
+            "wordlimb equality",
+            word_constrians.expr().is_eq(&new_word.expr()),
```
ed255, line 416: I noticed something here that generates confusion to me.
We have the `Expr` trait with the `expr()` method that returns `Expression<F>`
But `Word` implements `expr()` but instead of returning `Expression<F>` it returns `Word<Expression<F>>`.
So every time I see `.expr()` I think of a single expression, but here we have two` (the low and high).

So I would suggest finding a different name.  Maybe `word_expr()`?

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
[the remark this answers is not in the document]
ed255: It looks a bit strange that `query_word` takes an argument.
Perhaps we could have these functions:
- query_word_unchecked
- query_word4_unchecked
- query_word16
- query_word32

And for gadgets that require word in lo/hi format with range checks we do:
```rust
let a = cb.query_word32();
Gadget::construct(a.to_word());
```
And for gadgets that don't require the check we do:
```rust
let a = cb.query_word_unchecked();
Gadget::construct(a.clone());
```

Another issue I find in the current form is that we allocate 2 extra cells that are not really needed, because the lo/hi can be expressions from the 16 lo limbs and 16 hi limbs.

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -47,12 +47,12 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
...
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let mstart = cb.query_cell_phase2();
-        let msize = cb.query_word_rlc();
+        let mstart = cb.query_word32();
+        let msize = cb.query_word32();
 
         // Pop mstart_address, msize from stack
-        cb.stack_pop(mstart.expr());
-        cb.stack_pop(msize.expr());
+        cb.stack_pop(mstart.expr().to_word());
```
KimiWu123, line 54: This part looks strange to me. It queries a `word32` cell, convert to `expr` and then convert to `word`. If we don't declare the local var, `mstart`, then this line would look like 
```
cb.stack_pop(cb.query_word32().expr().to_word());
```
Is it possible to do like
```
cb.stack_pop(mstart.to_word());
```

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -499,10 +499,179 @@ impl<F: FieldExt, const N: usize> Expr<F> for RandomLinearCombination<F, N> {
...
+                .collect()
+        })
+    }
+
+    fn expr(&self) -> WordLimbs<Expression<F>, N> {
+        return WordLimbs::new(self.limbs.map(|cell| cell.expr()));
+    }
+}
+
+// `Word`, special alias for Word2.
```
KimiWu123, line 567: I'm a little bit confused, I thought the number `2` stands for the number of byte. So, `Word2` is a 2 bytes data structure, isn't it? And from my understanding, word hi/lo is to represent 32 bytes evm word. Could you please explain it more? thanks

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -499,10 +499,179 @@ impl<F: FieldExt, const N: usize> Expr<F> for RandomLinearCombination<F, N> {
     }
 }
 
-pub(crate) type Word<F> = RandomLinearCombination<F, 32>;
+#[derive(Clone, Debug)]
+pub(crate) struct WordLimbs<T, const N: usize> {
```
KimiWu123, line 503: I found all the `Word<F>` became `Word<Cell<F>>`. Wondering if we could add `Cell` in `WordLimbs` just like original `RandomLinearCombination` does. The code would look more succint, then it'll look like,
```
pub(crate) struct WordLimbs<T, const N: usize> {
    pub limbs: [T; N],
    pub(crate) cells: [Cell<F>; N],
}

File: zkevm-circuits/src/evm_circuit/util/common_gadget.rs [hunk 13]
```diff
@@ -366,8 +371,8 @@ impl<F: Field> TransferWithGasFeeGadget<F> {
         receiver_address: Expression<F>,
         receiver_exists: Expression<F>,
         must_create: Expression<F>,
-        value: Word<F>,
-        gas_fee: Word<F>,
+        value: Word32<F>,
```
KimiWu123, line 374: Above [line#276](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394/files#diff-08d58ebc59ab6c11bf7c049e859441eed2319a03feb3fdffaddaba73f8a8c829R276), the type changed from `Vec<Word<F>>` to `Vec<Word32<Cell<F>>>` but here, it changed from `Word<F>` to `Word32<F>`. Is it possible to make it consistent? It'll be easier for ppl to refactor other circuits.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,8 +399,39 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
+    pub(crate) fn query_word16<const N: usize>(&mut self) -> Word16<Cell<F>> {
+        // TODO figure out Word16 for range check
+        Word16::new(
+            self.query_cells(CellType::StoragePhase1, N)
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    pub(crate) fn query_word32(&mut self) -> Word32<Cell<F>> {
```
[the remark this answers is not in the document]
hero78119: Done with tuned comments `each limb is 8 bits, and any conversion to smaller limbs inherits the type check.` :)

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,8 +399,39 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
+
+    pub(crate) fn query_word4<const N: usize>(&mut self) -> Word4<Cell<F>> {
+        Word4::new(
+            self.query_cells(CellType::StoragePhase1, N)
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    pub(crate) fn query_word16<const N: usize>(&mut self) -> Word16<Cell<F>> {
```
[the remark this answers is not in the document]
hero78119: Add documentation with TODO to implement 16 bits range check soon

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
hero78119: Good idea, renamed to `word_expr()` for clarification

zkevm-circuits/src/evm_circuit/util.rs
hero78119: I moved the new `Word` type to new file `word.rs`, and add simple comment.

Word`N` means splits 256 bit into N chunk. Each chunk is also called limb. (would chunk be better naming 😢 ?)

Therefore, Word2 means 2 limb, each with 128bit

zkevm-circuits/src/evm_circuit/util.rs
hero78119: In new `Word` generic type, there is no single expression to represent multiple cells via RLC. So `RandomLinearCombination ` concept is no longer be used here.

Besides, I try to avoid putting limbs and cells together, I think it somehow violate pattern for single-responsibility of a class definition.

But, naming shorter is indeed better!! so i add few type alias to try make common type more shorter comfortable :P

ed255:
BTW, currently this PR targets the `main` branch.  Merging this to main will probably not be possible due to many tests not passing (or even the code not compiling).  I created a [`word-lo-hi` branch](https://github.com/privacy-scaling-explorations/zkevm-circuits/tree/word-lo-hi) that we can use to keep the development of the word lo/hi refactor, so this PR could be updated to target that branch; and once people start working on refactoring the circuits, they can also target that branch.  This way we can still review everything via PRs, while the CI is not passing, without breaking main.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
hero78119: Fixed compile error in this file

zkevm-circuits/src/evm_circuit/util.rs
KimiWu123: haha, got it. I thought the other way around. I thought `Word32` includes 32 limbs and each limb is 1 byte. So, `Word2` is a 2 bytes data.

KimiWu123: What if we name these types like this,
- `Word` and `Word128` stands for 2 x 128 bits limbs
- `Word64` stands for 4 x `64` bits limbs
- `Word16` stands for 16 x `16` bits limbs
- `Word8`   stands for 32 x `8` bits limbs

Just like `uintN` in solidity, the following `N` stands for bit numbers

hero78119:
> Also I think it's a very good idea for now to mechanically translate every WordRLC (from query_word_rlc) to Word32. That should work all the time, and we can focus on some optimizations like [#1394 (comment)](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394#discussion_r1187395854) in the future. No need to worry about it now.

After consolidate concept for range check on `write` timing only, everything become more clear. We can apply optimisation for read part soon 👍

zkevm-circuits/src/evm_circuit/execution/logs.rs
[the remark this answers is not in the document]
hero78119: Have few change
- rename `from_lo()` to `from_lo_unchecked()`
- add TODO for write part range check 

I think we can separate read and write part instead of single function `is_write: bool`. Then in read part we can use `from_lo_unchecked()` safely, and only range check on write part.  If this plan sound good, I will proceed this change soon

hero78119:
> Why not rename `Word` to `Word2`? I think that is more consistent with Word4, Word32, etc..  

'Word' is like default

later we can apply dereference if change to smaller fields , for example goldilocks, which is word4 will be default. Have default word make it happend without need tk change most of the codebaae

zkevm-circuits/src/evm_circuit/util.rs
ed255: I want to say that I like the original nomenclature that @hero78119 proposed (where `N` is the number of limbs).
If `N` means number of bits, I find it easier to mistake it for the word size, like the `uintN` example; but that's not the case: all Words are 32 bytes in Ethereum.
Also I think it's more direct to see `WordN` and then inspect the type and see an array of length `N`.
The other way would be inspecting the type and seeing an array of length `256 / N`

zkevm-circuits/src/util/word.rs
```diff
@@ -0,0 +1,206 @@
...
+            .limbs
+            .chunks(N1 / N2)
+            .map(|chunk| from_bytes::expr(chunk))
+            .collect_vec()
+            .try_into()
+            .unwrap();
+        WordLimbs::<Expression<F>, N2>::new(limbs)
+    }
+
+    pub fn to_word(&self) -> Word<Expression<F>> {
```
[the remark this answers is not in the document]
hero78119: `to_word` is a alias for `to_word2`. Here source and destination already be expression.

But thanks for the nice calls, I think `to_wordN` is better than `to_wordlimbs`. Will try to rename it

zkevm-circuits/src/evm_circuit/execution/add_sub.rs
hero78119: Hi @ed255 This is a good consolidation for when should do range check! And it definitely can be applied on other lookup with tag, since all value should be checked during write.

Just have small add-on, this `add-sub.rs` is pretty typical that we can just assume value pop from stack it's low/hi is within 128 bits and fit optimisation! But there are some other edge cases, for example value pop from stack must be an address, therefore at least 20 bytes. I applied partial optimisation on address type as well in latest commits. We can apply same tricks on other type, such as U64 range etc.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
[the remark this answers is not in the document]
ed255: > Btw, where is `2 extra cells` :) ? I dont really get it

Ah that was before, when `qwery_word` took an argument.  It looked like this:
```
let word32 = cb.query_word32(); // Perform 32 queries
let word = cb.query_word(word32); // Perform 2 extra queries for lo/hi, and constraint them to be eq to the word32
```

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -568,32 +577,34 @@ impl<F: Field> TransferGadget<F> {
...
         self.value_is_zero
-            .assign_value(region, offset, region.word_rlc(value))?;
+            .assign_value(region, offset, Value::known(Word::from_u256(value)))?;
         Ok(())
     }
 }
 
 #[derive(Clone, Debug)]
-pub(crate) struct CommonCallGadget<F, const IS_SUCCESS_CALL: bool> {
+pub(crate) struct CommonCallGadget<F, const IS_SUCCESS_CALL: bool, T> {
```
ed255, line 586: I beliebe this `T` should already be known statically.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -397,14 +417,57 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
 
-    pub(crate) fn query_word_rlc<const N: usize>(&mut self) -> RandomLinearCombination<F, N> {
-        RandomLinearCombination::<F, N>::new(self.query_bytes(), self.challenges.evm_word())
+    // TODO remove me
+    // pub(crate) fn query_word_rlc<const N: usize>(&mut self) -> RandomLinearCombination<F, N> {
+    //     RandomLinearCombination::<F, N>::new(self.query_bytes(), self.challenges.evm_word())
+    // }
+
+    // default query_word is 2 limbs. Each limb is not guaranteed to be 128 bits.
+    pub fn query_word_unchecked<const N: usize, const N2: usize>(&mut self) -> Word<Cell<F>> {
```
ed255, line 426: Please document what is `N` and `N2` in all the query methods that use these associated consts.

zkevm-circuits/src/evm_circuit/util/math_gadget/cmp_words.rs
```diff
@@ -1,48 +1,40 @@
...
-        b: &util::Word<F>,
-    ) -> Self {
-        // `a[0..16] <= b[0..16]`
-        let comparison_lo = ComparisonGadget::construct(
-            cb,
-            from_bytes::expr(&a.cells[0..16]),
-            from_bytes::expr(&b.cells[0..16]),
-        );
+impl<F: Field, T: WordExpr<F>> CmpWordsGadget<F, T> {
+    pub(crate) fn construct(cb: &mut EVMConstraintBuilder<F>, a: T, b: T) -> Self {
```
ed255, line 23: Consider having two types (one for `a` and one for `b`).  In theory we don't need them to be the same, we just need to be able to get the `lo/hi` of each one.

zkevm-circuits/src/util/word.rs
```diff
@@ -0,0 +1,214 @@
...
+    circuit::{AssignedCell, Value},
+    halo2curves::FieldExt,
+    plonk::{Error, Expression},
+};
+use itertools::Itertools;
+
+use crate::evm_circuit::util::{from_bytes, CachedRegion, Cell};
+
+#[derive(Clone, Debug)]
+pub(crate) struct WordLimbs<T, const N: usize> {
```
ed255, line 18: Please add a small documentation line about what is `N` here.  Also below there are some methods that use `N1` and `N2`, add also documentation about what those consts are.

ed255 (approved):
I've done a quick overview.  I think this PR is in a good state to be merged into the `word-lo-hi` branch and continue iterating it with future PRs while we also work on refactoring the other circuits.  So I'm approving it already!

The notes I added are about adding some documentation for the new types and methods, and also I think there are some generics that may be unneeded because they types are already known.

zkevm-circuits/src/evm_circuit/execution/addmod.rs
```diff
@@ -9,45 +9,48 @@ use crate::{
...
 };
 
 use bus_mapping::evm::OpcodeId;
 use eth_types::{Field, ToLittleEndian, U256, U512};
-use halo2_proofs::plonk::Error;
+use halo2_proofs::{circuit::Value, plonk::Error};
 
 #[derive(Clone, Debug)]
-pub(crate) struct AddModGadget<F> {
+pub(crate) struct AddModGadget<F, T> {
```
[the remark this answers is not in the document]
hero78119: Thanks for the review, I remove the unnecessary generics in latest commit!

zkevm-circuits/src/evm_circuit/util/math_gadget/cmp_words.rs
hero78119: This is really nice call!! Type `a, b` is separated

hero78119:
Excepted for documentation, other issues are addressed :P

hero78119:
Updated: just fix few obvious errors in common/math gadget https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394/commits/2a14cd363a7e2f61209ec7082a44946a28931be0

zkevm-circuits/src/evm_circuit/table.rs
hero78119: @ed255 I realized there are `table.annotations()` which still hardcode column to annotation string. So my methodology doenst really help to generalized. I adapt your method 👍

hero78119:
Update: revamp few types and util, e.g. RwTable `storage_key` type to `Word`

Issue #1393, opened 2023-05-04
Activate `circuit-params` feature from halo2


ed255:
After release `v2023_04_20` there's a new feature that allows passing runtime parameters to the `configure` implementation of a `Circuit`.

We can leverage this to remove all the associated const that we have currently in the circuit types like `MAX_TXS`, `MAX_CALLDATA` and instead pick these parameters via a new `param` type.

This feature should also help improving the ergonomics of https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1331

PR #1399, opened 2023-05-10
Use halo2 circuit-params feature

ed255:
Refactor SuperCircuit and PiCircuit to use circuit parameters instead of associated constants.

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1393

### Type of change

### Contents

Using the `circuit-params` feature recently introduced in our fork of halo2 allows simplifying the code for circuits that require runtime parameters at configuration time.  It removes the annoying associated const of the circuit types, and also allows having a single circuit type (no need for an extra circuit test type).

### DIFF
--- a/circuit-benchmarks/src/pi_circuit.rs
+++ b/circuit-benchmarks/src/pi_circuit.rs
@@ -24,7 +24,7 @@ mod tests {
     use rand_xorshift::XorShiftRng;
     use std::env::var;
     use zkevm_circuits::{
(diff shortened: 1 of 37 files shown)


han0110 (approved):
LGTM! It looks better now without a bunch of const generics!

leolara (approved):
LGTM!! This is a great change from a Software Engineering point of view :-)

Issue #1160, opened 2023-02-10
Calling a contract that contains an invalid opcode (e.g. `DD`) panics


adria0:
### What command(s) is the bug in?

`testool --oneliner 'call 10;DD'`

### Describe the bug

```
+--------------------------------------------+--------------------------------------------------------------------+
| coinbase                                   | 0x0000000000000000000000000000000000000000                         |
+--------------------------------------------+--------------------------------------------------------------------+
| difficulty                                 | 0                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| number                                     | 1                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| timestamp                                  | 1                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| prev_hash                                  | 0x0000000000000000000000000000000000000000000000000000000000000000 |
+--------------------------------------------+--------------------------------------------------------------------+
| sk                                         | 0101010101010101010101010101010101010101010101010101010101010101   |
+--------------------------------------------+--------------------------------------------------------------------+
| from                                       | 0x1a642f0e3c3af545e7acbd38b07251b3990914f1                         |
+--------------------------------------------+--------------------------------------------------------------------+
| to                                         | Some(0x0000000000000000000000000000000000000010)                   |
+--------------------------------------------+--------------------------------------------------------------------+
| gas_limit                                  | 10000000                                                           |
+--------------------------------------------+--------------------------------------------------------------------+
| gas_price                                  | 1                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| nonce                                      | 0                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| value                                      | 0                                                                  |
+--------------------------------------------+--------------------------------------------------------------------+
| data                                       |                                                                    |
+--------------------------------------------+--------------------------------------------------------------------+
| exception                                  | false                                                              |
+--------------------------------------------+--------------------------------------------------------------------+
| 0x0000000000000000000000000000000000000010 | balance :0                                                         |
|                                            | code :dd                                                           |
|                                            | nonce :1                                                           |
+--------------------------------------------+--------------------------------------------------------------------+
| 0x1a642f0e3c3af545e7acbd38b07251b3990914f1 | balance :1000000000000000000                                       |
|                                            | code :                                                             |
|                                            | nonce :0                                                           |
+--------------------------------------------+--------------------------------------------------------------------+

[...]

PR #1400, opened 2023-05-10
fix(testool): fix coinbase, solves #1160

adria0:
### Description

When running `testool --oneliner`, set the coinbase to `MOCK_COINBASE` and top up with 1 wei to register it into the state db.

### Issue Link

#1160 

### Type of change


ed255 (approved):
LGTM! Thanks for the fix!  This is a very small PR so I think you can just merge it with one approval :)

Issue #1391, opened 2023-05-04
Merge structs between CIB and evm circuit witness


ChihChengLiang:
### Describe the feature you would like

This is a continuation of unfinished tasks in #528

We explore more merging opportunities between the structs in the circuit input builder(CIB) and the evm circuit witness.

- [ ] Bytecode. The witness has one more field, the bytecode hash. If we add the bytecode hash in CIB's Bytecode struct, then we can safely remove the struct in witness.rs. WIP
- [ ] Block. witness::Block and witness::BlockContext have many overlapping fields with CIB::Block. We can create better structs in CIB and remove some of the witness's block structs.
- [x] ExecStep. Many duplicated fields between witness::ExecStep and CIB::ExecStep, which suggests possible struct merging.
- [ ] Tx. Further deduplication seems still possible. In review https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1457
- [x] Call

ed255:
A bit related to this issue: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1451

PR #1403, opened 2023-05-11
Replace Witness.rs's ExecStep with CIP's

ChihChengLiang:
### Description

- We replace ExecStep struct in witness.rs with the one in the Circuit Input Builder.
- We add some convenient methods in CIP's ExecStep so that we don't create extra struct fields.
- Add `get_rws` method to `witness::Block` to streamline the logic of getting Read-Write records.

### Issue Link

Part of #1391

### Type of change

### Rationale

- Notes for reviewers, one place that might be error-prone would be the rw_indices. After the cleanup in this PR, we can replace the manual counting of rw indices with something like `evm_circuit::utils::StepRws`.


bus-mapping/src/circuit_input_builder/execution.rs
```diff
@@ -86,6 +86,45 @@ impl ExecStep {
...
+    pub fn stack_pointer(&self) -> u64 {
+        1024 - self.stack_size as u64
+    }
+
+    /// The memory size in word **before** this step
+    pub fn memory_word_size(&self) -> u64 {
+        let n_bytes_word = 32u64;
+        let memory_size = self.memory_size as u64;
+        // EVM always pads the memory size to word size
+        // https://github.com/ethereum/go-ethereum/blob/master/core/vm/interpreter.go#L212-L216
```
ed255, line 118:
```suggestion
        // https://github.com/ethereum/go-ethereum/blob/a340721aa909ea4b541ffd1ea5e9c7bd441ff769/core/vm/interpreter.go#L201-L205
```
Updated link: the original was using a reference to `master` which changes over time, and it had already changed, so using a commit will make it stable over time.

zkevm-circuits/src/evm_circuit/step.rs
```diff
@@ -121,6 +125,154 @@ impl Display for ExecutionState {
...
+        }
+    }
+}
+
+pub trait HasExecutionState {
+    fn execution_state(&self) -> ExecutionState;
+}
+
+impl HasExecutionState for ExecStep {
+    fn execution_state(&self) -> ExecutionState {
```
ed255, line 272: I see that the trait `HasExecutionState` is only implemented for `ExecStep`.  Did you implement `execution_state` via a trait in order to keep this implementation in the zkevm-circuits crate?  Did you consider implementing it directly in the `bus-mapping` crate so that no trait is needed?

eth-types/src/evm_types.rs
```diff
@@ -54,12 +54,24 @@ impl ProgramCounter {
 #[derive(Default, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct Gas(pub u64);
 
+impl fmt::Display for Gas {
```
ed255, line 57: I have the feeling that all these wrapper types `ProgramCounter` `Gas` and `GasCost` add verbosity to the code but don't provide many benefits; and using `usize` and `u64` directly will lead to cleaner code.  We would avoid having to do things like `Gas(0)` instead of just `0`.  Also avoid doing `step.gas.0` instead of `step.gas`. What do you think about this?
Note: this comment is unrelated to this PR.

ed255:
The changes look good!  One less duplicated struct :tada: 
I see the PR is in Draft, so maybe you plan to do more updates?

>     * Notes for reviewers, one place that might be error-prone would be the rw_indices. After the cleanup in this PR, we can replace the manual counting of rw indices with something like `evm_circuit::utils::StepRws`.

Sounds great!

ChihChengLiang:
Hey @ed255,

Thanks for the review. I actually got many integration test bugs, so I turned it back to draft to fix it.

bus-mapping/src/circuit_input_builder/execution.rs
ChihChengLiang: Good call!

zkevm-circuits/src/evm_circuit/step.rs
ChihChengLiang: The problem is that `ExecutionState` is only in zkevm-circuit crate, so I can not implement the `execution_state` method for the `ExecStep` in the `bus-mapping` crate.

eth-types/src/evm_types.rs
ChihChengLiang: I feel you strongly on this too! We should remove ProgramCounter, Gas, and GasCost structs in other PR.

zkevm-circuits/src/evm_circuit/step.rs
ed255: I see, thanks for the clarification!

eth-types/src/evm_types.rs
ChihChengLiang: Created #1404 to track this.

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -205,28 +203,27 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
...
 
         let is_persistent = call.is_persistent as u64;
-        let mut topic_stack_entry = if topic_count > 0 {
-            step.rw_indices[6 + call.is_persistent as usize]
-        } else {
-            // if topic_count == 0, this value will be no used anymore
-            (RwTableTag::Stack, 0usize)
-        };
+        let topic_stack_entry =
+            (topic_count > 0).then(|| step.rw_index(6 + call.is_persistent as usize));
```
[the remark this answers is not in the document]
ed255: I think this can be explained by looking at the rw operations that the log does. See https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/14bd7c9ca95a5b0ed50d78750414e8c11f799e8e/bus-mapping/src/evm/opcodes/logs.rs#L53

- 0-1. Stack read
- 2-5. call context read
- 6? tx log write
- 6 / 7. Stack read

zkevm-circuits/src/evm_circuit/execution/logs.rs
```diff
@@ -205,28 +203,27 @@ impl<F: Field> ExecutionGadget<F> for LogGadget<F> {
...
-                topic = region.word_rlc(block.rws[topic_stack_entry].stack_value());
-                self.topic_selectors[i].assign(region, offset, Value::known(F::ONE))?;
-                topic_stack_entry.1 += 1;
-            } else {
-                self.topic_selectors[i].assign(region, offset, Value::known(F::ZERO))?;
-            }
-            self.phase2_topics[i].assign(region, offset, topic)?;
+            let topic = topic_stack_entry
+                .map(|OperationRef(target, index)| {
+                    block.rws[OperationRef(target, index + i)].stack_value()
```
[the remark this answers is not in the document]
ed255: I believe this could be rewritten to use `block.get_rws`. This code is reading the topic words from the stack (which can be 0,1,2,3 or 4; depending on LOG0, LOG1, LOG2, LOG3, LOG4).
And it should be easy to figure out the offset of the first one (it should be `6 + call.is_persistent as usize`).
Probably it will be easier to rewrite the loop like `for i in 0..topic_count` and leave the other values as 0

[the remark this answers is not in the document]
ChihChengLiang: I just saw your comments. Let me think and figure it out.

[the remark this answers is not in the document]
ChihChengLiang: Thank you, @ed255, for the explanation. I know how it works now. Patched at [4945a56](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1403/commits/4945a56dab4b32a906c04f3898de0337f1ebd369)

ed255 (approved):
LGTM!  Nice work, using `block.get_rws` looks cleaner than the previous style :)

han0110 (approved):
LGTM! Nice refactor!

PR #1409, opened 2023-05-14
Retype Gas limit and nonce to u64

### LINKED ISSUE #1377: align gas_limits & nonce from word to u64

Due to `ether-core` lib some basic type defining gas_limits and nonce in Word type, hence in current codebase most of place we are aligned with it. However this two types should be u64 instead.

### Additional context

_No response_

ChihChengLiang:
### Description

- Gas limit from Word to u64
- Nonce from Word to u64

### Issue Link

Fix #1377

### Type of change

### Contents

- change geth_type Account's nonce and gas_limit to eth_core::U64. We can not use simple u64 because we need to serialize nonce and gas_limit to hex value with 0x prefix for Geth RPC to use.
- simple u64 is used in rest of the places.


bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -318,7 +318,7 @@ impl<'a> CircuitInputStateRef<'a> {
         // Perform the write to the account in the StateDB
         if matches!(rw, RW::WRITE) {
             match op.field {
-                AccountField::Nonce => account.nonce = op.value,
+                AccountField::Nonce => account.nonce = op.value.low_u64(),
```
ed255, line 321:
```suggestion
                AccountField::Nonce => account.nonce = op.value.as_u64(),
```
With `as_u64` we get a panic in case the value doesn't fit in u64.  I think this can be useful to avoid silent bugs.

eth-types/src/geth_types.rs
```diff
@@ -171,8 +175,8 @@ impl From<&crate::Transaction> for Transaction {
         Transaction {
             from: tx.from,
             to: tx.to,
-            nonce: tx.nonce,
-            gas_limit: tx.gas,
+            nonce: tx.nonce.low_u64().into(),
+            gas_limit: tx.gas.low_u64().into(),
```
ed255, line 179: I think using `as_u64` is safer.

ed255 (commented):
Overall looks good! Also good catch on simplifying how accounts are set in tests!
I have only a small request, which is replacing the usage of `U256.low_u64()` by `U256.as_u64()` to have safety checks.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: Addressed in [0fc8072](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1409/commits/0fc80723760ada93d2128cae58c6e19a09e8ba62)

ed255 (approved):
LGTM!

Issue #1379, opened 2023-05-03
EVM Circuit: Refactor word_rlc into word lo/hi


ed255:
Specs: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/390

Blocked/Conflicted by:
- [ ] CREATE A https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1356
- [ ] CREATE B https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1357
- [ ] CREATE C https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1358
- [ ] https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1265
- [ ] https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1292

Depends on:
- https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1388

hero78119:
- [ ] adapt lo/hi refactor
- [ ] pass all unittest
   - [ ]  review H256 and U256 endian handling

ChihChengLiang:
@hero78119 Can we move the pending tasks listed here to #1487 and close this issue?

hero78119:
> @hero78119 Can we move the pending tasks listed here to #1487 and close this issue?

Yes! I closed this issue and move related to #1487

PR #1411, opened 2023-05-16
[word-lo-hi] evm circuit 

hero78119:
### Description

This PR is based on https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394
Need to merge https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1394 first before review this.

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1379

### Type of change

### Contents

- [x] fixed most of op compiling errors other than `callop.rs` and `begin_tx.rs`.
- [x] fixed `callop.rs` and `begin_tx.rs` 
- [x] remove all compatible workaround `construct_new` under evm circuit
- [x] unittest under evm circuit all pass `cargo test --features warn-unimplemented --features test --package zkevm-circuits --lib -- evm_circuit::execution::` 
- [x] fix few `word` gadgets generics to take `Word<T>`  instead of `T`  to restrict it flexibility, since it's non sense to put type not related to word
- [x] remove most of `deprecated` api under evm circuits
- [x] add IntDecomposition type as an alternative to RandomLinearComposition, with base 256

#### Storage_1
```
Main:
+-----------------------------------+------------------------------+-------------------------+
| "storage_1" total_available_cells | "storage_1" total_used_cells | "storage_1" Utilization (%) |
+-----------------------------------+------------------------------+-------------------------+
| 25480                             | 6482                         | 25.4                    |
+-----------------------------------+------------------------------+-------------------------+

Word-lo-hi
+-----------------------------------+------------------------------+-----------------------------+
| "storage_1" total_available_cells | "storage_1" total_used_cells | "storage_1" Utilization (%) |
+-----------------------------------+------------------------------+-----------------------------+
| 24080                             | 7078                         | 29.4                        |
+-----------------------------------+------------------------------+-----------------------------+
```

#### Storage_2
```
Main
+-----------------------------------+------------------------------+-------------------------+
| "storage_2" total_available_cells | "storage_2" total_used_cells | "storage_2" Utilization |
+-----------------------------------+------------------------------+-------------------------+
| 1456                              | 467                          | 32.1                    |
+-----------------------------------+------------------------------+-------------------------+

Word-lo-hi
+-----------------------------------+------------------------------+-----------------------------+
| "storage_2" total_available_cells | "storage_2" total_used_cells | "storage_2" Utilization (%) |
+-----------------------------------+------------------------------+-----------------------------+
| 1376                              | 14                           | 1.0                         |
+-----------------------------------+------------------------------+-----------------------------+
```

#### Byte_lookup
```
Main
 +-------------------------------------+--------------------------------+---------------------------+
| "byte_lookup" total_available_cells | "byte_lookup" total_used_cells | "byte_lookup" Utilization |
+-------------------------------------+--------------------------------+---------------------------+
| 8736                                | 6786                           | 77.7                      |
+-------------------------------------+--------------------------------+---------------------------+

Word-lo-hi
+-------------------------------------+--------------------------------+-------------------------------+
| "byte_lookup" total_available_cells | "byte_lookup" total_used_cells | "byte_lookup" Utilization (%) |
+-------------------------------------+--------------------------------+-------------------------------+
| 8256                                | 6566                           | 79.5                          |
+-------------------------------------+--------------------------------+-------------------------------+
```

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/balance.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/balance.rs
@@ -47,11 +47,11 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
         let tx_id = cb.call_context(None, CallContextFieldTag::TxId);
         let mut reversion_info = cb.reversion_info_read(None);
         let is_warm = cb.query_bool();
-        cb.account_access_list_write_word(
+        cb.account_access_list_write(
             tx_id.expr(),
             address.to_word(),
-            Word::from_lo_unchecked(1.expr()),
-            Word::from_lo_unchecked(is_warm.expr()),
+            1.expr(),
+            is_warm.expr(),
             Some(&mut reversion_info),
         );
         let code_hash = cb.query_word_unchecked();
--- a/zkevm-circuits/src/evm_circuit/execution/balance.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/balance.rs
@@ -1,6 +1,7 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
+        param::N_BYTES_ACCOUNT_ADDRESS,
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
@@ -8,29 +9,29 @@ use crate::{
                 ConstrainBuilderCommon, EVMConstraintBuilder, ReversionInfo, StepStateTransition,
                 Transition::Delta,
             },
-            math_gadget::IsZeroGadget,
-            not, select, CachedRegion, Cell, Word,
+            math_gadget::IsZeroWordGadget,
+            not, select, AccountAddress, CachedRegion, Cell,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
     table::{AccountFieldTag, CallContextFieldTag},
     util::{
-        word::{Word32Cell, WordExpr},
+        word::{Word, Word32Cell, WordCell, WordExpr},
         Expr,
     },
 };
-use eth_types::{evm_types::GasCost, Field, ToLittleEndian};
+use eth_types::{evm_types::GasCost, Field, ToLittleEndian, ToWord};
 use halo2_proofs::{circuit::Value, plonk::Error};
 
 #[derive(Clone, Debug)]
 pub(crate) struct BalanceGadget<F> {
     same_context: SameContextGadget<F>,
-    address_word: Word<F>,
+    address: AccountAddress<F>,
     reversion_info: ReversionInfo<F>,
     tx_id: Cell<F>,
     is_warm: Cell<F>,
-    code_hash: Word32Cell<F>,
-    not_exists: IsZeroGadget<F>,
+    code_hash: WordCell<F>,
+    not_exists: IsZeroWordGadget<F, WordCell<F>>,
     balance: Word32Cell<F>,
 }
 
@@ -40,31 +41,35 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
     const EXECUTION_STATE: ExecutionState = ExecutionState::BALANCE;
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let address = cb.query_word_rlc();
+        let address = cb.query_account_address();
         cb.stack_pop_word(address.to_word());
 
         let tx_id = cb.call_context(None, CallContextFieldTag::TxId);
         let mut reversion_info = cb.reversion_info_read(None);
         let is_warm = cb.query_bool();
-        cb.account_access_list_write(
+        cb.account_access_list_write_word(
             tx_id.expr(),
-            address.expr(),
-            1.expr(),
-            is_warm.expr(),
+            address.to_word(),
+            Word::from_lo_unchecked(1.expr()),
+            Word::from_lo_unchecked(is_warm.expr()),
             Some(&mut reversion_info),
         );
-        let code_hash = cb.query_word32();
+        let code_hash = cb.query_word_unchecked();
         // For non-existing accounts the code_hash must be 0 in the rw_table.
         cb.account_read_word(
-            address.expr(),
+            address.to_word(),
             AccountFieldTag::CodeHash,
             code_hash.to_word(),
         );
-        let not_exists = IsZeroGadget::construct(cb, code_hash.expr());
+        let not_exists = IsZeroWordGadget::construct(cb, &code_hash);
         let exists = not::expr(not_exists.expr());
         let balance = cb.query_word32();
         cb.condition(exists.expr(), |cb| {
-            cb.account_read_word(address.expr(), AccountFieldTag::Balance, balance.to_word());
+            cb.account_read_word(
+                address.to_word(),
+                AccountFieldTag::Balance,
+                balance.to_word(),
+            );
         });
         cb.condition(not_exists.expr(), |cb| {
             cb.require_zero_word("balance is zero when non_exists", balance.to_word());
@@ -92,7 +97,7 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
 
         Self {
             same_context,
-            address_word: address,
+            address,
             reversion_info,
             tx_id,
             is_warm,
@@ -114,8 +119,15 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
         self.same_context.assign_exec_step(region, offset, step)?;
 
         let address = block.get_rws(step, 0).stack_value();
-        self.address_word
-            .assign(region, offset, Some(address.to_le_bytes()))?;
+        self.address.assign(
+            region,
+            offset,
+            Some(
+                address.to_le_bytes()[0..N_BYTES_ACCOUNT_ADDRESS]
+                    .try_into()
+                    .unwrap(),
+            ),
+        )?;
 
         self.tx_id
             .assign(region, offset, Value::known(F::from(tx.id as u64)))?;
@@ -133,16 +145,15 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
 
         let code_hash = block.get_rws(step, 5).account_value_pair().0;
         self.code_hash
-            .assign(region, offset, Some(code_hash.to_le_bytes()))?;
+            .assign_u256(region, offset, code_hash.to_word())?;
         self.not_exists
-            .assign_value(region, offset, region.word_rlc(code_hash))?;
+            .assign_value(region, offset, Value::known(Word::from(code_hash)))?;
         let balance = if code_hash.is_zero() {
             eth_types::Word::zero()
         } else {
             block.get_rws(step, 6).account_value_pair().0
         };
-        self.balance
-            .assign(region, offset, Some(balance.to_le_bytes()))?;
+        self.balance.assign_u256(region, offset, balance)?;
 
         Ok(())
     }
--- a/zkevm-circuits/src/evm_circuit/execution/balance.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/balance.rs
@@ -1,7 +1,6 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
-        param::N_BYTES_ACCOUNT_ADDRESS,
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
@@ -20,7 +19,7 @@ use crate::{
         Expr,
     },
 };
-use eth_types::{evm_types::GasCost, Field, ToLittleEndian, ToWord};
+use eth_types::{evm_types::GasCost, Field, ToWord};
 use halo2_proofs::{circuit::Value, plonk::Error};
 
 #[derive(Clone, Debug)]
@@ -119,15 +118,7 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
         self.same_context.assign_exec_step(region, offset, step)?;
 
         let address = block.get_rws(step, 0).stack_value();
-        self.address.assign(
-            region,
-            offset,
-            Some(
-                address.to_le_bytes()[0..N_BYTES_ACCOUNT_ADDRESS]
-                    .try_into()
-                    .unwrap(),
-            ),
-        )?;
+        self.address.assign_u256(region, offset, address)?;
 
         self.tx_id
             .assign(region, offset, Value::known(F::from(tx.id as u64)))?;
--- a/zkevm-circuits/src/evm_circuit/execution/balance.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/balance.rs
@@ -47,7 +47,7 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
         let tx_id = cb.call_context(None, CallContextFieldTag::TxId);
         let mut reversion_info = cb.reversion_info_read(None);
         let is_warm = cb.query_bool();
-        cb.account_access_list_write(
+        cb.account_access_list_write_unchecked(
             tx_id.expr(),
             address.to_word(),
             1.expr(),
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -272,7 +272,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             cb.account_write_word(
                 call_callee_address.to_word(),
                 AccountFieldTag::Nonce,
-                Word::from_lo_unchecked(1.expr()),
+                Word::one(),
                 Word::zero(),
                 Some(&mut reversion_info),
             );
@@ -394,10 +394,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             |cb| {
                 // Setup first call's context.
                 for (field_tag, value) in [
-                    (
-                        CallContextFieldTag::Depth,
-                        Word::from_lo_unchecked(1.expr()),
-                    ),
+                    (CallContextFieldTag::Depth, Word::one()),
                     (
                         CallContextFieldTag::CallerAddress,
                         tx_caller_address.to_word(),
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -15,7 +15,7 @@ use crate::{
                 ConstantDivisionGadget, ContractCreateGadget, IsEqualWordGadget, IsZeroWordGadget,
                 MulWordByU64Gadget, RangeCheckGadget,
             },
-            not, or, select, AccountAddress, CachedRegion, Cell, StepRws, U64Cell,
+            not, or, select, AccountAddress, CachedRegion, Cell, StepRws,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
@@ -38,7 +38,7 @@ use halo2_proofs::{
 #[derive(Clone, Debug)]
 pub(crate) struct BeginTxGadget<F> {
     // tx_id is query in current scope. The range should be determined here
-    tx_id: U64Cell<F>,
+    tx_id: Cell<F>,
     tx_nonce: Cell<F>,
     tx_gas: Cell<F>,
     tx_gas_price: Word32Cell<F>,
@@ -78,11 +78,11 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         // Use rw_counter of the step which triggers next call as its call_id.
         let call_id = cb.curr.state.rw_counter.clone();
 
-        let tx_id = cb.query_u64();
+        let tx_id = cb.query_cell(); // already constrain `if step_first && tx_id = 1` and `tx_id += 1` at EndTx
         cb.call_context_lookup_write(
             Some(call_id.expr()),
             CallContextFieldTag::TxId,
-            tx_id.to_word(),
+            Word::from_lo_unchecked(tx_id.expr()),
         ); // rwc_delta += 1
         let mut reversion_info = cb.reversion_info_write_unchecked(None); // rwc_delta += 2
         cb.call_context_lookup_write(
@@ -534,7 +534,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         };
 
         self.tx_id
-            .assign(region, offset, Some(tx.id.to_le_bytes()))?;
+            .assign(region, offset, Value::known(F::from(tx.id as u64)))?;
         self.tx_nonce
             .assign(region, offset, Value::known(F::from(tx.nonce)))?;
         self.tx_gas
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -1,7 +1,7 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
-        param::{N_BYTES_ACCOUNT_ADDRESS, N_BYTES_GAS, N_BYTES_U64, N_BYTES_WORD},
+        param::{N_BYTES_ACCOUNT_ADDRESS, N_BYTES_GAS, N_BYTES_U64},
         step::ExecutionState,
         util::{
             and,
@@ -12,34 +12,42 @@ use crate::{
             },
             is_precompiled,
             math_gadget::{
-                ConstantDivisionGadget, ContractCreateGadget, IsEqualGadget, IsZeroGadget,
+                ConstantDivisionGadget, ContractCreateGadget, IsEqualWordGadget, IsZeroWordGadget,
                 MulWordByU64Gadget, RangeCheckGadget,
             },
-            not, or, select, CachedRegion, Cell, StepRws,
+            not, or, select, AccountAddress, CachedRegion, Cell, StepRws, U64Cell,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
     table::{
         AccountFieldTag, BlockContextFieldTag, CallContextFieldTag, TxFieldTag as TxContextFieldTag,
     },
-    util::{word::Word32Cell, Expr},
+    util::{
+        word::{Word, Word32Cell, WordCell, WordExpr},
+        Expr,
+    },
 };
-use eth_types::{evm_types::GasCost, Field, ToLittleEndian, ToScalar};
+use bus_mapping::state_db::CodeDB;
+use eth_types::{evm_types::GasCost, Field, ToWord, U256};
 use ethers_core::utils::{get_contract_address, keccak256};
-use gadgets::util::expr_from_bytes;
-use halo2_proofs::{circuit::Value, plonk::Error};
+use halo2_proofs::{
+    circuit::Value,
+    plonk::{Error, Expression},
+};
+use itertools::Itertools;
 
 #[derive(Clone, Debug)]
 pub(crate) struct BeginTxGadget<F> {
-    tx_id: Cell<F>,
+    // tx_id is query in current scope. The range should be determined here
+    tx_id: U64Cell<F>,
     tx_nonce: Cell<F>,
     tx_gas: Cell<F>,
     tx_gas_price: Word32Cell<F>,
     mul_gas_fee_by_gas: MulWordByU64Gadget<F>,
-    tx_caller_address: Cell<F>,
-    tx_caller_address_is_zero: IsZeroGadget<F>,
-    tx_callee_address: Cell<F>,
-    call_callee_address: Cell<F>,
+    tx_caller_address: WordCell<F>,
+    tx_caller_address_is_zero: IsZeroWordGadget<F, WordCell<F>>,
+    tx_callee_address: WordCell<F>,
+    call_callee_address: AccountAddress<F>,
     tx_is_create: Cell<F>,
     tx_value: Word32Cell<F>,
     tx_call_data_length: Cell<F>,
@@ -48,14 +56,14 @@ pub(crate) struct BeginTxGadget<F> {
     reversion_info: ReversionInfo<F>,
     sufficient_gas_left: RangeCheckGadget<F, N_BYTES_GAS>,
     transfer_with_gas_fee: TransferWithGasFeeGadget<F>,
-    phase2_code_hash: Cell<F>,
-    is_empty_code_hash: IsEqualGadget<F>,
-    caller_nonce_hash_bytes: [Cell<F>; N_BYTES_WORD],
+    code_hash: WordCell<F>,
+    is_empty_code_hash: IsEqualWordGadget<F, Word<Expression<F>>, Word<Expression<F>>>,
+    caller_nonce_hash_bytes: Word32Cell<F>,
     create: ContractCreateGadget<F, false>,
-    callee_not_exists: IsZeroGadget<F>,
+    callee_not_exists: IsZeroWordGadget<F, WordCell<F>>,
     is_caller_callee_equal: Cell<F>,
     // EIP-3651 (Warm COINBASE)
-    coinbase: Cell<F>,
+    coinbase: Word32Cell<F>,
     // Caller, callee and a list addresses are added to the access list before
     // coinbase, and may be duplicate.
     // <https://github.com/ethereum/go-ethereum/blob/604e215d1bb070dff98fb76aa965064c74e3633f/core/state/statedb.go#LL1119C9-L1119C9>
@@ -71,47 +79,49 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         // Use rw_counter of the step which triggers next call as its call_id.
         let call_id = cb.curr.state.rw_counter.clone();
 
-        let tx_id = cb.query_cell();
-        cb.call_context_lookup(
-            1.expr(),
+        let tx_id = cb.query_u64();
+        cb.call_context_lookup_write_unchecked(
             Some(call_id.expr()),
             CallContextFieldTag::TxId,
-            tx_id.expr(),
+            tx_id.to_word(),
         ); // rwc_delta += 1
         let mut reversion_info = cb.reversion_info_write(None); // rwc_delta += 2
-        cb.call_context_lookup(
-            1.expr(),
+        cb.call_context_lookup_write_unchecked(
             Some(call_id.expr()),
             CallContextFieldTag::IsSuccess,
-            reversion_info.is_persistent(),
+            Word::from_lo_unchecked(reversion_info.is_persistent()),
         ); // rwc_delta += 1
 
-        let [tx_nonce, tx_gas, tx_caller_address, tx_callee_address, tx_is_create, tx_call_data_length, tx_call_data_gas_cost] =
-            [
-                TxContextFieldTag::Nonce,
-                TxContextFieldTag::Gas,
-                TxContextFieldTag::CallerAddress,
-                TxContextFieldTag::CalleeAddress,
-                TxContextFieldTag::IsCreate,
-                TxContextFieldTag::CallDataLength,
-                TxContextFieldTag::CallDataGasCost,
-            ]
-            .map(|field_tag| cb.tx_context(tx_id.expr(), field_tag, None));
-        let tx_caller_address_is_zero = IsZeroGadget::construct(cb, tx_caller_address.expr());
+        let [tx_nonce, tx_gas, tx_is_create, tx_call_data_length, tx_call_data_gas_cost] = [
+            TxContextFieldTag::Nonce,
+            TxContextFieldTag::Gas,
+            TxContextFieldTag::IsCreate,
+            TxContextFieldTag::CallDataLength,
+            TxContextFieldTag::CallDataGasCost,
+        ]
+        .map(|field_tag| cb.tx_context(tx_id.expr(), field_tag, None));
+        let [tx_gas_price, tx_value] = [TxContextFieldTag::GasPrice, TxContextFieldTag::Value]
+            .map(|field_tag| cb.tx_context_as_word32(tx_id.expr(), field_tag, None));
+
+        let [tx_caller_address, tx_callee_address] = [
+            TxContextFieldTag::CallerAddress,
+            TxContextFieldTag::CalleeAddress,
+        ]
+        .map(|field_tag| cb.tx_context_as_word(tx_id.expr(), field_tag, None));
+
+        let tx_caller_address_is_zero = IsZeroWordGadget::construct(cb, &tx_caller_address);
         cb.require_equal(
             "CallerAddress != 0 (not a padding tx)",
             tx_caller_address_is_zero.expr(),
             false.expr(),
         );
-        let [tx_gas_price, tx_value] = [TxContextFieldTag::GasPrice, TxContextFieldTag::Value]
-            .map(|field_tag| cb.tx_context_as_word(tx_id.expr(), field_tag, None));
 
-        let call_callee_address = cb.query_cell();
+        let call_callee_address = cb.query_account_address();
         cb.condition(not::expr(tx_is_create.expr()), |cb| {
-            cb.require_equal(
+            cb.require_equal_word(
                 "Tx to non-zero address",
-                tx_callee_address.expr(),
-                call_callee_address.expr(),
+                tx_callee_address.to_word(),
+                call_callee_address.to_word(),
             );
         });
 
@@ -122,11 +132,11 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
 
         // Increase caller's nonce.
         // (tx caller's nonce always increases even tx ends with error)
-        cb.account_write(
-            tx_caller_address.expr(),
+        cb.account_write_word(
+            tx_caller_address.to_word(),
             AccountFieldTag::Nonce,
-            tx_nonce.expr() + 1.expr(),
-            tx_nonce.expr(),
+            Word::from_lo_unchecked(tx_nonce.expr() + 1.expr()),
+            Word::from_lo_unchecked(tx_nonce.expr()),
             None,
         ); // rwc_delta += 1
 
@@ -134,7 +144,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         // transaction format)
         // Calculate transaction gas fee
         let mul_gas_fee_by_gas =
-            MulWordByU64Gadget::construct(cb, tx_gas_price.clone().into(), tx_gas.expr());
+            MulWordByU64Gadget::construct(cb, tx_gas_price.clone(), tx_gas.expr());
 
         let tx_call_data_word_length =
             ConstantDivisionGadget::construct(cb, tx_call_data_length.expr() + 31.expr(), 32);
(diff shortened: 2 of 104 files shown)


hero78119:
related to https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1415

hero78119:
Update: [fix caller/callee/coinbase cell to word type](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/ec8f365be75c21d9f373c333988bea7a4e43dcec)

hero78119:
Update: fix `log` opcode topic type to word https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/43c4d9f73bbaee02d17ebe613afeec6619fb8116

hero78119:
Updated
- [x] remove all compatible workaround `construct_new` under evm circuit
- [x] unittest under evm circuit all pass `cargo test --features warn-unimplemented --features test --package zkevm-circuits --lib -- evm_circuit::execution::`

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -118,10 +118,15 @@ impl<F: Field> RestoreContextGadget<F> {
     ) -> Self {
         // Read caller's context for restore
         let caller_id = cb.call_context(None, CallContextFieldTag::CallerId);
-        let [caller_is_root, caller_is_create, caller_program_counter, caller_stack_pointer, caller_gas_left, caller_memory_word_size, caller_reversible_write_counter] =
+        let [caller_is_root, caller_is_create] =
+            [CallContextFieldTag::IsRoot, CallContextFieldTag::IsCreate]
+                .map(|field_tag| cb.call_context(Some(caller_id.expr()), field_tag));
+
+        let caller_code_hash =
```
hero78119, line 125: here separate word lookup into `call_context_read_as_word` but still need to keep lookup order correct.

hero78119:
Updated: rebase onto latest word-lo-hi branch change. Unittest under evm_circuit pass if masked state circuit in test_utils. (it failed for now since touch unimplemented error in state circuit). Ready for review

hero78119:
updated: fix copy circuit lookup order.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -521,33 +559,31 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             tx.value,
             gas_fee,
         )?;
-        self.phase2_code_hash
-            .assign(region, offset, region.word_rlc(callee_code_hash))?;
-        self.is_empty_code_hash.assign_value(
+        // TODO handle endian issue
+        self.code_hash
+            .assign_u256(region, offset, callee_code_hash)?; // big endian assignment
```
ChihChengLiang, line 564: Is this TODO resolved?

zkevm-circuits/src/witness/block.rs
```diff
@@ -239,7 +243,8 @@ pub fn block_convert<F: Field>(
     code_db: &bus_mapping::state_db::CodeDB,
 ) -> Result<Block<F>, Error> {
     let rws = RwMap::from(&block.container);
-    rws.check_value();
+    // TODO add me back
+    // rws.check_value();
```
ChihChengLiang, line 247:
```suggestion
    rws.check_value();
```

zkevm-circuits/src/witness/bytecode.rs
```diff
@@ -24,16 +27,14 @@ impl Bytecode {
     /// Assignments for bytecode table
     pub fn table_assignments<F: Field>(
         &self,
-        challenges: &Challenges<Value<F>>,
-    ) -> Vec<[Value<F>; 5]> {
+        _challenges: &Challenges<Value<F>>,
```
ChihChengLiang, line 30: This table_assignment is not a trait method. I think the challenge can be safely removed.
```suggestion
```

zkevm-circuits/src/witness/tx.rs
```diff
@@ -42,66 +43,71 @@ impl Transaction {
     /// calldata) and tx_calldata
     pub fn table_assignments<F: Field>(
         &self,
-        challenges: Challenges<Value<F>>,
-    ) -> [Vec<[Value<F>; 4]>; 2] {
+        _challenges: Challenges<Value<F>>,
```
ChihChengLiang, line 46: This challenge can go too.
```suggestion
```

zkevm-circuits/src/witness/block.rs
```diff
@@ -156,46 +156,50 @@ pub struct BlockContext {
 
 impl BlockContext {
     /// Assignments for block table
-    pub fn table_assignments<F: Field>(&self, randomness: Value<F>) -> Vec<[Value<F>; 3]> {
+    pub fn table_assignments<F: Field>(&self, _randomness: Value<F>) -> Vec<[Value<F>; 4]> {
```
ChihChengLiang, line 159: The randomness can go
```suggestion
    pub fn table_assignments<F: Field>(&self) -> Vec<[Value<F>; 4]> {
```

ChihChengLiang (commented):
First pass.

zkevm-circuits/src/witness/block.rs
hero78119: Have clean up the _randomness var in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/0646b790010306d9df41cd27002bf9890d017202

zkevm-circuits/src/witness/tx.rs
hero78119: Have clean up the _randomness var in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/commit/0646b790010306d9df41cd27002bf9890d017202

zkevm-circuits/src/witness/block.rs
hero78119: Thanks for catch up this, retain it back in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/commit/0646b790010306d9df41cd27002bf9890d017202

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
hero78119: Here the type is U256, but it's big endian. The reason is because we implement `ToWord` for `Hash` without align to little endian so cause confusing. I try to fixed in  https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/5f0bdfc85f79ef835cafded98df8d76005251822 and unittest still pass (just verify test under evm circuit) Do you think above fixed make sense~?

ChihChengLiang: Fixing in another PR makes sense. Can we create an issue to track this and add the issue number in the TODO note?

hero78119: issue raised, https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1467 and rollback fixed in this pr to limit the scope

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -97,31 +100,31 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
...
         let is_warm_prev = cb.query_bool();
-        cb.account_access_list_write(
+        cb.account_access_list_write_word(
             tx_id.expr(),
-            call_gadget.callee_address_expr(),
-            is_warm.expr(),
-            is_warm_prev.expr(),
+            call_gadget.callee_address_word(),
+            Word::from_lo_unchecked(is_warm.expr()),
+            Word::from_lo_unchecked(is_warm_prev.expr()),
```
ChihChengLiang, line 127: We can move all the `Word::from_lo_unchecked` inside the function body of `account_access_list_write_word`. 

- `is_warm` is intended to be a boolean cell.
- The conversion of `is_warm` to word is required by the RW table. 

So I think it makes sense for the circuits to be agnostic of the conversion. `account_access_list_write_word`  communicate with RW table directly and it should handle the conversion.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -62,47 +70,49 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
+        ]
+        .map(|field_tag| cb.tx_context(tx_id.expr(), field_tag, None));
+        let [tx_gas_price, tx_value] = [TxContextFieldTag::GasPrice, TxContextFieldTag::Value]
+            .map(|field_tag| cb.tx_context_as_word32(tx_id.expr(), field_tag, None));
+
+        let [tx_caller_address, tx_callee_address] = [
+            TxContextFieldTag::CallerAddress,
+            TxContextFieldTag::CalleeAddress,
+        ]
+        .map(|field_tag| cb.tx_context_as_word(tx_id.expr(), field_tag, None));
```
ed255, line 101: ALERT: this is changing the order of the TxContext rw operations, which affects the rwc associated to them.  If you change the order here you must also change it in the bus-mapping, otherwise we will encounter lookup failures when testing that are hard to debug.  My advice is to just keep the same order of rw operations (even if the code looks a bit more verbose)

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -62,47 +70,49 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
-        let call_callee_address = cb.query_cell();
+        let call_callee_address = cb.query_account_address();
         cb.condition(not::expr(tx_is_create.expr()), |cb| {
-            cb.require_equal(
+            cb.require_equal_word(
                 "Tx to non-zero address",
-                tx_callee_address.expr(),
-                call_callee_address.expr(),
+                tx_callee_address.to_word(),
+                call_callee_address.to_word(),
```
ed255, line 115: Maybe we can replace all instances of `require_equal_word` where the two values are address to `require_equal(address_word_to_expr(a), b.expr())` or something similar.

Also, I notice that `address_word_to_expr` could be a method on Word maybe?  And possibly add an `unsafe` prefix, because it assumes that the word represents an address, but is unsafe if it doesn't.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -451,46 +491,44 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         };
 
         self.tx_id
-            .assign(region, offset, Value::known(F::from(tx.id as u64)))?;
+            .assign(region, offset, Some(tx.id.to_le_bytes()))?;
```
ed255, line 494: I think there are quite a few of these assign calls of the `RandomLinearCombination` that take an array of bytes, but I think many times we have integers instead of byte arrays.  Maybe we can add a method like `assign_h160`, `assign_usize` to `RandomLinearCombination` to avoid calling `foo.to_le_bytes()` all the time.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -451,46 +491,44 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
+            } else {
+                tx.callee_address
+            }
+            .to_fixed_bytes()
+            .iter()
+            .rev()
+            .copied()
+            .collect_vec()
+            .try_into()
+            .ok(),
```
ed255, line 526: wow!  Hopefully this can be improvement following my previous comment and having `.assign_h160`

ed255 (commented):
First pass review.  I've just reviewed 9 files.  Please take a look at my comments and discuss: some of them contain suggestions that would apply to many files :sweat_smile:

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -40,31 +41,35 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
...
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let address = cb.query_word_rlc();
+        let address = cb.query_account_address();
         cb.stack_pop_word(address.to_word());
 
         let tx_id = cb.call_context(None, CallContextFieldTag::TxId);
         let mut reversion_info = cb.reversion_info_read(None);
         let is_warm = cb.query_bool();
-        cb.account_access_list_write(
+        cb.account_access_list_write_word(
```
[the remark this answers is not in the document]
hero78119: There is a similar discussion in another pr https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1441#discussion_r1210535805 and finally reach consensus that we should take use `address` as word across whole codebase to align the knowledge.  That's why I deprecated this api from `word` in this pr. @ed255 appreciate your opinion in more detail if possible. To me it's just coding style, so im ok for both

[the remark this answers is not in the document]
ed255: I see, I was not aware of that discussion!  I understand CC's reasoning.  I'm not yet 100% convinced about it but I think this is a small detail and I should not get stuck in it!  So it sounds good to me to go his way :)

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
hero78119: unlike rwtable, TxContext there is no rw counter, so the order doesn't matter by design. TxContext is only filled once via Pi Circuit, for lookup part it should be read only. @ed255 if the order doens't matter, do you think we still need to keep the order ~?

hero78119:
Update: rebase to adopt latest change on main branch.

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -40,31 +41,35 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
...
             tx_id.expr(),
-            address.expr(),
-            1.expr(),
-            is_warm.expr(),
+            address.to_word(),
+            Word::from_lo_unchecked(1.expr()),
+            Word::from_lo_unchecked(is_warm.expr()),
```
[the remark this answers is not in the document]
ed255: I realized this is a duplicate of https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411#discussion_r1226830265

zkevm-circuits/src/evm_circuit/execution/balance.rs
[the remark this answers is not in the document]
ed255: I just created an issue to discuss this in the future https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1479
We can revisit the discussion after the word-lo-hi refactor is merged in main.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -79,30 +75,33 @@ pub(crate) mod address_high {
...
-    ) -> Self {
-        todo!()
-    }
-
-    pub(crate) fn construct_new(
         cb: &mut EVMConstraintBuilder<F>,
-        memory_offset: Word32Cell<F>,
-        memory_length: Word32Cell<F>,
+        memory_offset_word: WordCell<F>,
+        memory_length: MemoryAddress<F>,
```
ed255, line 88: I wonder if `memory_length` should be `Word32` for these two reasons:
- Usually these arguments come from the stack, and the user may put anything there.  Having a length too big will case an error but we're actually using these gadgets in Error states, so we need to be able to read the stack value and prove it's too big.  Perhaps we can get away with it:  The ExecStep could allocate Word32, and from it build a MemoryAddress (by taking the less significant 5 bytes).

After thinking a bit more about this: maybe it's OK to use `MemoryAddress` here:
- Allocating MemoryAddress for length and reading it from stack in an opcode ExecStep should be OK, because we only prove that when the step succeeds, so it should fit.
- For the ExecStates that do OOG due to length being more than 5 bytes, we can't use `MemoryAddress` for the stack read of length.

zkevm-circuits/src/evm_circuit/execution/codesize.rs
```diff
@@ -35,19 +34,19 @@ impl<F: Field> ExecutionGadget<F> for CodesizeGadget<F> {
...
 
         let code_hash = cb.curr.state.code_hash.clone();
         let codesize = cb.query_cell();
-        cb.bytecode_length(code_hash.expr(), codesize.expr());
+        cb.bytecode_length_word(code_hash.to_word(), codesize.expr());
 
         cb.require_equal(
             "Constraint: bytecode length lookup == codesize",
-            from_bytes::expr(&codesize_bytes),
+            codesize_bytes.to_word().lo(),
```
ed255, line 45: Can we do `codesize_bytes.expression` or `codesize_bytes.expr()` here?

ed255 (commented):
Second pass review.  I still haven't reviewed all the files, there are many changes :sweat_smile:

ed255:
This is a big PR and while reviewing I think I can categorize my comments into:
- a. Something that I think is wrong
- b. Something that I think should be improved at API level / code style
- c. Something that I think should be improved for optimization (cell usage or constraints)
- d. Improvements that I see that are unrelated to this refactor.  I try not to focus on these to avoid noise!

I think the most important one to resolve is `a`.  If it's an over-constraint, we should be able to detect them also via the unit tests.  If it's an under constraint, that's more difficult to detect other than via review.

For `b` and `c` I think we could defer to resolve some of the discussions to a future PR if @hero78119 considers it (on a case by case basis).  This way we can get this PR merged sooner, and work on `b` and `c` after the unit tests are passing, so that we can apply changes more confidently.

ChihChengLiang:
@ed255 Good point. I'll focus on finding issues in `a` in the review.

hero78119:
> This is a big PR and while reviewing I think I can categorize my comments into:
> 
> * a. Something that I think is wrong
> * b. Something that I think should be improved at API level / code style
> * c. Something that I think should be improved for optimization (cell usage or constraints)
> * d. Improvements that I see that are unrelated to this refactor.  I try not to focus on these to avoid noise!
> 
> I think the most important one to resolve is `a`. If it's an over-constraint, we should be able to detect them also via the unit tests. If it's an under constraint, that's more difficult to detect other than via review.
> 
> For `b` and `c` I think we could defer to resolve some of the discussions to a future PR if @hero78119 considers it (on a case by case basis). This way we can get this PR merged sooner, and work on `b` and `c` after the unit tests are passing, so that we can apply changes more confidently.

It make senses! since evm circuit touch many change, I think first huge pr we can focus `a` more. Besides I will try to fix few related to b, c, d as well. My principle for this pr is all evm circuit unittest test should pass (Right now it's already in this state ✅  and so as later change should keep all uniitest pass as well!)

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -1,36 +1,37 @@
...
 };
-use eth_types::{evm_types::GasCost, Field, ToLittleEndian};
+use eth_types::{evm_types::GasCost, Field, ToLittleEndian, ToWord};
 use halo2_proofs::{circuit::Value, plonk::Error};
 
 #[derive(Clone, Debug)]
 pub(crate) struct BalanceGadget<F> {
     same_context: SameContextGadget<F>,
-    address: Word<F>,
+    address: AccountAddress<F>,
```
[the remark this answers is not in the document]
hero78119: Pretty agree with this optimisation to have clear definition, and `ByteDecomposition` is a good naming with hardcode 256 base. 

I am thinking a design problem. For `wordXX`, (XX means number of limbs) cell, the calling of assign_<type> make sense, because we have loosely requirements: `number of bytes` % `number of limbs`. For `ByteDecomposition`, `number of bytes` need to strictly equals with `number of limbs`. Therefore, it will be nice to have a restriction usage by type system itself. What I means is

```
- ByteDecomposition<F, 8> only have function assign(value: u64)
- ByteDecomposition <F, 20> only have function assign(value: H160)
- ... (so as other type)
```

strictly one-to-one mapping. It's to avoid some mal-usage like `ByteDecomposition<F, 8>.assign_h160(value: H160)` because this pair really not making any sense, so intuitively maybe here shouldn't apply generic, instead it should leverage `macro` system. 
@ed255 do you think it's an over-design or is there any other simple way to leverage type system to make it works? :)

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -71,47 +79,49 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
-        let tx_caller_address_is_zero = IsZeroGadget::construct(cb, tx_caller_address.expr());
+        let [tx_nonce, tx_gas, tx_is_create, tx_call_data_length, tx_call_data_gas_cost] = [
+            TxContextFieldTag::Nonce,
+            TxContextFieldTag::Gas,
+            TxContextFieldTag::IsCreate,
+            TxContextFieldTag::CallDataLength,
+            TxContextFieldTag::CallDataGasCost,
+        ]
+        .map(|field_tag| cb.tx_context(tx_id.expr(), field_tag, None));
+        let [tx_gas_price, tx_value] = [TxContextFieldTag::GasPrice, TxContextFieldTag::Value]
```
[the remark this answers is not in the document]
hero78119: Here is the comparison for main branch/evm word-lo-hi branch utilization

### Storage_1
```
Main:
+-----------------------------------+------------------------------+-------------------------+
| "storage_1" total_available_cells | "storage_1" total_used_cells | "storage_1" Utilization |
+-----------------------------------+------------------------------+-------------------------+
| 25480                             | 6482                         | 25.4                    |
+-----------------------------------+------------------------------+-------------------------+

+-----------------------------------+------------------------------+-------------------------+
| "storage_1" total_available_cells | "storage_1" total_used_cells | "storage_1" Utilization |
+-----------------------------------+------------------------------+-------------------------+
| 24220                             | 7273                         | 30.0                    |
+-----------------------------------+------------------------------+-------------------------+
```

### Storage_2
```
Main
+-----------------------------------+------------------------------+-------------------------+
| "storage_2" total_available_cells | "storage_2" total_used_cells | "storage_2" Utilization |
+-----------------------------------+------------------------------+-------------------------+
| 1456                              | 467                          | 32.1                    |
+-----------------------------------+------------------------------+-------------------------+

+-----------------------------------+------------------------------+-------------------------+
| "storage_2" total_available_cells | "storage_2" total_used_cells | "storage_2" Utilization |
+-----------------------------------+------------------------------+-------------------------+
| 1384                              | 14                           | 1.0                     |
+-----------------------------------+------------------------------+-------------------------+
```

### byte_lookup
```
Main 
+-------------------------------------+--------------------------------+---------------------------+
| "byte_lookup" total_available_cells | "byte_lookup" total_used_cells | "byte_lookup" Utilization |
+-------------------------------------+--------------------------------+---------------------------+
| 8736                                | 6786                           | 77.7                      |
+-------------------------------------+--------------------------------+---------------------------+

+-------------------------------------+--------------------------------+---------------------------+
| "byte_lookup" total_available_cells | "byte_lookup" total_used_cells | "byte_lookup" Utilization |
+-------------------------------------+--------------------------------+---------------------------+
| 8304                                | 6606                           | 79.6                      |
+-------------------------------------+--------------------------------+---------------------------+

```

Overall before/after utilization look not significant diff, some even decrease. After u16 optimisation it will be optimise more.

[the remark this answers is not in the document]
hero78119: > TransferWithGasFeeGadget

E.g. `MulWordByU64Gadget` there is problematics design which make the refactor harder. The thing is, `multiplicand` is passed from caller, however assign inside `MulWordByU64Gadget`. So, `multiplicand` can not took generics type, otherwise it make `assign` harder, because for now we need to know exact type when we doing assigment. 

I would suggest to refactor cell usage gradually maybe later as an optimisation. For now we took functionality correct first :P

zkevm-circuits/src/evm_circuit/execution/block_ctx.rs
```diff
@@ -6,29 +6,32 @@ use crate::{
...
 use bus_mapping::evm::OpcodeId;
-use eth_types::{Field, ToLittleEndian};
+use eth_types::Field;
 use halo2_proofs::plonk::Error;
 
 #[derive(Clone, Debug)]
 pub(crate) struct BlockCtxGadget<F, const N_BYTES: usize> {
     same_context: SameContextGadget<F>,
-    value: RandomLinearCombination<F, N_BYTES>,
+    value: WordCell<F>,
```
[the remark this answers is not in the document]
hero78119: Nice catch and done in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/7b1dc943f3303efc6eed0c497e86f3141f9f28c3

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1420,8 +1190,7 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
                 Word::zero(),
                 value,
                 Word::zero(),
-                0.expr(),
-                0.expr(),
+                Word::zero(),
```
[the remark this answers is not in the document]
hero78119: After the reminding, I applied a principle in latest commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/498c5025af7471d86dedb9cec5d99a49b7a1bf6c

> Principle: an api in constraint builder will have suffix `_unchecked`, If a api interface take `Expression<F>` as function argument and internally convert to word by `Word::from_lo_unchecked`

zkevm-circuits/src/evm_circuit/execution/balance.rs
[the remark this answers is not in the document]
hero78119: > Making sure we don't call the assignment function with a value bigger than the type can hold sounds great! Maybe we can skip the `macros` and just add some asserts like:
> 
> ```rust
> fn assign_h160(...) {
>     assert!(N >= 20);
> }
> ```
> 
> We would get a runtime panic when the assign is used incorrectly.

(Trying the rephrase in case not get my points)

I think in this case we need to be exactly equal, right? `assert!(N == 20)` and if it's true, actually assign_h160 is a bit verbose in `ByteDecomposition` instead of just `assign` because

1. assign(word: H160) imply ByteDecomposition <F, 20>
2. assign(word: U64) imply ByteDecomposition <F, 8>

Thats why I think it will be nice to have compiler time check and catch error like `ByteDecomposition <F, 20>.assign(u64)` But seems there is no better way to do that

zkevm-circuits/src/evm_circuit/execution/balance.rs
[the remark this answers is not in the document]
hero78119: It's done in latest commit

ChihChengLiang (approved):
LGTM after addressing all issues.

zkevm-circuits/src/evm_circuit/execution/exp.rs
```diff
@@ -92,12 +86,7 @@ impl<F: Field> ExecutionGadget<F> for ExponentiationGadget<F> {
         let exponent_is_one_expr =
             and::expr([exponent_lo_is_one.expr(), exponent_hi_is_zero.expr()]);
 
-        let zero_rlc = cb.query_word_rlc();
-        cb.require_zero(
-            "base * base + c == base^2 (c == 0)",
```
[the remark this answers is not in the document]
hero78119: I think it's some left-over since very early stage.

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -323,54 +330,80 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
...
-                        Some(callee_call_id.expr()),
-                        field_tag,
-                        value,
-                    );
+                    cb.call_context_lookup_write(Some(callee_call_id.expr()), field_tag, value);
+                }
+
+                for (field_tag, value) in [] {
+                    cb.call_context_lookup_write(Some(callee_call_id.expr()), field_tag, value);
                 }
```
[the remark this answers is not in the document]
hero78119: it's some left over, have clean up it in latest commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/6bc88fe5762a0ab3e7c8f7462a6e19f9fa8f7572

zkevm-circuits/src/evm_circuit/execution/error_invalid_jump.rs
```diff
@@ -26,8 +29,8 @@ pub(crate) struct ErrorInvalidJumpGadget<F> {
     is_code: Cell<F>,
     is_jump_dest: IsEqualGadget<F>,
     is_jumpi: IsEqualGadget<F>,
-    phase2_condition: Cell<F>,
-    is_condition_zero: IsZeroGadget<F>,
+    condition: Word32Cell<F>,
```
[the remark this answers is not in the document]
hero78119: Have updated in latest commit! thanks for catch up this
I also checked jumpi.rs and already used optimisation version

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1473,29 +1241,14 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
         self.reversion_info(call_id, false)
     }
 
-    pub(crate) fn reversion_info_write(
+    pub(crate) fn reversion_info_write_unchecked(
```
hero78119, line 1244: Discussion:  I noticed in [begin_tx.rs](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs#L81) and [ReversionInfo](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/util/constraint_builder.rs#L96-L104) main branch there is no range check, does it safe?

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -752,39 +729,37 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
...
-        let word = self.query_word32();
+    ) -> WordCell<F> {
+        let word = self.query_word_unchecked();
         self.tx_context_lookup_word(id, field_tag, index, word.to_word());
         word
     }
 
-    #[deprecated(note = "tx_context_lookup_word is favored")]
-    pub(crate) fn tx_context_lookup(
+    pub(crate) fn tx_context_as_account_address(
```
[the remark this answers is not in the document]
hero78119: Removed this method in latest commit, since it doesn't used anymore

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -114,8 +119,15 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
...
-            .assign(region, offset, Some(address.to_le_bytes()))?;
+        self.address.assign(
+            region,
+            offset,
+            Some(
+                address.to_le_bytes()[0..N_BYTES_ACCOUNT_ADDRESS]
+                    .try_into()
+                    .unwrap(),
+            ),
+        )?;
```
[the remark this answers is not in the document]
hero78119: Adding intDecomposition type and support assign_h160 https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/87370f81b0f15c6809fe0a14e908b6c0aa2cdb78

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
hero78119: Added intDecomposition type https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/87370f81b0f15c6809fe0a14e908b6c0aa2cdb78

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -71,47 +79,49 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         // Use rw_counter of the step which triggers next call as its call_id.
         let call_id = cb.curr.state.rw_counter.clone();
 
-        let tx_id = cb.query_cell();
-        cb.call_context_lookup(
-            1.expr(),
+        let tx_id = cb.query_u64();
```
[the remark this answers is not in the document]
hero78119: Applied this optimisation in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/43b03c83687e7380a19668d0174cb0307f9fd043

zkevm-circuits/src/evm_circuit/execution/balance.rs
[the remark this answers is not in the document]
ed255: > > Making sure we don't call the assignment function with a value bigger than the type can hold sounds great! Maybe we can skip the `macros` and just add some asserts like:
> > ```rust
> > fn assign_h160(...) {
> >     assert!(N >= 20);
> > }
> > ```
> > We would get a runtime panic when the assign is used incorrectly.
> 
> (Trying the rephrase in case not get my points)
> 
> I think in this case we need to be exactly equal, right? `assert!(N == 20)` and if it's true, actually assign_h160 is a bit verbose in `ByteDecomposition` instead of just `assign` because
> 
>     1. assign(word: H160) imply ByteDecomposition <F, 20>
>     2. assign(word: U64) imply ByteDecomposition <F, 8>
> 
> 
> Thats why I think it will be nice to have compiler time check and catch error like `ByteDecomposition <F, 20>.assign(u64)` But seems there is no better way to do that

Right, I think checking equality sounds better!

I agree that it looks redundant, because the N tells you already which is the compatible type, but you still need to write it explicitly.  Nevertheless I think there's no nice solution in Rust to implement the same function with different types.  I know you could do traits like `AssignU64`, `AssignH160` with `assign(word: H160)` and `assign(word: U64)` methods; but then we need to import the traits (which is a bit annoying) and if we import multiple types we need to do things like:
`<value as AssignU64>.assign(u64)` which are a bit ugly.

So implementing `assign_h160` and `assign_u64` for `ByteDecomposition <F, N>` and then checking the correct `N` with assert at runtime is the nicest thing I can imagine.

zkevm-circuits/src/evm_circuit/execution/balance.rs
```diff
@@ -114,8 +118,7 @@ impl<F: Field> ExecutionGadget<F> for BalanceGadget<F> {
         self.same_context.assign_exec_step(region, offset, step)?;
 
         let address = block.get_rws(step, 0).stack_value();
-        self.address_word
-            .assign(region, offset, Some(address.to_le_bytes()))?;
+        self.address.assign_u256(region, offset, address)?;
```
ed255, line 121: Shouldn't this be `assign_h160(..., address.to_address())`?

zkevm-circuits/src/util/int_decomposition.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /// assign bytes to cells
+    pub fn assign<const N_LIMBS_ASSIGN: usize>(
+        &self,
+        region: &mut CachedRegion<'_, '_, F>,
+        offset: usize,
+        bytes: Option<[u8; N_LIMBS_ASSIGN]>,
+    ) -> Result<Vec<AssignedCell<F, F>>, Error> {
+        assert!(N_LIMBS_ASSIGN >= N_LIMBS);
```
ed255, line 37: Shouldn't this be `N_LIMBS >= N_LIMBS_ASSIGN`?
Otherwise there are more bytes to assign than limbs, so the more significant bytes are ignored.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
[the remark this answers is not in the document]
ed255: Sounds good, we can leave this for a future optimization :)  Thanks for the utilization results!  They are very useful because they show that the cell usage hasn't grown!

zkevm-circuits/src/evm_circuit/execution/pc.rs
```diff
@@ -31,17 +30,17 @@ impl<F: Field> ExecutionGadget<F> for PcGadget<F> {
...
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let value = cb.query_word_rlc();
+        let value = cb.query_u64();
 
         // program_counter is limited to 64 bits so we only consider 8 bytes
         cb.require_equal(
             "Constrain program_counter equal to stack value",
-            from_bytes::expr(&value.cells),
+            value.to_word().lo(),
```
ed255, line 38:
```suggestion
            value.expr(),
```
I think it's more clear like this.

zkevm-circuits/src/util/int_decomposition.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// assign all limbs into one expression
+    pub fn sum_expr(&self) -> Expression<F> {
+        sum::expr(self.limbs.clone())
+    }
+}
+
+impl<F: Field, const N_LIMBS: usize> Expr<F> for IntDecomposition<F, N_LIMBS> {
+    fn expr(&self) -> Expression<F> {
+        rlc::expr(&self.limbs.clone().map(|limb| limb.expr()), 256.expr())
```
ed255, line 79: I would suggest doing `assert!(N_LIMBS <= 31)` as a safety check here.  This is because for N_LIMBS > 31 we overflow the last byte in the field.

zkevm-circuits/src/evm_circuit/execution/returndatasize.rs
```diff
@@ -70,14 +72,12 @@ impl<F: Field> ExecutionGadget<F> for ReturnDataSizeGadget<F> {
...
             region,
             offset,
-            Some(
-                return_data_size.to_le_bytes()[..N_BYTES_U64]
-                    .try_into()
-                    .expect("could not encode return_data_size as byte array in little endian"),
-            ),
+            Value::known(from_bytes::value(
+                &return_data_size.to_le_bytes()[..N_BYTES_U64],
+            )),
```
ed255, line 80: Suggestion:
```rust
self.return_data_size.assign_u64(region, offset, return_data_size.to_u64());
```

The general rule I follow for this suggestion is to:
- Avoid calling `to_le_bytes()` manually because it adds verbosity
- If we want to assign a Word to a smaller signal, instead of discarding bytes, doing a conversion like `to_u64()` which panics when the value is too big.

zkevm-circuits/src/util/word.rs
```diff
@@ -44,6 +44,10 @@ impl<T, const N: usize> WordLimbs<T, N> {
     pub fn n() -> usize {
         N
     }
+    /// get limbs by index
+    pub fn get(&self, index: usize) -> &T {
```
ed255, line 48: When I saw this being called I didn't know what it was by the function name.  Maybe calling this `limb` would make it more clear?

Or maybe remove this method and just use `word.limbs[index]`? Considering we have `pub limbs`

zkevm-circuits/src/evm_circuit/execution/shl_shr.rs
```diff
@@ -59,38 +60,50 @@ impl<F: Field> ExecutionGadget<F> for ShlShrGadget<F> {
...
-        cb.stack_pop(is_shl.expr() * quotient.expr() + is_shr.expr() * dividend.expr());
-        cb.stack_push(
-            (is_shl.expr() * dividend.expr() + is_shr.expr() * quotient.expr())
-                * (1.expr() - divisor_is_zero.expr()),
+        cb.stack_pop_word(shift.to_word());
+        cb.stack_pop_word(
+            quotient
+                .to_word()
+                .mul_selector(is_shl.expr())
+                .add_unchecked(dividend.to_word().mul_selector(is_shr.expr())),
```
ed255, line 86: I think we could find a better way to write these kind of expressions (I also saw them in another opcode).

So originally we had:
```rust
v = is_shl.expr() * dividend.expr() + is_shr.expr() * quotient.expr())
                * (1.expr() - divisor_is_zero.expr()
```

The logic for this is the following:
```rust
v = if (not (divisor_is_zero) {
  if (is_shl) { 
    dividend 
  } else if (is_shr) {
    quotient 
  } else {
    0
  }
} else {
  0
};
```

So maybe we could have some helpers to write these if/else where the conditions are booleans and the values are words.  But let's leave this for a future PR!  I just wanted to share my thoughts here :)

zkevm-circuits/src/evm_circuit/execution/balance.rs
hero78119: `assign_u256` is correct because here address already in u256 little endian 20 bytes

zkevm-circuits/src/util/int_decomposition.rs
hero78119: More significant bytes ignored is expected because we have so many place to put smaller value in u256 so here we allow larger bytes to smaller limbs, which also make caller side easiler

zkevm-circuits/src/util/word.rs
hero78119: I roughly remember, this is to avoid clone of whole limbs if we only want to fetch single limb refereence. If we took `word.limbs[index]` it compiler might treat it as move, so it make us to clone something. But I think right now there is no space to use `word.limbs[index]`

Removed this api in commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits

zkevm-circuits/src/evm_circuit/execution/returndatasize.rs
hero78119: Yes, `assign_lo` is an old api, and after few api iteration, now we have assign_u64. 

I removed assign_lo api in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits, along with some places to switch to proper `assign_u64`

zkevm-circuits/src/util/int_decomposition.rs
hero78119: Added in this commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits

zkevm-circuits/src/evm_circuit/execution/pc.rs
hero78119: Updated in this commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits. And I also check all place which have `value.to_word().lo()` and fixed to `value.expr()` after having `IntDecomposition` type

zkevm-circuits/src/evm_circuit/execution/shl_shr.rs
hero78119: Yes, agree current unchecked style looks very ugly and hard to read 🥲 
It will be good to left it as a future PR and check whether can leverage `macro!` features to make it looks confortable!

hero78119:
> This is a big PR and while reviewing I think I can categorize my comments into:
> 
> * a. Something that I think is wrong
> * b. Something that I think should be improved at API level / code style
> * c. Something that I think should be improved for optimization (cell usage or constraints)
> * d. Improvements that I see that are unrelated to this refactor.  I try not to focus on these to avoid noise!
> 
> I think the most important one to resolve is `a`. If it's an over-constraint, we should be able to detect them also via the unit tests. If it's an under constraint, that's more difficult to detect other than via review.
> 
> For `b` and `c` I think we could defer to resolve some of the discussions to a future PR if @hero78119 considers it (on a case by case basis). This way we can get this PR merged sooner, and work on `b` and `c` after the unit tests are passing, so that we can apply changes more confidently.

@ed255 I create another issue https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1487 for recording future optimisation and housekeeping related to word-lo-hi, if you thought any related to `b, c, d` and can be addressed in the future :)

zkevm-circuits/src/util/int_decomposition.rs
ed255: Hm, I see.  I think it would be better if we only assign types that can fit (and if we have an address in u256, have the caller convert it to address before assigning).  But we can leave this discussion for the future.

Nevertheless, to follow this current implementation I would ask for the following safety check:
for each byte that is not assigned, assert it is 0.
Something like this:
```rust
for byte in bytes[N_LIMBS_ASSIGN-N_LIMBS..] {
    assert_eq(byte, 0);
}
```

zkevm-circuits/src/evm_circuit/execution/stop.rs
```diff
@@ -54,7 +57,11 @@ impl<F: Field> ExecutionGadget<F> for StopGadget<F> {
         );
 
         // Call ends with STOP must be successful
-        cb.call_context_lookup(false.expr(), None, CallContextFieldTag::IsSuccess, 1.expr());
+        cb.call_context_lookup_read(
+            None,
+            CallContextFieldTag::IsSuccess,
+            Word::from_lo_unchecked(1.expr()),
```
ed255, line 63: Idea: implement `Word::one()` and `Word::zero()`. I found 5 instances of this pattern.  Maybe it's not enough instances to justify a new method :thinking:

zkevm-circuits/src/util/word.rs
```diff
@@ -408,12 +414,9 @@ impl<F: Field> Word<Expression<F>> {
...
     }
 
-    /// Compress the lo and hi limbs into an expression without checking the overflow.
-    /// So far only use it for address.
-    /// TODO We should remove it before merging to the main branch.
-    #[deprecated(note = "no overflow check and unsafe. please consider keep word type")]
-    pub fn expr_unchecked(&self) -> Expression<F> {
-        self.lo() + self.hi() * (1 << (N_BYTES_HALF_WORD * 8)).expr()
+    /// No overflow check on lo/hi limbs
+    pub fn mul_unchecked(self, rhs: Self) -> Self {
```
ed255, line 418: For a future PR I think we should try to remove `add_unchecked`, `sub_unchecked` and `mul_unchecked` and rewrite the gadgets that use these functionalities in a more clear way.

ed255 (approved):
I've finished reviewing all the files, phew!

There are still comments from me marked as unresolved but I think we can discuss them / figure them out in a future PR.  The only comment that I would like to really discuss before merging the PR is this one https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411#discussion_r1235123444

Aside from that everything looks good to me!  This was a massive work!  Awesome job @hero78119 !!! :confetti_ball:

I'm approving the PR already so that you can merge when you see fit!

zkevm-circuits/src/util/int_decomposition.rs
hero78119: Really nice suggestion! Added check in this commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1411/commits/f245af8248fa073011e0c0d7675c00a206b4da53

zkevm-circuits/src/evm_circuit/execution/stop.rs
hero78119: These 2 method already there 😅 ~~ I just replace all of them so it's consistent now

hero78119:
Appreciate @ChihChengLiang and @ed255 great effort on reviewing 🔥 👏 🙏 Will merge it sooner :)

PR #1414, opened 2023-05-17
[word lo/hi] retype table, common gadget and util design

hero78119:
### Description

Few fix to iterating utilities api.

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1388

### Type of change

### Contents

- more documentations
- word limbs add `assign_lo_hi`, while lo, hi part support assign with different length of bytes respectively.  Refer the use case in `evm_circuit/execution/address.rs`, `evm_circuit/execution/caller.rs`.
- `rwtable.aux2` change to `Word` type to fit committed value
- `lt_word` gadget usage fix
- fix `MemoryAddressGadget` to fit EVM spec, for length = 0 then `offset` can be arbitrary value still legal (even larger than 5 bytes, since it will be ignore).
- fix `SstoreGasGadget` unnecessary assignment. For the `cells/word` passed from constructor should be assigned by caller, not callee.

### How to test
- based on reviewing/fixing compiling error in evm execution gadgets

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/address.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/address.rs
@@ -1,20 +1,17 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
-        param::N_BYTES_HALF_WORD,
+        param::{N_BYTES_ACCOUNT_ADDRESS, N_BYTES_HALF_WORD},
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
             constraint_builder::{EVMConstraintBuilder, StepStateTransition, Transition::Delta},
-            CachedRegion,
+            AccountAddress, CachedRegion,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
     table::CallContextFieldTag,
-    util::{
-        word::{WordCell, WordExpr},
-        Expr,
-    },
+    util::{word::WordExpr, Expr},
 };
 use bus_mapping::evm::OpcodeId;
 use eth_types::{Field, ToAddress, ToLittleEndian};
@@ -23,7 +20,7 @@ use halo2_proofs::plonk::Error;
 #[derive(Clone, Debug)]
 pub(crate) struct AddressGadget<F> {
     same_context: SameContextGadget<F>,
-    address: WordCell<F>,
+    address: AccountAddress<F>,
 }
 
 impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
@@ -32,7 +29,9 @@ impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
     const EXECUTION_STATE: ExecutionState = ExecutionState::ADDRESS;
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let address = cb.query_word_unchecked();
+        // TODO switch to query_word_unchecked once callcontext -> XXXaddress encoded to Word2 type
+        // properly refer discussion thread https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1414/files#r1197845688
+        let address = cb.query_account_address();
 
         // Lookup callee address in call context.
         cb.call_context_lookup_read(None, CallContextFieldTag::CalleeAddress, address.to_word());
@@ -70,11 +69,12 @@ impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
         let address = block.rws[step.rw_indices[1]].stack_value();
         debug_assert_eq!(call.address, address.to_address());
 
-        self.address.assign_lo_hi(
+        self.address.assign(
             region,
             offset,
-            address.to_le_bytes()[0..N_BYTES_HALF_WORD].try_into().ok(),
-            address.to_le_bytes()[N_BYTES_HALF_WORD..].try_into().ok(),
+            address.to_le_bytes()[0..N_BYTES_ACCOUNT_ADDRESS]
+                .try_into()
+                .ok(),
         )?;
 
         Ok(())
--- a/zkevm-circuits/src/evm_circuit/execution/address.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/address.rs
@@ -1,16 +1,20 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
+        param::N_BYTES_HALF_WORD,
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
             constraint_builder::{EVMConstraintBuilder, StepStateTransition, Transition::Delta},
-            AccountAddress, CachedRegion,
+            CachedRegion,
         },
         witness::{Block, Call, ExecStep, Transaction},
     },
     table::CallContextFieldTag,
-    util::{word::WordExpr, Expr},
+    util::{
+        word::{WordCell, WordExpr},
+        Expr,
+    },
 };
 use bus_mapping::evm::OpcodeId;
 use eth_types::{Field, ToAddress, ToLittleEndian};
@@ -19,7 +23,7 @@ use halo2_proofs::plonk::Error;
 #[derive(Clone, Debug)]
 pub(crate) struct AddressGadget<F> {
     same_context: SameContextGadget<F>,
-    address: AccountAddress<F>,
+    address: WordCell<F>,
 }
 
 impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
@@ -28,7 +32,7 @@ impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
     const EXECUTION_STATE: ExecutionState = ExecutionState::ADDRESS;
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let address = cb.query_account_address();
+        let address = cb.query_word_unchecked();
 
         // Lookup callee address in call context.
         cb.call_context_lookup_read(None, CallContextFieldTag::CalleeAddress, address.to_word());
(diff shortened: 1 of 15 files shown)


zkevm-circuits/src/evm_circuit/execution/address.rs
```diff
@@ -28,7 +32,7 @@ impl<F: Field> ExecutionGadget<F> for AddressGadget<F> {
     const EXECUTION_STATE: ExecutionState = ExecutionState::ADDRESS;
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let address = cb.query_account_address();
```
hero78119, line 31: For `address` read part, can use just default word2 which just 2 limbs. Then assign first 16 byte to `lo` limb, and 4 bytes to `hi` limb

ed255: Yes.  In this case we're doing a read (cb.call_context_lookup_read, CallContextFieldTag::CalleeAddress), so we can just use 2 limbs to get that address and we can assume it's the right type, which means (lo: 16 byte, hi: 4 byte).
Next we have the `cb.stack_push` which uses that address.  Since the address is in a word obtained from a read, we already know it's correct, so there's no need to do any extra range check on this word in this gadget.

The same logic would apply if we read an Address from the RwTable, and we need to built a single expression.  We can just do: `let addr = addr_word.lo() + addr_word.hi() << 2^128`, and this new value is guaranteed to be 160 bits following the previous assumptions.

ed255: Sorry let me give an update because I think we can't use `cb.query_word_unchecked()` here.
In this opcode we're:
- Reading `address` from CallContext->CalleeAddress
- Writing `address` to Stack

Following the spec, stack values are Words, but CallContext->CalleeAddress is single field value https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#rw_table
So we read `[lo: address, hi: 0]`, and we write `[lo: address_lo, hi: address_hi]` where address_lo is the first 128 bits and address_hi is the remaining bits.

We need to guarantee the range check at the write, and since `address_lo` and `address_hi` will be new expressions, we need to range check them.  For example we can split `address` into range-checked bytes, and then from the bytes build the address_lo and address_hi expressions.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -901,12 +889,6 @@ impl<F: Field> SstoreGasGadget<F> {
         original_value: eth_types::Word,
         is_warm: bool,
     ) -> Result<(), Error> {
-        self.value
-            .assign(region, offset, Some(value.to_le_bytes()))?;
-        self.value_prev
-            .assign(region, offset, Some(value_prev.to_le_bytes()))?;
-        self.original_value
-            .assign(region, offset, Some(original_value.to_le_bytes()))?;
```
ed255, line 909: Perhaps these assignments were duplicated?

zkevm-circuits/src/evm_circuit/util/math_gadget/lt_word.rs
```diff
@@ -84,7 +84,8 @@ mod tests {
         fn configure_gadget_container(cb: &mut EVMConstraintBuilder<F>) -> Self {
             let a = cb.query_word32();
             let b = cb.query_word32();
-            let ltword_gadget = LtWordGadget::<F>::construct(cb, &a, &b);
+            let ltword_gadget =
+                LtWordGadget::<F, Word32Cell<F>, Word32Cell<F>>::construct(cb, a, b);
```
ed255, line 88: You could consider writing it like this, I think it should work:
```suggestion
            let ltword_gadget =
                LtWordGadget::<F, _, _>::construct(cb, a, b);
```

zkevm-circuits/src/util/word.rs
```diff
@@ -70,6 +77,56 @@ impl<F: Field, const N: usize> WordLimbs<Cell<F>, N> {
...
 
+    // N_LO, N_HI are number of bytes to assign to first half and second half of size N limbs,
+    // respectively N_LO and N_HI can be different size, the only requirement is N_LO % (N/2)
+    // and N_HI % (N/2) [N/2] limbs will be assigned separately.
+    // E.g. N_LO = 4 => [nl1, nl2, nl3, nl4]
+    // N_HI = 2 => [nh1, nh2]
+    // N = 2 => [l1, l2]
+    // it equivalent l1.assign(nl1.expr() + nl2.expr() * 256 + nl3.expr() * 256^2 +  nl3.expr() *
+    // 256^3) and l2.assign(nh1.expr() + nh2.expr() * 256)
+    pub fn assign_lo_hi<const N_LO: usize, const N_HI: usize>(
```
ed255, line 88: I think it will be very convenient to have an assign function that just takes a `eth_types::Word`, because that's the most common type we find from the witness data for Words.
Then the assignment can take the Word, turn it into bytes, and group them as necessary to assign each limb.

ed255 (commented):
Overall looks good! The only issue I found is that from the spec, an address in the CallContext is not split, so a conversion is needed when taking an address from CallContext and writing it to the Stack.

The other point is that I think it's convenient to have a Word assign method that takes an `eth_types::Word`

zkevm-circuits/src/evm_circuit/execution/address.rs
hero78119: Thanks for the review and explanation!

I checked [and](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#rw_table) and it seems that in CallContext, only CalleeAddress/CallerAddress those 2 currently cause value lo exceeding 128 bit, so it wont be a huge tech dept. I will rollback this part of change, and added the fix in pending task lists

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
hero78119: yes they are duplicate assignment. And those 3 cell(s) are passed from caller via constructor. Assignment them should be the caller responsibility.

zkevm-circuits/src/evm_circuit/execution/address.rs
hero78119: Fixed in commits https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1414/commits/7f2a59c8221a2cac5c09efa26025e4661bf0bfde

zkevm-circuits/src/evm_circuit/execution/caller.rs
```diff
@@ -30,18 +33,17 @@ impl<F: Field> ExecutionGadget<F> for CallerGadget<F> {
     const EXECUTION_STATE: ExecutionState = ExecutionState::CALLER;
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
-        let caller_address = cb.query_word_rlc();
+        let caller_address = cb.query_word_unchecked();
 
         // Lookup rw_table -> call_context with caller address
-        cb.call_context_lookup(
-            false.expr(),
+        cb.call_context_lookup_read(
```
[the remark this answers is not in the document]
hero78119: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1414/commits/7f2a59c8221a2cac5c09efa26025e4661bf0bfde
Adding `TODO`, and rollback this optimisation

zkevm-circuits/src/util/word.rs
hero78119: Nice suggestion 🙏 I added `assign_u256`

zkevm-circuits/src/evm_circuit/execution/address.rs
hero78119: Noticed that `TxContextFieldTag->CallerAddress`, `TxContextFieldTag->CalleeAddress` is the same situation. We can fix it later

hero78119: Hi @ed255 I just have another thought. We better fix it as earlier as possible instead of put it in optimisation. This is not purely a cell optimisation problem, instead more like a bug haha. Because design philosiphy we assume `lo`, `hi` < 128 bits however those 2 address type break the rules.  Besides, when I implementing `beingTx.rs`, `callop.rs`, it need some workaround code just to serve this purpose.  

So I suggest we fix specs part asap.

zkevm-circuits/src/evm_circuit/execution/caller.rs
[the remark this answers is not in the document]
hero78119: Still preserve the design in latest commit :), and reasoning in another comment https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1414/files#r1198700672

hero78119:
> @ed255 @hero78119 at this moment we have 3 equivalent types:
> 
> * `primitive_types::U256`
> * `eth_types::Word`
> * `utils::word::Word<F>`
> 
> I think that it will be nice if we have a rule when using this types.
> 
> I personally find it confusing the type `eth_types::Word`, so I would prefer always use `U256` and `utils::word::Word<F>` when we need to use to split into `F`s
> 
> WDYT?

@adria0 it sounds good to me! I think we can do this change at once in another PR :)

hero78119:
Hi @ed255 @adria0 Do you think it's good we merge this pr first once no other concern, since it's util related and shared for all circuit.

hero78119:
Update: support `assign_h160` on `Word` type for address assignment

hero78119:
closed, as word partially migrate to https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1441

Issue #971, opened 2022-12-07
CopyCircuit design and ChallengeAPI


ed255:
Currently the CopyCircuit contains a column called `value`.
When the dest is of type `CopyDataType::RlcAcc`, the read rows contain read bytes, and the write rows contain the accumulated RLC of the read bytes.

This means that the column contains the RLC inputs and outputs.  I fear this may break the soundness of using the Challenge API, where we need the inputs in phase n, and the RLC output in phase n+1.  I can only imagine that this would be OK if all the input bytes are already in other parts of the circuits (and connected to the CopyCircuit via lookups) and are in phase n.  But it seems a bit fragile?

ed255:
This issue may have been resolved recently.  We require a careful review to determine if this is the case.  See https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419#discussion_r1203740740

ed255:
After https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419 a new column was introduced to the CopyCircuit called `value_acc_rlc`.  Now the bytes stay in `value` column which is Phase1, and the rlc stays in `value_acc_rlc` which is Phase2, so we're no longer mixing assignments of RLC inputs and outputs in the same column.  This resolves this issue.

PR #1358, opened 2023-04-14
[CREATE Part C] bus-mapping and gadget implementation

roynalnaruto:
### Description

Part C

This PR is based on Part B (https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1357)
Part B is based on Part A (https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1356)

The commits to be reviewed are:
- https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1358/commits/699163689ad1fe00246d3c2d4867a478bf841f36
- https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1358/commits/2d2bfc6ccf179ade1a8d063f9586b93e5283a557

### Issue Link

#1130 

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Contents

- Bus mapping implementation for CREATE/CREATE2
- EVM circuit's gadget for CREATE/CREATE2

### How Has This Been Tested?

Tests can be found here: https://github.com/scroll-tech/zkevm-circuits/blob/2d2bfc6ccf179ade1a8d063f9586b93e5283a557/zkevm-circuits/src/evm_circuit/execution/create.rs#L678

PR #1357, opened 2023-04-14
[CREATE Part B] Error cases that will be handled within opcode gadgets

roynalnaruto:
### Description

Part B

This PR is actually based on top of #1356 . The only commit to be reviewed is: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1357/commits/65f72e4ded5aa907da0a4da5b33f9f8bd6e91549

### Issue Link

#1130 

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Contents

Error types for insufficient balance and nonce overflow. These are supporting changes required for the `CREATE`/`CREATE2` opcodes' gadget.

### Rationale

The above errors will be handled within the `CREATE`/`CREATE2` opcodes' gadget. In case of insufficient balance, it is also handled in the `CallOp` gadget for call related opcodes.

Issue #1130, opened 2023-02-02
Implement CREATE & CREATE2


aguzmant103:
Implementing CREATE & CREATE2 execution states in the EVM circuit. See the spec first https://github.com/privacy-scaling-explorations/zkevm-specs/issues/371

roynalnaruto:
@aguzmant103 I am working on this currently. WIP is in the forked repo here: https://github.com/scroll-tech/zkevm-circuits/blob/feat/create-create2/zkevm-circuits/src/evm_circuit/execution/create.rs

hero78119:
> @aguzmant103 I am working on this currently. WIP is in the forked repo here: https://github.com/scroll-tech/zkevm-circuits/blob/feat/create-create2/zkevm-circuits/src/evm_circuit/execution/create.rs

@adria0 here :)

roynalnaruto:
Another scenario to be handled within the `Create` gadget is "insufficient balance" in caller's account.

WIP on the fork repo: https://github.com/scroll-tech/zkevm-circuits/pull/369

PR #1419, opened 2023-05-19
[CREATE Part A (updated)] Modification to Copy Circuit

ed255:
### Description

NOTE: This is an updated version of https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1356

This is Part A of a 3 part pull request to add support for `CREATE`/`CREATE2` opcodes.

Part A: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1356
Part B: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1357
Part C: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1358

### Issue Link

#1130 

### Type of change

### Contents
 
As part of the bigger additions needed for the `CREATE`/`CREATE2` opcodes' gadget, this PR adds support for the copy circuit to "always" have a value accumulator field `value_acc`.

### Rationale

We need a value accumulator (of the random linear combination) in order to get the `RLC(bytes)` for the bytes copied from `Memory` to `Bytecode` (specifically the init code). This RLC is later used to do a lookup to the Keccak table to check the value of `keccak256(init_code)`.

### How Has This Been Tested?

The existing tests for copy circuit pass for the updated constraints on the copy circuit.


ed255:
This PR is a continuation of https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1356
The requests and comments have not been addressed yet, that's the next step before merging.

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -301,25 +322,9 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
...
-                    + meta.query_advice(value, Rotation::next()),
-            );
-
             cb.gate(and::expr([
                 meta.query_fixed(q_enable, Rotation::cur()),
-                not::expr(meta.query_selector(q_step)),
-                not::expr(meta.query_advice(is_last, Rotation::cur())),
-                tag.value_equals(CopyDataType::RlcAcc, Rotation::cur())(meta),
-                not::expr(meta.query_advice(is_pad, Rotation::cur())),
+                meta.query_selector(q_step),
```
han0110, line 327: We can remove line 326 since `q_enable = 1` when `q_step = 1`.

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -265,25 +280,31 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
...
+                meta.query_advice(value_acc_rlc, Rotation::next()),
+            );
             cb.condition(
-                not::expr(tag.value_equals(CopyDataType::RlcAcc, Rotation::next())(
-                    meta,
-                )),
+                and::expr([
+                    not::expr(meta.query_advice(is_last, Rotation::next())),
+                    not::expr(meta.query_advice(is_pad, Rotation::cur())),
+                ]),
```
han0110, line 297: This'd be a little bit cheaper
```suggestion
                not::expr(
                    and::expr([
                        meta.query_advice(is_last, Rotation::next()),
                        meta.query_advice(is_pad, Rotation::cur()),
                    ]),
                ),
```

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -265,25 +280,31 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
...
                 |cb| {
                     cb.require_equal(
-                        "write value == read value (if not rlc acc)",
-                        meta.query_advice(value, Rotation::cur()),
-                        meta.query_advice(value, Rotation::next()),
+                        "value_acc_rlc(2) == value_acc_rlc(0) * r + value(2)",
+                        meta.query_advice(value_acc_rlc, Rotation(2)),
+                        meta.query_advice(value_acc_rlc, Rotation::cur())
+                            * challenges.keccak_input()
+                            + meta.query_advice(value, Rotation(2)),
```
han0110, line 304: It seems we don't have constraint to enforce for the first step of each copy event, the `value_acc_rlc` should be equal to `value`. A malicious prover now can start with any value.

zkevm-circuits/src/copy_circuit.rs
KimiWu123: Sorry, I didn't see any place constraint `q_enable = 1` when `q_step = 1`. Could you please show me where?

zkevm-circuits/src/copy_circuit.rs
roynalnaruto: You're right. After changing the layout I must have forgotten to re-add that constraint. Thank you for noticing!

zkevm-circuits/src/copy_circuit.rs
KimiWu123: fixed in [d2bdc68](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419/commits/d2bdc68414d218732a56d4e1d038b1c11859c9e6)

zkevm-circuits/src/copy_circuit.rs
KimiWu123: fixed in [d2bdc68](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419/commits/d2bdc68414d218732a56d4e1d038b1c11859c9e6)

zkevm-circuits/src/copy_circuit.rs
han0110: Since both `q_enable` and `q_step` are fixed column, and from the design (or assignment) `q_enable` is for every row (except blinding factors) and `q_step` is for every 2-row (except blinding factors).

KimiWu123: fixed in [d591f47](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419/commits/d591f47ab9558031eacf11e4e04b148685db7654)

zkevm-circuits/src/copy_circuit.rs
han0110: The following question is, is there any constraint to enforce `is_first` to be `1` when we are entering next copy event? Since I couldn't find an explicit one for that, and after I change the `is_first` assignment to all `0` the tests are still passing.

zkevm-circuits/src/copy_circuit.rs
KimiWu123: @han0110 , it should be `or` instead of `and`. Could you help to confirm? fixed in [b696d4a](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419/commits/b696d4a9bf92904ad71c8234fd21a9986091eb40)
```rust
                not::expr(
                    or::expr([
                        meta.query_advice(is_last, Rotation::next()),
                        meta.query_advice(is_pad, Rotation::cur()),
                    ]),
                ),
```

zkevm-circuits/src/copy_circuit.rs
KimiWu123: >  from the design (or assignment) q_enable is for every row (except blinding factors) and q_step is for every 2-row (except blinding factors).

I could understand from the design we'll assign `q_enable` for every row but I thought we can't trust assignment part. Which means malicious provers could set `q_enable = 0` when `q_step = 1` and the constraints here will be bypassed.

han0110: But for `q_enable` and `q_step` they are both preprocess polynomial, so I think we can trust the assignment, right?

KimiWu123: yup, make sense, you're right! thanks.

zkevm-circuits/src/copy_circuit.rs
KimiWu123: https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1432 created for `is_first` and `is_last`

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -225,23 +228,35 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
...
+                cb.gate(and::expr([
+                    meta.query_fixed(q_enable, Rotation::cur()),
+                    meta.query_advice(is_last, Rotation::next()),
+                    or::expr([
+                        and::expr([
+                            tag.value_equals(CopyDataType::Memory, Rotation::cur())(meta),
+                            tag.value_equals(CopyDataType::Bytecode, Rotation::next())(meta),
+                        ]),
+                        tag.value_equals(CopyDataType::RlcAcc, Rotation::next())(meta),
+                    ]),
```
ed255, line 255: The expression from here may be causing the final gate degree ti increase more than 9 (and that's why tests are not passing).  A simple solution would be to split this into 2 constraints. Following this pattern:
`if A or B then X` -> `if A then X`, `if B then X`.

Another thing that comes to my mind is that `or::expr` returns a boolean, but to build a selector expression just having `0` when false and `!= 0` when true is enough, so perhaps you could replace the `or` by a `+`.  This would give `2` when both expressions are true, but that's ok.

ed255 (commented):
Currently the tests are not passing due to a degree increase of the gates expressions, take a look at my comment for a possible fix.  Following Han observation about the missing constraints on `is_first` and `is_last` I think a careful review of this circuit could be really useful!  I suggest that we resolve the underconstraints in a future PR (Kimi already opened an issue for that), and move on with this PR once the tests pass to unblock the CREATE.

zkevm-circuits/src/copy_circuit.rs
```diff
@@ -50,6 +50,8 @@ pub struct CopyCircuitConfig<F> {
     pub is_last: Column<Advice>,
     /// The value copied in this copy step.
     pub value: Column<Advice>,
+    /// Random linear combination accumulator value.
+    pub value_acc_rlc: Column<Advice>,
```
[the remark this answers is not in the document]
KimiWu123: This was @roynalnaruto's fix and not sure the reason behind. But it seems solved the issue you mentioned.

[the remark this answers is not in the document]
han0110: I think to have an extra `value_acc_rlc` is to support `memory <-> bytecode` copy and rlc at the same time without changing other parts too much.

And it seems to also resolve #971, so we could try to move the phase of `value` from second to first and see if it works as expected.

zkevm-circuits/src/copy_circuit.rs
KimiWu123: Nice idea! It looks nice and clean now, fixed in [52d480a](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1419/commits/52d480a1dcade244606b6d2598abd1f756a80c1d)

han0110 (approved):
LGTM now! Thanks for addressing all the comments.

KimiWu123 (approved):

Issue #1420, opened 2023-05-19
testool run very slow after `69cb8a932`


ChihChengLiang:
### What command(s) is the bug in?

`../target/release/testool --oneliner "call 620a90005100"`

### Describe the bug

This script `../target/release/testool --oneliner "call 620a90005100"` takes about 10~15 secs to run on my M1 device before `69cb8a932`. After that, this script hangs forever.

Git bisect shows this seemly innocent PR #1406 introduced the behavior.

### Concrete steps to reproduce the bug. If it's able reproduce via testool, please share `test_id` from jenkins report

`cd testool`

Create a script `bisect.sh` with the following content

```
# bisect.sh
cargo build --release
time ../target/release/testool --oneliner "call 620a90005100"
```
`git checkout 438bac5bd && sh bisect.sh` this should show `result=Ok(())` in about 15 sec.

`git checkout 69cb8a932 && sh bisect.sh` this hangs forever.

ed255:
I can reproduce this on linux amd64 arch.  I've also tried with a debug build and I get a segfault.  I encountered segfaults before on debug builds coming from a stack overflow.  I've tried a very big stack with `RUST_MIN_STACK=167772160` but still get the segfault

ed255:
oh, I may have found something, I think there may be an infinite loop of `into()`s

ed255:
I believe this may be the source of the problem:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/69cb8a9329a3f5263e0ebb4ee003dd26a9470107/bus-mapping/src/operation.rs#L122-L126

When calling `value.into()` we want to convert a `Target` into `usize`, so this function calls itself in an infinite loop.

ed255:
I've made a PR with a possible fix, let's wait for the CI tests to pass to confirm it's resolved: https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1421

Issue #1417, opened 2023-05-19
The recent CI hanging and failure


ChihChengLiang:
### Describe the bug

There are two styles of failure:

The first one is that the self-host runner couldn't pick up the task. The current instances were canceled, so I'll post a new one here if the failure happens again.

The second one is the CI hangs at a particular stage. Notably, the "Run light tests"
- https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/5011576441/jobs/9003329648 ![Screenshot 2023-05-19 at 3 39 33 PM](https://github.com/privacy-scaling-explorations/zkevm-circuits/assets/3391420/14f1f1b6-b643-4672-8b70-6cfd89e43fbc)
- https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/5004358516/jobs/8966828870 This one, the operation was canceled after hanging 6hrs.

AronisAt79:
@ChihChengLiang 

same freeze condition occurs when testtool is invoked manually.
```
running 13 tests
test abi::test::test_abi_encoding ... ok
test statetest::json::test::test_json_parse ... ok
test statetest::yaml::test::combinations ... ok
test utils::test::networks ... ok
test statetest::yaml::test::parse ... ok
test statetest::yaml::test::marked_as_exception_and_fails ... ok
test statetest::yaml::test::marked_as_exception_but_does_not_fail ... ok
test statetest::yaml::test::bad_balance has been running for over 60 seconds
test statetest::yaml::test::bad_code has been running for over 60 seconds
test statetest::yaml::test::bad_nonce has been running for over 60 seconds
test statetest::yaml::test::result_pass has been running for over 60 seconds
test statetest::yaml::test::sstore has been running for over 60 seconds
test statetest::yaml::test::test_result_bad_storage has been running for over 60 seconds
```
strace shows that testtool process is using a futex system call to wait for some condition to be met:

```
futex(0x55a5655c4c88, FUTEX_WAIT_BITSET_PRIVATE, 4294967295, NULL, FUTEX_BITSET_MATCH_ANY
```
we also notice that nightly run on jenkins test automation is stack in a similar manner since May 18th. ( we have similar strace indication), but this happens in heavy tests as well.

While investigation is ongoing, 
@adria0 are you able to further debug via a local run of the light test suite? Are you also maybe aware if there are any recent changes in testtool that could break the workflow or cause it to get stack in waiting mode?

[...]

PR #1406, opened 2023-05-12
Remove RwTableTag

ChihChengLiang:
### Description

`RwTableTag` can be fully replaced by `Target`.

### Issue Link

Part of #1391 

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Contents

We add EnumIter, Hash, Expr to Target to let it serve different needs.

Note to reviewers: the `impl From<Target> for usize` is required by BinaryNumberConfig.

PR #1421, opened 2023-05-19
Fix infinite loop

ed255:
### Description

Fix an infinite loop from a `From` implementation calling itself.

### Issue Link

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1420
Related https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1417

### Type of change

### Contents

@ChihChengLiang found that this issue was introduced in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1406 and by debugging on the debug build I found the infinite loop (which caused the stack to overflow which terminated the program with a segfault).  The segfault is not observed in the release build, I guess the compiler optimizes the recursive call into an infinite loop, so the stack is never exhausted.

### DIFF
--- a/bus-mapping/src/operation.rs
+++ b/bus-mapping/src/operation.rs
@@ -121,7 +121,7 @@ impl_expr!(Target);
 
 impl From<Target> for usize {
     fn from(value: Target) -> usize {
-        value.into()
+        value as usize
     }
 }
 


ChihChengLiang (approved):
Nice catch!
Didn't know that into was that dangerous

ed255:
I'll merge this with one approve because it's a very small fix.

Issue #1382, opened 2023-05-03
State Circuit: Refactor word_rlc into word lo/hi


ed255:
Specs: https://github.com/privacy-scaling-explorations/zkevm-specs/pull/390

Depends on:
- https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1388

ChihChengLiang:
Fixed by #1423

PR #1423, opened 2023-05-19
State Circuit: Refactor word_rlc into word lo/hi

adria0:
### Description

State Circuit: Refactor word_rlc into word lo/hi

This PR does not compile for itself, anyway the changes introduced in this PR passes the state circuit tests [in another branch](https://github.com/privacy-scaling-explorations/zkevm-circuits/tree/statecircuit-lo-hi-hack)

### Issue Link

- https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1382

### Contents

- As expected, changing rlc encoded values into `word::Word`, as defined in [`rw_table` specs](https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#rw_table)
- Add helper methods to `ConstraintBuilder` (like `require_word_zero` )
- Make MPI able to manage multi-field values (to be able to use U256's)
- Remove `random_linear_combination.rs` because Word uses MPI instead rlc
- Removed test focused on test rlc
- Removed unused field `aux1` and renamed `aux2` to `init_val`
- Added helper methods to `word::Word`
- Also updated `mpt` witness to pass the state tests
- Apply https://github.com/privacy-scaling-explorations/zkevm-specs/pull/424

### Rationale

The lexicographic ordering, used word_rlc randomness for [another use](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/e3e6034a11349cb07dd08e88026e3ca40d771a66/zkevm-circuits/src/state_circuit/lexicographic_ordering.rs#L312). Since this randomess is going to be removed, we have three options here:

- Use another method to build the constraints that does not uses rlc at all
- Create another challenge value
- Use another existing randomness

In order to change only the code concerning hi/lo refactor, the choice is to use the [keccak randomness](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/e3e6034a11349cb07dd08e88026e3ca40d771a66/zkevm-circuits/src/state_circuit.rs#L97) instead

### DIFF
--- a/zkevm-circuits/src/state_circuit.rs
+++ b/zkevm-circuits/src/state_circuit.rs
@@ -270,7 +270,7 @@ impl<F: Field> StateCircuitConfig<F> {
             }
 
             // The initial value can be determined from the mpt updates or is 0.
-            let initial_value = word::Word::<F>::from_u256(
+            let initial_value = word::Word::<F>::from(
                 updates
                     .get(row)
                     .map(|u| u.value_assignments().1)
@@ -292,8 +292,8 @@ impl<F: Field> StateCircuitConfig<F> {
                     .unwrap_or_default();
                 let value = row.value_assignment();
                 (
-                    word::Word::<F>::from_u256(committed_value),
-                    word::Word::<F>::from_u256(value),
+                    word::Word::<F>::from(committed_value),
+                    word::Word::<F>::from(value),
                 )
             };
 
@@ -340,7 +340,7 @@ impl<F: Field> StateCircuitConfig<F> {
             // State root assignment is at previous row (offset - 1) because the state root
             // changes on the last access row.
             if offset != 0 {
-                word::Word::<F>::from_u256(state_root)
+                word::Word::<F>::from(state_root)
                     .into_value()
                     .assign_advice(region, || "state root", self.state_root, offset - 1)?;
             }
@@ -355,7 +355,7 @@ impl<F: Field> StateCircuitConfig<F> {
                         new_root
                     };
                 }
-                word::Word::<F>::from_u256(state_root)
+                word::Word::<F>::from(state_root)
                     .into_value()
                     .assign_advice(region, || "last row state_root", self.state_root, offset)?;
             }
--- a/zkevm-circuits/src/util/word.rs
+++ b/zkevm-circuits/src/util/word.rs
@@ -3,7 +3,7 @@
 // - Limbs: An EVN word is 256 bits. Limbs N means split 256 into N limb. For example, N = 4, each
 //   limb is 256/4 = 64 bits
 
-use eth_types::{Field, ToLittleEndian, U256};
+use eth_types::{Field, ToLittleEndian, H160};
 use gadgets::util::{not, or, Expr};
 use halo2_proofs::{
     circuit::{AssignedCell, Region, Value},
@@ -12,6 +12,9 @@ use halo2_proofs::{
 use itertools::Itertools;
 
 use crate::evm_circuit::util::{from_bytes, CachedRegion, Cell, RandomLinearCombination};
+
+/// evm word 32 bytes, half word 16 bytes
+const N_BYTES_HALF_WORD: usize = 16;
 
 /// The EVM word for witness
 #[derive(Clone, Debug, Copy)]
@@ -83,6 +86,13 @@ pub trait WordExpr<F> {
 
 impl<F: Field, const N: usize> WordLimbs<Cell<F>, N> {
     /// assign limbs
+    /// N1 is number of bytes to assign, while N is number of limbs.
+    /// N1 % N = 0 (also implies N1 >= N, assuming N1 and N are not 0)
+    /// If N1 > N, then N1 will be chunk into N1 / N size then aggregate to single expression
+    /// then assign to N limbs respectively.
+    /// e.g. N1 = 4 bytes, [b1, b2, b3, b4], and N = 2 limbs [l1, l2]
+    /// It equivalent `l1.assign(b1.expr() + b2.expr * F(256))`, `l2.assign(b3.expr() + b4.expr *
+    /// F(256))`
     pub fn assign<const N1: usize>(
         &self,
         region: &mut CachedRegion<'_, '_, F>,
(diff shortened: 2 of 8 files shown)


ChihChengLiang:
Can we do a Clippy round before the review?

zkevm-circuits/src/state_circuit.rs
```diff
@@ -529,6 +515,14 @@ fn queries<F: Field>(meta: &mut VirtualCells<'_, F>, c: &StateCircuitConfig<F>)
         + meta.query_advice(first_different_limb.bits[4], Rotation::cur());
     let mpt_update_table_expressions = c.mpt_table.table_exprs(meta);
 
+    let meta_query_word =
+        |metap: &mut VirtualCells<'_, F>, word_column: word::Word<Column<Advice>>, at: Rotation| {
+            word::Word::new([
+                metap.query_advice(word_column.lo(), at),
+                metap.query_advice(word_column.hi(), at),
+            ])
+        };
```
ChihChengLiang, line 524: I was thinking about what is a good general API for querying the word limbs. In the keccak PR, I did

`hash_word.map(|col| meta.query_advice(col, Rotation::cur())),`

But it feels verbose.

zkevm-circuits/src/state_circuit.rs
```diff
@@ -576,55 +588,54 @@ fn queries<F: Field>(meta: &mut VirtualCells<'_, F>, c: &StateCircuitConfig<F>)
...
-            bytecode_prefix_op_big_rws,
-            |block, _, step_index| {
-                let step = &block.txs[0].steps()[step_index];
-                let step_next = &block.txs[0].steps()[step_index + 1];
-                step_next.rwc.0 - step.rwc.0
-            },
-        );
-    }
-}
+// #[cfg(test)]
```
ChihChengLiang, line 606: Can we keep these tests as they were?

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -626,7 +671,10 @@ impl<F: Field> Queries<F> {
     }
 
     fn address_change(&self) -> Expression<F> {
-        self.rw_table.address.clone() - self.rw_table.prev_address.clone()
+        let pow_2_128 = Expression::Constant(F::from_str_vartime(POW_2_128_STR).unwrap());
+
+        (self.rw_table.address.hi() - self.rw_table.prev_address.hi()) * pow_2_128
+            + (self.rw_table.address.lo() - self.rw_table.prev_address.lo())
```
ChihChengLiang, line 677: Is this safe?

zkevm-circuits/src/state_circuit/test.rs
```diff
@@ -421,38 +406,11 @@ fn storage_key_mismatch() {
...
-    let overrides = HashMap::from([((AdviceColumn::StorageKeyByte1, 0), Fr::ONE)]);
+    let overrides = HashMap::from([((AdviceColumn::StorageKeyLimb0, 0), Fr::ONE)]);
 
     let result = verify_with_overrides(rows, overrides);
 
-    assert_error_matches(result, "rlc encoded value matches bytes");
-}
-
-#[test]
-fn storage_key_byte_out_of_range() {
```
ChihChengLiang, line 432: Do we remove this test because it is rlc related, so it has become obsolete now?

ChihChengLiang (commented):
I did first pass.
Would be great if we can fix the Rust fmt.

zkevm-circuits/src/table.rs
```diff
@@ -444,28 +444,17 @@ pub struct RwTable {
     /// Key1 (Id)
     pub id: Column<Advice>,
     /// Key2 (Address)
-    pub address: Column<Advice>,
+    pub address: word::Word<Column<Advice>>,
```
ChihChengLiang, line 447: We shouldn't turn this into a word. As the spec says, the address column is used for multiple purposes. Let me quote them below.

https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/specs/tables.md#rw_table

>  - col. 4 *Address* is the position to **Stack**, **Memory**, or account, where the read or write takes place, depending on the cell value of the *Tag* column.
        - If the *Tag* value is "Account", the cell represents 160 bits **address**.
        - If the *Tag* value is "Stack", the cell represents 10 bits  **stackPointer**.
        - If the *Tag* value is "Memory", the cell represents 32 bits **memoryAddress**.
        - If the *Tag* value is "TxLog", then the cell represents the packing of 2 values:
            - **logID**: 32 bits, starts at 1 (corresponds to `logIndex + 1`), unique per tx/receipt.
            - **topicIndex, byteIndex**: 32 bits, indicates order in tx log topics or data.

Most of the usages can fit into a cell. The only usage that might be a word is the "Eth address". Since lots of our API returns address as a word, we might tempt to change Rwtable's address column into word columns. But considering the other use cases, changing the address to word columns adds confusion to the stackPointer and Memory address use cases.

ChihChengLiang (commented):
I figured out the `address_change` function. I added some proposals for the `rw_table.address` field.

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -153,10 +168,14 @@ impl<F: Field> ConstraintBuilder<F> {
...
         self.condition(q.first_access(), |cb| {
             cb.require_zero(
-                "first access reads don't change value",
-                q.is_read() * (q.rw_table.value.clone() - q.initial_value()),
+                "first access reads don't change value (hi)",
+                q.is_read() * (q.rw_table.value.hi() - q.initial_value().hi()),
+            );
+            cb.require_zero(
+                "first access reads don't change value (lo)",
+                q.is_read() * (q.rw_table.value.lo() - q.initial_value().lo()),
```
hero78119, line 176: line 171, 174 (and few ditto below) can be combined into
```
cb.require_zero_word(
    "first access reads don't change value",
    q.rw_table.value.sub_unchecked(q.initial_value()).mul_selector(q.is_read())
)
```

zkevm-circuits/src/state_circuit/multiple_precision_integer.rs
```diff
@@ -1,32 +1,48 @@
...
 impl ToLimbs<N_LIMBS_ACCOUNT_ADDRESS> for Address {
-    fn to_limbs(&self) -> [u16; 10] {
+    fn to_limbs(&self) -> [u16; N_LIMBS_ACCOUNT_ADDRESS] {
         // address bytes are be.... maybe just have everything be later?
         // you will need this in the future later because it makes the key ordering more
         // obvious
         let le_bytes: Vec<_> = self.0.iter().rev().cloned().collect();
         le_bytes_to_limbs(&le_bytes).try_into().unwrap()
     }
+    fn annotation() -> &'static str {
```
hero78119, line 25: just a qn: for this kind of constant retrieval, should we add `#[inline(always)]` for it

zkevm-circuits/src/table.rs
hero78119: This is quite an open question and discussed in another pr. I also vote +1  to have just single cell in `rwtable.address` since it's used for multi-purposes. and also benefit to save one extra `hi` cell for lookup.

hero78119 (commented):
First round quick pass, it looks really nice to clean up many randomness in state circuit. Beside also replace aux1 aux2 with better naming! few other comments

ChihChengLiang:
I removed the review request for @ed255 since we have two reviewers here already.

zkevm-circuits/src/table.rs
```diff
@@ -937,8 +945,7 @@ impl<F: Field> LookupTable<F> for KeccakTable {
             String::from("is_enabled"),
             String::from("input_rlc"),
             String::from("input_len"),
-            String::from("output_lo"),
-            String::from("output_hi"),
+            String::from("output_rlc"),
```
ChihChengLiang, line 948: Do we intend to revert something here?

zkevm-circuits/src/util/word.rs
```diff
@@ -1,13 +1,14 @@
+#![allow(missing_docs)]
```
ChihChengLiang, line 1: It's okay to keep this here. This is a good tech debt.
I just want to check if this is a leftover.

ChihChengLiang (commented):
I finished reading all the diffs and added more comments.

ChihChengLiang (commented):
I finished reading all the diffs and added more comments.

zkevm-circuits/src/table.rs
```diff
@@ -517,39 +506,47 @@ impl RwTable {
...
+            value_prev: word::Word::new([meta.advice_column(), meta.advice_column()]),
+            init_val: word::Word::new([meta.advice_column(), meta.advice_column()]),
         }
     }
     fn assign<F: Field>(
         &self,
         region: &mut Region<'_, F>,
         offset: usize,
-        row: &RwRow<Value<F>>,
+        row: &RwRow<F>,
```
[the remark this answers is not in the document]
ChihChengLiang: If an assign function is not using `Value<F>` then it is a tech debt. Bugs like #1456 would happen if we use F in an assignment function.

zkevm-circuits/src/state_circuit/constraint_builder.rs
hero78119: > Why not implement
> 
> ```
> impl<F: Field> std::ops::Add for Word<Expression<F>>{
>     type Output = Word<Expression<F>>;
>     
>     fn add(self, rhs: Word<Expression<F>>) -> Self::Output {
>         Word::new([self.lo() + rhs.lo(), self.hi() + rhs.hi()]) 
>     }
> } 
> 
> impl<F: Field> std::ops::Sub for Word<Expression<F>>{
>     type Output = Word<Expression<F>>;
>     
>     fn sub(self, rhs: Word<Expression<F>>) -> Self::Output {
>         Word::new([self.lo() - rhs.lo(), self.hi() - rhs.hi()])
>     }
> } 
> ```
> 
> ?

Because by spec lo/hi should within 128bit, and purely add/sub we can't avoid overflow/underflow issue

zkevm-circuits/src/table.rs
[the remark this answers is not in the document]
hero78119: With more precisely,  consider below case
case 1: caller have `Value` in hands and calling a lib(chip) assign fn taking `F`
case 2: caller have `Value` in hands and calling a lib(chip) assign fn taking `Value`
case 3: caller have `F` in hands and calling a lib(chip) assign fn taking `F`
case 4: caller have `F` in hands and calling a lib(chip) assign fn taking `Value`

only case 1 will cause bug a bug for `assign` not called during key gen phase, since Value::unknown.map(fn) wont call. 
Case 3, 4 style will not have the bug.

Case 2 style can be our rule-of-thumb, but do you have any idea to introduce a way by `trait` + type constraints, or a lint to assure case 2 style across the whole codebase? It will be good to have comparing with manual checking

ChihChengLiang (approved):
LGTM after addressing @hero78119's comments. Great word adding lo-hi and cleaning up this big chunk of the codebase.

hero78119 (approved):
All LGTM!! Few discussion can be moved to discuss in separate issue

PR #1429, opened 2023-05-22
Fix EXTCODECOPY with empty account

ed255:
### Description

We treat empty accounts by storing their code_hash in the RwTable as 0.  EXTCODECOPY was obtaining the bytecode length by querying the bytecode table with code_hash=0 on existing accounts, but that entry should be invalid (there's no bytecode with code_hash=0).  Skip the bytecode table length lookup when code_hash=0.

I've also reintroduced the `Block::debug_print_txs_steps_rw_ops` function, updated to use the new `Block::get_rws` API.  This function is not used in the code, but it's very convenient to call it when debugging.

### Issue Link

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1190

### Type of change

### DIFF
--- a/bus-mapping/src/evm/opcodes/extcodecopy.rs
+++ b/bus-mapping/src/evm/opcodes/extcodecopy.rs
@@ -11,9 +11,6 @@ use eth_types::{Bytecode, GethExecStep, ToAddress, ToWord, H256, U256};
 #[derive(Clone, Copy, Debug)]
 pub(crate) struct Extcodecopy;
 
-// TODO: Update to treat code_hash == 0 as account not_exists once the circuit
-// is implemented https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/720
-
 impl Opcode for Extcodecopy {
     fn gen_associated_ops(
         state: &mut CircuitInputStateRef,
--- a/bus-mapping/src/state_db.rs
+++ b/bus-mapping/src/state_db.rs
@@ -8,6 +8,7 @@ use std::collections::{HashMap, HashSet};
 
 lazy_static! {
     static ref ACCOUNT_ZERO: Account = Account::zero();
+    /// Hash value for empty code hash.
     static ref EMPTY_CODE_HASH: Hash = CodeDB::hash(&[]);
     /// bytes of empty code hash, in little endian order.
     pub static ref EMPTY_CODE_HASH_LE: [u8; 32] = {
--- a/zkevm-circuits/src/evm_circuit/execution/extcodecopy.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/extcodecopy.rs
@@ -9,6 +9,7 @@ use crate::{
                 Transition,
             },
             from_bytes,
+            math_gadget::IsZeroGadget,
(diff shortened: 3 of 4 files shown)


ChihChengLiang (approved):
LGTM for the addition and the reintroduction of debug function. Feel free to merge.

I noticed that we have `cb.bytecode_length` in 6 places. 

- In CODESIZE and EXTCODESIZE, we look up the bytecode length for the opcode output.
- In CODECOPY and EXTCODECOPY, length determines the destination address in memory.
- In ErrInvalidJUMP, we check if we jump out of the range of the code size. 
- In STOP, we check program counter is not out of the range of code size.

For CODESIZE, CODECOPY, ErrInvalidJUMP, and STOP, we can safely assume we don't query empty hash in those cases, not even in the init code. 

For EXTCODESIZE and EXTCODECOPY, the empty hash checks look very similar, maybe execution state merging is a possible optimization in the future.

ed255:
> I noticed that we have `cb.bytecode_length` in 6 places.
> 
>     * In CODESIZE and EXTCODESIZE, we look up the bytecode length for the opcode output.
>     * In CODECOPY and EXTCODECOPY, length determines the destination address in memory.
>     * In ErrInvalidJUMP, we check if we jump out of the range of the code size.
>     * In STOP, we check program counter is not out of the range of code size.
> 
> 
> For CODESIZE, CODECOPY, ErrInvalidJUMP, and STOP, we can safely assume we don't query empty hash in those cases, not even in the init code.
> 
> For EXTCODESIZE and EXTCODECOPY, the empty hash checks look very similar, maybe execution state merging is a possible optimization in the future.

Nice summary!  Indeed `EXTCODESIZE` and `EXTCODECOPY` are the two opcodes that can deal with non-existing bytecode. The code that overlaps between the two is:

bus-mapping code:
```rust
        let exists = !account.is_empty();
        let code_hash = if exists {
            account.code_hash
        } else {
            H256::zero()
        };
```

circuit code:
```rust
        /// configure
        let code_hash = cb.query_cell_phase2();
        // For non-existing accounts the code_hash must be 0 in the rw_table.
        cb.account_read(address.expr(), AccountFieldTag::CodeHash, code_hash.expr());
        let not_exists = IsZeroGadget::construct(cb, code_hash.expr());
        let exists = not::expr(not_exists.expr());

        let code_size = cb.query_word_rlc();
        cb.condition(exists.expr(), |cb| {
            cb.bytecode_length(code_hash.expr(), from_bytes::expr(&code_size.cells));
        });
        cb.condition(not_exists.expr(), |cb| {
            cb.require_zero("code_size is zero when non_exists", code_size.expr());
        });
        /// assignment
        self.code_hash
            .assign(region, offset, region.word_rlc(code_hash))?;
        self.not_exists
            .assign_value(region, offset, region.word_rlc(code_hash))?;
```

I think it's not a lot of code to worry about code deduplication.  About merging the two `ExecState`s, the EXTCODECOPY performs a lot of extra logic that would need to become conditional (potentially increasing the constraints degree, which means introducing new intermediate cells to keep the degree low by the split_expression engine), so maybe it can be explored but I'm not confidence that there's a clear gain.

ed255:
I have just merged this PR with a single review because it was a very small fix and it doesn't introduce any new pattern.

PR #1435, opened 2023-05-26
Fix build on word-lo-hi branch

ChihChengLiang:
### Description

The current plan to implement word-lo-hi is to divide the work for different circuits among different implementors.

The problem is that the word-lo-hi branch does not build, so the implementors of different circuits do not know if they are implementing the circuit correctly.

The word-lo-hi branch does not build because we introduced many incompatible changes that propagate in the codebase beyond the reasonable effort to fix the build. We can resolve this issue by bringing back some old functions and gadgets for compatibility. Meanwhile, we mark them as deprecated.


zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -848,15 +864,26 @@ pub(crate) struct SstoreGasGadget<F> {
 
 impl<F: Field> SstoreGasGadget<F> {
     pub(crate) fn construct(
+        _cb: &mut EVMConstraintBuilder<F>,
+        _value: Cell<F>,
+        _value_prev: Cell<F>,
+        _original_value: Cell<F>,
+        _is_warm: Cell<F>,
+    ) -> Self {
+        todo!()
```
ChihChengLiang, line 873: I left many todos. Those can make the build pass but cripple the tests.

zkevm-circuits/src/evm_circuit/util.rs
```diff
@@ -1,3 +1,8 @@
+pub use crate::util::{
+    query_expression,
+    word::{Word as WordNew, WordExpr, WordLegacy as Word},
```
ChihChengLiang, line 3: Heads up for the messy import/export here.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -978,7 +1005,7 @@ impl<F: Field> CommonErrorGadget<F> {
     ) -> Self {
         cb.opcode_lookup(opcode.expr(), 1.expr());
 
-        let rw_counter_end_of_reversion = cb.query_word_unchecked(); // rw_counter_end_of_reversion just used for read lookup, therefore skip range check
+        let rw_counter_end_of_reversion = cb.query_word_unchecked::<2>(); // rw_counter_end_of_reversion just used for read lookup, therefore skip range check
```
ChihChengLiang, line 1008: Note that the compiler requires me to add `::<T>` here. And I'm not sure if `::<2>` is sensible here. cc @hero78119.

zkevm-circuits/src/evm_circuit/util/math_gadget.rs
```diff
@@ -39,7 +39,8 @@ pub(crate) use is_equal_word::IsEqualWordGadget;
 pub(crate) use is_zero::IsZeroGadget;
 pub(crate) use is_zero_word::IsZeroWordGadget;
 pub(crate) use lt::LtGadget;
-pub(crate) use lt_word::LtWordGadget;
+pub(crate) use lt_word::LtWordGadget as LtWordGadgetNew;
+pub(crate) use lt_word::LtWordGadgetLegacy as LtWordGadget;
```
ChihChengLiang: Messy re-export here. But it helps reduce diffs.

zkevm-circuits/src/evm_circuit/util/math_gadget/abs_word.rs
```diff
@@ -33,8 +34,10 @@ impl<F: Field> AbsWordGadget<F> {
         let x = cb.query_word32();
         let x_abs = cb.query_word32();
         let sum = cb.query_word32();
-        let (x_lo, x_hi) = x.to_word().to_lo_hi();
-        let (x_abs_lo, x_abs_hi) = x_abs.to_word().to_lo_hi();
+        let binding = x.to_word();
+        let (x_lo, x_hi) = binding.to_lo_hi();
+        let binding = x_abs.to_word();
+        let (x_abs_lo, x_abs_hi) = binding.to_lo_hi();
```
ChihChengLiang, line 40: I'm keep getting this rust compiler complains for me to add these `binding` variables.

```
temporary value dropped while borrowed
consider using a `let` binding to create a longer lived value 
```

zkevm-circuits/src/lib.rs
```diff
@@ -16,6 +16,8 @@
 #![deny(missing_docs)]
 #![deny(unsafe_code)]
 #![deny(clippy::debug_assert_with_mut_call)]
+#![allow(deprecated)]
+#![allow(unreachable_code)]
```
ChihChengLiang, line 20: Suppressed deprecated code so they don't create noises in the cargo build.

unreachable_code was caused by those `todo!`s. We allow unreachable_code for now

zkevm-circuits/src/table.rs
ChihChengLiang, line 1: I reverted all the `Word` changes on the table side because they cause a lot of interdependencies. 

It would be better for circuit implementors to fix them on their pr.

The fact that we have interdependencies also indicates that we lack better table interfaces. We currently access the table struct members directly in the circuits. We should create some functions for that to decouple the table structs and the circuits.

zkevm-circuits/src/util.rs
```diff
@@ -18,7 +16,8 @@ use eth_types::{Field, ToAddress, Word};
 pub use ethers_core::types::{Address, U256};
 pub use gadgets::util::Expr;
 
-pub(crate) fn query_expression<F: Field, T>(
+/// Steal the expression from gate
+pub fn query_expression<F: Field, T>(
```
ChihChengLiang, line 20: I have no idea why we touched this file.

ChihChengLiang (commented):
I left some comments about design decisions for the reviewers.

zkevm-circuits/src/util/word.rs
```diff
@@ -48,11 +51,38 @@ impl<T: Default, const N: usize> Default for WordLimbs<T, N> {
...
+    }
+}
+
+impl<F: Field> Expr<F> for Word32Cell<F> {
+    fn expr(&self) -> Expression<F> {
+        unreachable!(
+            "This function is meat to fix the build. Remove once we fixed the word lo-hi refactor"
+        )
+    }
+}
```
ChihChengLiang, line 76: These are the dirty secrets.

zkevm-circuits/src/evm_circuit/util/math_gadget/abs_word.rs
ed255: I see that both `to_word` returns an owned type, but `to_lo_hi` returns a borrowed type, so the result of `to_word` needs to be alive for the result of `to_lo_hi` to be available.

I think an issue here is the naming.  `to_*` methods should not return a borrowed type (for that case there's `as_*` methods).  Possible solutions are:
- Rename `to_lo_hi` to `as_lo_hi` to make the borrowed result clear, and then perhaps use some clone helper methods so that we can do `x.to_word().as_lo_hi().cloned()`
- Implement `to_lo_hi` to return an owned result.

zkevm-circuits/src/table.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
     evm_circuit::util::rlc,
     exp_circuit::param::{OFFSET_INCREMENT, ROWS_PER_STEP},
     impl_expr,
-    util::{build_tx_log_address, word, Challenges},
+    util::{build_tx_log_address, Challenges},
```
ed255, line 8: The changes in this file revert all the words lo/hi in tables.  Each table in general is used in at least 2 circuits, so for circuits word lo/hi refactors, if people stick to a single circuit there will be compilation failures (because one circuit expects a table with words lo/hi and the other doesn't).  How should we proceed with this?

I think one option is to duplicate all the table types (and associated methods) and for each table have `TableLegacy` and `Table`.  Or `Table` and `TableNew`.  One with word_rlc the other with word lo/hi.  Then the refactors can be done incrementally.

ed255 (commented):
Overall looks good! My main comment and discussion is how to proceed with the refactors and how they affect the shared tables.

hero78119 (approved):
Overall looks good to me! Thanks for the fix to make build pass.

For evm circuit part, im ok with any hack under `zkevm-circuits/evm_circuit/execution and utils`, I can remove all workaround in evm_circuit word lo/hi mr.


A concern same as @ed255: this pr revert the change for shared table word-lo-hi refactor. It make current branch build pass, but later any other circuit pr will definitely change shared table, then the branch pr build still failed. So, I also wondering how to deal with this issue, if we stick to current plan.  I believe to solve this issue, we need to separate circuits into different `cargo.toml`, then each pr patch individually.

Besides, we can introduce a rust build features to separate 2 different shared tables implementation by having mutual inclusive build tag, e.g. `--features "word-lo-hi"`. Although this doesn't mean to solve the mentioned issue above, but good thing is, at least can avoid each developer create their own version of shared table so in the end we dont know which version we should stick to.

.github/workflows/ProverBenchFromHalo2.yml
```diff
@@ -1,45 +0,0 @@
-name: Prover Bench on halo2 PR
```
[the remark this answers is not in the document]
ChihChengLiang: Yes, the plan is to bring it back before merging to main. Let me add TODOs.

.github/workflows/lints.yml
```diff
@@ -63,4 +63,4 @@ jobs:
         with:
           name: Clippy
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --all-targets -- -D warnings
```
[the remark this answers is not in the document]
ChihChengLiang: The goal of this PR is to build only the library.
Tests are not fixed yet. Tests need some extensive fixes. With the main library able to be built, tests can be fixed in parallel.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
ChihChengLiang: Make sense. Let me remove this N-type, and we can decide later what makes sense in future pr.

ChihChengLiang: Addressed in [2bc396f](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1435/commits/2bc396f1fabe2039378711494681c54309bb9590)

.github/workflows/ProverBenchFromHalo2.yml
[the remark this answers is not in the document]
ChihChengLiang: Addressed in [f9e8f0d](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1435/commits/f9e8f0d571e7a1ea2ec9b1909ec04aa09233aa74)

zkevm-circuits/src/evm_circuit/util/math_gadget/abs_word.rs
ChihChengLiang: I went with option 2, "Implement to_lo_hi to return an owned result."

[31be90a](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1435/commits/31be90a97d6d528cb7e00014d2869ea5b11e0314)

zkevm-circuits/src/table.rs
```diff
@@ -132,7 +132,10 @@ pub struct TxTable {
     /// Index for Tag = CallData
     pub index: Column<Advice>,
     /// Value
-    pub value: word::Word<Column<Advice>>,
+    pub value_word: word::Word<Column<Advice>>,
+    /// Value
+    #[deprecated(note = "value_word is fav")]
+    pub value: Column<Advice>,
```
ChihChengLiang, line 138: @ed255 @hero78119 

Inspired by @ed255's idea of duplicating the struct, I figured out that we only need to duplicate the fields. We can maintain the legacy fields and work on everything on top of the new fields. We now revert almost no existing progress by @hero78119 and still keep the compatibility.

ed255 (approved):
LGTM!  Duplicating the word fields in tables is really neat!

ed255:
While reviewing this PR I took notes of the things that need to be cleaned once we merge word-lo-hi to main:

- Rename `foo_new` to `foo`, `FooNew` to `Foo`:
	- Delete `Foo::construct`, rename `Foo::construct_new` to `Foo::construct` for
	  `AddWordsGadget`, `MulAddWordsGadget`, `TransferWithGasFeeGadget`,
	  `MemoryAddressGadget`, `ModGadget`, 
	- `WordNew` import alias to `Word`
	- `original_word_new` to `original_word`
- Rename `foo_word` methods in ConstraintBuilder to `foo`:
	- `stack_pop_word`
	- `stack_push_word`
	- `account_read_word`
	- `account_write_word`
	- `tx_context_lookup_word`
	- `block_lookup_word`
	- `bytecode_lookup_word`
	- `bytecode_length_word`
	- `account_storage_access_list_read_word`
	- `tx_refund_write_word`
	- `account_storage_read_word`
	- `account_storage_write_word`
	- `stack_lookup_word`
	- `keccak_table_lookup_word`
	- Rename `tx_context_as_word32` to `tx_context_as_word`
- Remove `_word` from `AbsWord::{x_word, x_abs_word}`
- Remove `_word` from `memory_gadget expr_word`
- Remove anything that contains "legacy"
	- `MulAddWords512Gadget::legacy_construct`
- In `zkevm-circuits/src/table.rs` check all the fields in table called
  `foo_word` and remove the legacy ones (`foo`), rename `foo_word` to `foo.

ChihChengLiang:
@ed255 That's an excellent summary! Thank you @ed255 and @hero78119 for the review.

PR #1437, opened 2023-05-29
fix docker compose

wangtsiao:
### Description

Update integration-tests/run.sh docker_compose_cmd() function to use `docker-compose` instead of `docker compose`.

### Issue Link

[link](https://github.com/docker/compose/issues/8630)

When using `./run.sh --sudo`, I encountered the following error message.
```
docker: 'compose' is not a docker command.
See 'docker --help'
```

### Type of change

### DIFF
--- a/integration-tests/run.sh
+++ b/integration-tests/run.sh
@@ -76,9 +76,9 @@ done
 
 docker_compose_cmd() {
     if [ -n "$ARG_SUDO" ]; then
-        sudo docker compose $@
+        sudo docker-compose $@
     else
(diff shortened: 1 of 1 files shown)


ChihChengLiang:
@wangtsiao The `docker compose`, with a space, is a valid command and worked well in the CI environment. An older docker version might not have this command. You might need to check your local dependency.

wangtsiao:
@ChihChengLiang Thank you for your reply, I found that indeed I installed the old version docker.

PR #1440, opened 2023-05-30
Remove keccak256 crate and migrate keccak plain to eth-types

ChihChengLiang:
### Description

- We remove the keccak256 crate.
- We migrate the keccak256 plain implementation to the eth-types crate.

As we migrate keccak circuit to the zkevm-crate, most of the functions left in keccak256 crate are unused except of the plain keccak hash implementation.

Most of the crates in the codebase depend on keccak256 crate because they use the plain keccak hash. The plain keccak hash is only 300 lines big. It would be beneficial to keep a plain implementation for people to audit the optimized keccak circuit.

### Type of change


ed255 (approved):
LGTM! What a clean up!

PR #1445, opened 2023-05-31
Dynamic `CircuitsParameters`

### LINKED ISSUE #954: Unify parametrization of circuit sizes
See https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/650#issuecomment-1319441372

Our current circuits can be parametrized with a maximum number of operations supported:
- TxCircuit: max number of txs
- CopyCircuit: max number of copied bytes
- StateCircuit: max number of rw entries
- ExpCircuit: ?
- EvmCircuit: max number of rows (Evm Steps take a dynamic number of rows)
- MPTCircuit: max number of MPT proofs?
- PiCircuit: max number of txs + max calldata bytes
- BytecodeCircuit: max number of bytecode bytes

Sometimes these numbers don't correspond to the number of rows enabled (because an operation takes several rows)

For unit tests we'd like the number to be chosen at runtime based on the test input.
- For this, the bytecode circuit is using `Option<usize>` to specify the number of rows, and when set to `None`, it calculates the number of rows based on the input
- For other circuits it's difficult to decide this value based on the input because there are dependencies (Evm depends on max_txs and max_rws), but maybe it can be done

Tasks:
- Explore if we could set all the parameters as `Option<usize>` so that they can be optionally set, or calculated based on inputs
- Unify the meaning of the size parameter
    - I think what makes more sense is setting the number of operations that the circuit supports.  Note that a check should be done to make sure that the given number is possible to achieve based on the `k` value and the `unusable_rows`.
- Put all these parameters in `CircuitParams`, so that they can all be defined in a single place.

davidnevadoc:
### Description

The main goal of this PR is to allow for the dynamic selection of sub-circuit parameters (`CirctuisParams`) based on the block provided as input.  After this change there are 2 options for selecting circuit parameters:
  1. Having them fixed before witness generation
  2. Leaving them unset as `DynamicCP` and be dynamically generated together with the witness.

### Issue Link

Close #954

### Type of change

### Contents

 - Addition of trait `CircuitsParams`, implemented by the `ConcreteCP` (old `CircuitsParams`) and the new `DynamicCP`
 - Addition of generics in all the structs that previously contained `CircutisParams`: `Block`, `CircuitInputBuilder, `CircuitInputStateRef`
 - Modify unit tests to use dynamic parameters
 - Modify variadic test to use default fixed parameters instead of dynamic parameters

### How Has This Been Tested?

With the existing tests.

### Potential improvements
Sometimes it may be difficult to track which structures contain the circuit parameters. For clarity, we could remove  `CircuitsParams` from `Block` and create a new structure containing both, something like `BlockWithParams` or `CircuitInputs`. (suggested by @ed255 )
<hr>

Big thanks to @ed255 for providing the ideas and design on how to solve this issue !!:partying_face:

### DIFF
--- a/zkevm-circuits/src/exp_circuit.rs
+++ b/zkevm-circuits/src/exp_circuit.rs
@@ -291,11 +291,12 @@ impl<F: Field> ExpCircuitConfig<F> {
         &self,
         layouter: &mut impl Layouter<F>,
         exp_events: &[ExpEvent],
-        max_exp_steps: usize,
+        max_exp_rows: usize,
     ) -> Result<(), Error> {
-        let max_exp_rows = max_exp_steps * OFFSET_INCREMENT;
+        let min_n_rows = Self::min_num_rows(exp_events);
+        dbg!(max_exp_rows, min_n_rows);
         debug_assert!(
-            Self::min_num_rows(exp_events) <= max_exp_rows,
+            min_n_rows <= max_exp_rows,
             "insufficient rows to populate the exponentiation trace"
         );
 
@@ -484,7 +485,8 @@ pub struct ExpCircuit<F> {
 
 impl<F: Field> ExpCircuit<F> {
     /// Return a new ExpCircuit
-    pub fn new(exp_events: Vec<ExpEvent>, max_exp_rows: usize) -> Self {
+    pub fn new(exp_events: Vec<ExpEvent>, max_exp_steps: usize) -> Self {
+        let max_exp_rows = max_exp_steps * OFFSET_INCREMENT + UNUSABLE_EXP_ROWS;
         Self {
             exp_events,
             max_exp_rows,
--- a/zkevm-circuits/src/exp_circuit/test.rs
+++ b/zkevm-circuits/src/exp_circuit/test.rs
@@ -51,29 +51,37 @@ fn gen_code_multiple(args: Vec<(Word, Word)>) -> Bytecode {
     code
 }
 
-fn gen_data(code: Bytecode) -> CircuitInputBuilder<FixedCParams> {
+fn gen_data(code: Bytecode, default_params: bool) -> CircuitInputBuilder<FixedCParams> {
     let test_ctx = TestContext::<2, 1>::simple_ctx_with_bytecode(code).unwrap();
     let block: GethData = test_ctx.into();
     // Needs default parameters for variadic size test
-    let mut builder =
-        BlockData::new_from_geth_data_with_params(block.clone(), FixedCParams::default())
-            .new_circuit_input_builder();
-    builder
-        .handle_block(&block.eth_block, &block.geth_traces)
-        .unwrap();
+    let builder = if default_params {
+        let mut builder =
+            BlockData::new_from_geth_data_with_params(block.clone(), FixedCParams::default())
+                .new_circuit_input_builder();
+        builder
+            .handle_block(&block.eth_block, &block.geth_traces)
+            .unwrap();
+        builder
+    } else {
+        let builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        builder
+            .handle_block(&block.eth_block, &block.geth_traces)
+            .unwrap()
+    };
     builder
 }
 
 fn test_ok(base: Word, exponent: Word, k: Option<u32>) {
     let code = gen_code_single(base, exponent);
-    let builder = gen_data(code);
+    let builder = gen_data(code, false);
     let block = block_convert::<Fr>(&builder).unwrap();
     test_exp_circuit(k.unwrap_or(18), block);
 }
 
 fn test_ok_multiple(args: Vec<(Word, Word)>) {
     let code = gen_code_multiple(args);
-    let builder = gen_data(code);
+    let builder = gen_data(code, false);
     let block = block_convert::<Fr>(&builder).unwrap();
     test_exp_circuit(20, block);
 }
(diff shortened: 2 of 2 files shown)


bus-mapping/src/circuit_input_builder.rs
```diff
@@ -255,54 +310,75 @@ impl<'a> CircuitInputBuilder {
...
+            let max_copy_rows = self
+                .block
+                .copy_events
+                .iter()
+                .fold(0, |acc, c| acc + c.bytes.len())
+                * 2
+                + 2;
+            let max_rws: usize = self.block_ctx.rwc.into();
+            let max_evm_rows = 0;
+            let max_keccak_rows = 0;
```
ed255, line 358: I guess this is using the fact that the EVM and Keccak circuit support the parameter 0, and then they perform assignments up to the required number of rows for the given witness.

But maybe this would be more clear if the calculation of these values is done here.  This way when using the dynamic parameters, we can just print the generated parameters and see the values for each circuit at once.

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -255,54 +310,75 @@ impl<'a> CircuitInputBuilder {
...
+            ConcreteCP {
+                // TODO Re-check RW padding warning
```
ed255, line 360: I think the RwTable requires some extra rows (I think it may be just 1).  Padding rows in the RwTable have the Tag=Start, and we do 2 lookups to Start rows from EVM Circuit to guarantee the total number of rw entries.  I think both lookups could point to the same row, that's why I think it's just 1, but maybe I forgot something.

On the other hand... with this https://github.com/privacy-scaling-explorations/halo2/pull/185 we could find a better approach. But that's a topic for another task!

bus-mapping/src/evm/opcodes/caller.rs
```diff
@@ -12,8 +12,8 @@ use eth_types::GethExecStep;
 pub(crate) struct Caller;
 
 impl Opcode for Caller {
-    fn gen_associated_ops(
-        state: &mut CircuitInputStateRef,
+    fn gen_associated_ops<C: CircuitsParams>(
+        state: &mut CircuitInputStateRef<C>,
```
ed255, line 16: I think it would be great if we can remove this generic from `CircuitInputStateRef`.

bus-mapping/src/evm/opcodes/callop.rs
```diff
@@ -722,7 +722,7 @@ mod tests {
 
             let mut builder = BlockData::new_from_geth_data_with_params(
                 block.clone(),
-                CircuitsParams {
+                ConcreteCP {
```
ed255, line 725: Now that we have the dynamic generation of params, we should replace all unit tests that use default params / custom params, by dynamic calculation of params!

bus-mapping/src/evm/opcodes/chainid.rs
```diff
@@ -35,8 +35,8 @@ mod chainid_tests {
         .unwrap()
         .into();
 
-        let mut builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
-        builder
+        let builder = BlockData::new_from_geth_data(block.clone()).new_circuit_input_builder();
+        let builder = builder
```
ed255, line 39: I've seen this change in other files, but I think it doesn't have any effect.

bus-mapping/src/lib.rs
```diff
@@ -220,6 +220,7 @@
 //#![deny(unsafe_code)] Allowed now until we find a
 // better way to handle downcasting from Operation into it's variants.
 #![allow(clippy::upper_case_acronyms)] // Too pedantic
+#![feature(type_changing_struct_update)]
```
ed255, line 223: We should always be careful when enabling unstable features.  Do you think this is really necessary?

ed255 (commented):
First pass review.  Overall looks good!  The only topic I would like to discuss is the possibility to find an approach where `CircuitInputBuilder` and `CircuitInputStateRef` don't have the `C: CircuitsParams` generic.  This causees that all `gen_associated_ops` are generic over `C`, but they don't have any code that depends on `C`.  So I think it would be better if the generic appears at a later stage, when the paths bifurcate.

Currently:
- `CircuitInputStateRef` needs generic C because it contains `&'a mut Block<C>`
- `CircuitInputBuilder` needs generic C because it contains `&'a mut Block<C>`
- `Block` needs generic C because it contains `C: CircuitsParams>`

The params are used in the bus-mapping at the `CircuitInputBuilder::handle_block`.  In particular, they are required after processing each tx, before setting the end block steps.  It would be nice to find a way such that no generic `C` appears before the params are needed.  I'll think more about this!

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -39,7 +39,7 @@ pub use transaction::{Transaction, TransactionContext};
 
 /// Circuit Setup Parameters
 #[derive(Debug, Clone, Copy)]
-pub struct CircuitsParams {
+pub struct ConcreteCP {
```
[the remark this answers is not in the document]
miha-stopar: `Fixed` and `Computed`?

miha-stopar (approved):
LGTM!

bus-mapping/src/circuit_input_builder.rs
ed255: @davidnevadoc Could you take a look at this?  This will be my last request for this PR :P

ed255 (approved):
LGTM! Great work!

bus-mapping/src/circuit_input_builder.rs
ed255: I forgot to reply this!  So in the current merged state with `max_rws: max_rws + 1` there's a warning right?  If that's the case, could you open an issue to track this?  You can copy a link to this conversation in the issue :)

PR #1449, opened 2023-06-01
Fix RLC is_lt_128

naure:
### Description

In the RLC gadget, the indicator `is_lt_128` must be equivalent to `value < 128`.

However, currently only the `true` case is handled. This PR handles both `true` and `false` cases.

### Issue Link

Scroll internal.

### Type of change


zkevm-circuits/src/evm_circuit/util/math_gadget/rlp.rs
```diff
@@ -67,10 +72,19 @@ impl<F: Field> RlpU64Gadget<F> {
...
-        });
+        // If is_lt_128, then value < 128, checked by a lookup.
+
+        // Otherwise, then value >= 128, checked as follows:
+        // - Either the first byte is not the most significant, and there is a more significant one;
+        // - Or the first byte is the most significant, and it is >= 128. value ∈ [128, 256) (value
+        //   - 128) ∈ [0, 128)
+        let byte_128 = value_rlc.cells[0].expr() - 128.expr();
+        let is_first = is_most_significant_byte[0].expr();
+        cb.require_boolean("is_first", is_first.clone());
```
han0110, line 83: This seems unnecessary since all `is_most_significant_byte` are already bool when they are queried.
```suggestion
```

han0110 (approved):
LGTM! Just a small nitpick.

ed255 (approved):
LGTM!  I agree on the redundant constraint pointed out by Han.

PR #1452, opened 2023-06-05
Fix #1451, option 1: Simply upgrade ethers

### LINKED ISSUE #1451: Upgrade Ethers to 2.0.7 and boundary control

We currently use the 0.17.0 version of ethers-rs crate, which is old and contains unmaintained parity dependencies.

Upgrading to the latest version 2.0.7 can let us use the latest update like withdrawal root in the block.

In addition to upgrading packages directly, It is also desirable to build a boundary for third-party, so that our codebase doesn't depend on them directly. If some new features we want and some incompatibility we don't want are both released in a new version, we can build a wrapper for it and control the behavior of the third-party package. I'm not sure if boundary control is really desirable or what degree of it is optimal, so let me present 3 prs for us to choose.

### Additional context

_No response_

ChihChengLiang:
### Description

Change ethers-rs related packages from 0.17.0 to 2.0.7

### Issue Link

#1451

### Type of change

### Rationale

We need nothing more than simply upgrading the package.

### DIFF
--- a/integration-tests/src/bin/gen_blockchain_data.rs
+++ b/integration-tests/src/bin/gen_blockchain_data.rs
@@ -75,6 +75,8 @@ async fn main() {
 
     // Compile contracts
     info!("Compiling contracts...");
+    let solc = Solc::default();
+    info!("Solc version {}", solc.version().expect("version works"));
     let mut contracts = HashMap::new();
     for (name, contract_path) in CONTRACTS {
         let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
@@ -89,7 +91,7 @@ async fn main() {
             .clone()
             .evm_version(EvmVersion::London);
 
-        let compiled = Solc::default()
+        let compiled = solc
             .compile(&input)
             .unwrap_or_else(|_| panic!("solc compile error {:?}", path_sol));
         if !compiled.errors.is_empty() {
--- a/integration-tests/src/bin/gen_blockchain_data.rs
+++ b/integration-tests/src/bin/gen_blockchain_data.rs
@@ -81,9 +81,8 @@ async fn main() {
     for (name, contract_path) in CONTRACTS {
         let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
         let inputs = CompilerInput::new(&path_sol).expect("Compile success");
-        // ethers-solc defaults evm version to Shanghai
-        // We explicitly specify EvmVersion here for some old Solidity version to work.
-        // We can change to Shanghai later.
+        // ethers-solc: explicitly indicate the EvmVersion that corresponds to the zkevm circuit's
+        // supported Upgrade, e.g. `London/Shanghai/...` specifications.
         let input = inputs
             .clone()
             .first_mut()
--- a/integration-tests/src/bin/gen_blockchain_data.rs
+++ b/integration-tests/src/bin/gen_blockchain_data.rs
@@ -79,6 +79,9 @@ async fn main() {
     for (name, contract_path) in CONTRACTS {
         let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
         let inputs = CompilerInput::new(&path_sol).expect("Compile success");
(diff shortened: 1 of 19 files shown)


ChihChengLiang:
@adria0 I'm also excited about alloy, but I feel it's too soon to introduce alloy. The reasonings are as follows.

- Alloy was released few days ago, the API might be unstable. Our codebase depends on ethers-rs heavily and when some bug is introduced by the unstable API it might be hard for us to fix.
- We use providers and signers crates from ethers-rs. In Alloy these APIs seem unimplemented or redesigned. Migrating to Alloy requires some work. I suggest we wait a bit and see if we can introduce Alloy in the near future.
- There are two main purposes to upgrade ethers-rs.
  - First, we can remove the unmaintained Parity dependencies.
  - Second, we can get the latest update like withdrawal root in the block.

Let me know if you have thoughts on this.

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -78,8 +78,16 @@ async fn main() {
     let mut contracts = HashMap::new();
     for (name, contract_path) in CONTRACTS {
         let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
+        let inputs = CompilerInput::new(&path_sol).expect("Compile success");
+        let input = inputs
+            .clone()
+            .first_mut()
+            .expect("first exists")
+            .clone()
+            .evm_version(EvmVersion::London);
```
KimiWu123, line 87: Could you have some comments to explain why we don't need to set evm version before but we need it now? Or we just have to set evm version here bcs it's part of rules of 2.0.7? Just would like to know what kind of situations we might need to upgrade the version in the future.

KimiWu123 (approved):
LGTM. Just have a question.

integration-tests/src/bin/gen_blockchain_data.rs
ChihChengLiang: Good point that we should add comments here. I added the comment in [e118683](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1452/commits/e1186830fd1571c43a227a11548d0fd2c4dd1a07)

I checked ethers.rs and found it defaulted [evm version to Shanghai](https://github.com/gakonst/ethers-rs/blob/7b7c62327303866319a8b9ea84de517eef70dc09/ethers-solc/src/artifacts/mod.rs#L756), which is incompatible with my local `0.8.16` version.


I printed out the CI solidity [version](https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/5473937281/jobs/9968062166?pr=1452#step:10:96), which is `0.8.19`. The Solidity [doc](https://docs.soliditylang.org/en/v0.8.19/using-the-compiler.html) for `0.8.19` says Paris is the latest supported EVM version. So if the ethers-solc selects Shanghai as EVM version, the Solidity compiler would not recognize it.

hero78119 (approved):
With a comment decoration suggestions.
Other LGTM!

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -75,11 +75,24 @@ async fn main() {
...
     info!("Compiling contracts...");
+    let solc = Solc::default();
+    info!("Solc version {}", solc.version().expect("version works"));
     let mut contracts = HashMap::new();
     for (name, contract_path) in CONTRACTS {
         let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
-        let compiled = Solc::default()
-            .compile_source(&path_sol)
+        let inputs = CompilerInput::new(&path_sol).expect("Compile success");
+        // ethers-solc defaults evm version to Shanghai
```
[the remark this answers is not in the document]
ChihChengLiang: Simplying the comment sounds good. I applied the suggestion in [548c876](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1452/commits/548c876b41b9d8bcc980cb86946ca76725dda44f)

PR #1453, opened 2023-06-05
Fix #1451, option 2: Full boundary control

### LINKED ISSUE #1451: Upgrade Ethers to 2.0.7 and boundary control

ChihChengLiang:
### Description

We upgraded ethers packages and move all of them inside a third-party crate in eth-types.

### Issue Link

#1451

### Type of change

### Rationale

See #1451


ChihChengLiang:
close in favor of #1452

PR #1456, opened 2023-06-05
fix: modify assign function of LtChip to take `Value<F>` input instead of `F`

enricobottazzi:
### Description

The PR modifies the assign function of the LtChip to take `Value<F>` input instead of `F`. According to this discussion [Halo2 - Discord](https://discord.com/channels/995022112811135148/1113747239534346250/1113854493646397621) with @therealyingtong and @ed255, it emerged how the current design of the chip makes its usage prone to bugs. 

For example, this is how it is currently used in summa-solvency => https://github.com/summa-dev/summa-solvency/blob/a7d0c1f51c7e146bac2686c432c945ff6f3f063f/src/chips/merkle_sum_tree.rs#L338-L354

In order to extract `F` from an assigned cell, I need to create an if conditional that performs the `assign` function only if a witness value is Some. This is a problem as the assignment function won't be called during keygen.

### Type of change

### How Has This Been Tested?

`cargo test less_than`

### DIFF
--- a/gadgets/src/less_than.rs
+++ b/gadgets/src/less_than.rs
@@ -19,8 +19,8 @@ pub trait LtInstruction<F: Field> {
         &self,
         region: &mut Region<'_, F>,
         offset: usize,
-        lhs: F,
-        rhs: F,
+        lhs: Value<F>,
+        rhs: Value<F>,
     ) -> Result<(), Error>;
 
     /// Load the u8 lookup table.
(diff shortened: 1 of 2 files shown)


ChihChengLiang (approved):
LGTM. Using `Value<F>` makes a lot of sense. Thank you @enricobottazzi for the contribution!

ed255 (approved):
LGTM!  Thanks for the contribution :D !

Issue #1461, opened 2023-06-07
TestContext sets tx nonces incorrectly (sometimes)


ed255:
The TestContext is used to setup an Ethereum state, run txs on it with go-ethereum and get the resulting traces.  When using various transactions, the TestContext automatically assigns a default nonce to them, in sequence starting from 0 (and these values can then be overwritten by the user).  This only works assuming all txs come from the same address and this address has nonce 0 at the beginning: 
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/06cce9f01d252bf9b1e84cebc22ac982a12ddc99/mock/src/test_ctx.rs#L135-L146

This automatic assignment of nonces doesn't work for example when you have 2 txs from different accounts (each with `nonce = 0`).  I think the automatic assignment should only be implemented it if covers all cases, because otherwise the developer may take some time to realize what's failing: When having 2 txs from different accounts the developer may assume that each tx.nonce is 0 (because that's the default value of the `MockTransaction`), but they will find that the second tx is sent with `nonce = 1`.

So I think we should either remove this automatic (sometimes incorrect) nonce assignment, or alternatively do it right (by inspecting the from address, and checking the nonces in the setup `accounts`).

zemse:
Thanks for marking this as good first issue. I am interested in taking this up.

PR #1464, opened 2023-06-08
Assign correct nonce in TestContext

zemse:
### Description

This PR changes the way TestContext sets tx nonces. Before nonce is set serially. Applying this PR causes the tx to use nonce from the accounts config.

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1461

### Type of change

### Contents

- Assigns correct nonce just after `func_tx` is called.
- Adds a test case to ensure that the nonces are assigned as expected.

### Rationale

The automatic assignment only works if all txs use same account and its account nonce is zero. Otherwise dev has to manually override the nonce. Inferring the nonce from account state adds convenience such that it automatically works for same accounts as well as different accounts.

### How Has This Been Tested?

A test case has been added.


ed255 (commented):
The nonce assignment code looks great, thanks for working on this!  Thanks as well for adding a test :)

I've added a comment related to how the `func_tx` is now called in a different order and the associated behavior change, please take a look.

mock/src/test_ctx.rs
```diff
@@ -132,22 +131,26 @@ impl<const NACC: usize, const NTX: usize> TestContext<NACC, NTX> {
...
 
         let mut transactions = vec![MockTransaction::default(); NTX];
-        // By default, set the TxIndex and the Nonce values of the multiple transactions
-        // of the context correlative so that any Ok test passes by default.
-        // If the user decides to override these values, they'll then be set to whatever
-        // inputs were provided by the user.
-        transactions
-            .iter_mut()
-            .enumerate()
-            .skip(1)
```
zemse, line 142: There is a `skip(1)` here. The rationale was not clear to me and in the updated implementation the first element is not skipped. But please let me know if it is necessary to skip.

mock/src/test_ctx.rs
```diff
@@ -142,12 +142,25 @@ impl<const NACC: usize, const NTX: usize> TestContext<NACC, NTX> {
...
             .for_each(|(idx, tx)| {
                 let idx = u64::try_from(idx).expect("Unexpected idx conversion error");
-                tx.transaction_idx(idx).nonce(idx);
+                tx.transaction_idx(idx);
             });
         let tx_refs = transactions.iter_mut().collect();
 
         // Build Tx modifiers.
         func_tx(tx_refs, accounts.clone());
+
```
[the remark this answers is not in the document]
zemse: - Since using invalid nonce is not useful, I've just made the functions `pub(crate)`. 
- The tx_idx logic is moved to after `func_tx`. 
- Also rebased my branch over the latest main.

mock/src/test_ctx.rs
ed255: I guess it was skipping the first tx because it had a default id, which for an integer is already 0.  But I think the code looks more clear without the skip!

ed255 (approved):
LGTM!  Thanks a lot for your contribution!
We need one more review and we can merge this :)

PR #1481, opened 2023-06-19
Refactor: move `Sha3CodeGen` struct and `rand_bytes` function to mock crate

silathdiir:
### Description

1. Move `Sha3CodeGen` and `rand_bytes` from bus-mapping to mock crate. Since they are only used for test.

2. Delete `rand` dependency for bus-mapping.

### Issue Link

`cargo build` fails in `bus-mapping` folder.

### Type of change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,7 +410,6 @@ dependencies = [
  "log",
  "mock",
  "pretty_assertions",
- "rand",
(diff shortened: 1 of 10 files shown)


ChihChengLiang (approved):
LGTM. Thanks for the cleanup

Issue #1474, opened 2023-06-15
Inspect rw lookup index mismatch error related to contract creation


hero78119:
### What command(s) is the bug in?

RUST_LOG=DEBUG RUST_BACKTRACE=1 cargo test --features test --features warn-unimplemented --package zkevm-circuits --lib -- evm_circuit::execution::begin_tx::test::begin_tx_deploy_nonce_2bytes

### Describe the bug

Example unittest pass, but once enable RUST_LOG=DEBUG there will be error message.

In bus mapping, once its contract creation, there are pushing memory read op https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/bus-mapping/src/evm/opcodes/return_revert.rs#L52-L60 but in circuit return_revert.rs  there is no respective memory lookup operation, so the lookup by index check is failed.

Within the fix, we also should check whether it also imply potiential soundness or not.

PR #1482, opened 2023-06-19
Feat/#1474 rw lookup index error

KimiWu123:
### Description

address #1474 and also found #1483 when investigating this issue.

### Issue Link

#1474 

### Type of change

### Contents

- The root cause is that there are memory operations in block.get_rws() but there is only single `copy lookup` expression  while copy table lookup happens in opcode execution. Which makes the indexes of `assigned_rw_values` and `step.bus_mapping_instance` are inconsistent. 
- Add a new function `get_rws_for_check_rw_lookup` specific for `check_rw_lookup`. This function could bypass memory operations if any copy table lookup happens, and it makes the indexes of `assigned_rw_values` and `step.bus_mapping_instance` consistent.

### Rationale

- Add a new function instead of modifying `get_rws()` is because trying to minimize the impact to our codebase since this is a check function.


bus-mapping/src/evm/opcodes/create.rs
```diff
@@ -251,7 +251,6 @@ impl<const IS_CREATE2: bool> Opcode for Create<IS_CREATE2> {
                     (CallContextField::IsStatic, false.to_word()),
                     (CallContextField::IsCreate, true.to_word()),
                     (CallContextField::CodeHash, code_hash.to_word()),
-                    (CallContextField::Value, callee.value),
```
hero78119, line 254: Does it means this lookup is not using by circuit originally?

hero78119 (commented):
Just notice it's in draft, but already review few so the comments is published first, will have another round once it's ready

bus-mapping/src/evm/opcodes/create.rs
KimiWu123: yes, we don't have this constraint this in our circuit.

ChihChengLiang (approved):
LGTM. I can confirm that the command in #1474 no longer logs errors. 
Great work fixing the logs. It must take you a lot of effort to track this bug.

We should do a systematic refactor for check_rw_lookup to make the logging code easier to maintain.

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -1381,23 +1393,40 @@ impl<F: Field> ExecutionConfig<F> {
...
+
         // Check that the number of rw operations generated from the bus-mapping
         // correspond to the number of assigned rw lookups by the EVM Circuit
-        // plus the number of rw lookups done by the copy circuit.
-        if step.rw_indices_len() != assigned_rw_values.len() + step.copy_rw_counter_delta as usize {
+        // plus the number of rw lookups done by the copy circuit
+        // minus the number of copy lookup event.
+        if step.rw_indices_len()
+            != assigned_rw_values.len() + step.copy_rw_counter_delta as usize
+                - copy_lookup_included as usize
```
hero78119, line 1415: Here `copy_lookup_included` looks pretty special: it's a boolean flag and will only toggle from false -> true when suffer first copy event. Just wanna double confirm if there are more than 1 copy event does the idx calculation still correct~?

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -1411,14 +1440,46 @@ impl<F: Field> ExecutionConfig<F> {
...
             );
+
             // In the EVM Circuit, reversion rw lookups are assigned after their
             // corresponding rw lookup, but in the bus-mapping they are
             // generated at the end of the step.
             let idx = if is_rev {
                 step.rw_indices_len() - rev_count
             } else {
-                idx - rev_count
+                idx - rev_count + offset - copy_lookup_processed as usize
```
hero78119, line 1450: Here copy_lookup_included looks pretty special: it's a boolean flag and will only toggle from false -> true when suffer first copy event. Just wanna double confirm if there are more than 1 copy event does the idx calculation still correct~?

zkevm-circuits/src/evm_circuit/execution/create.rs
```diff
@@ -310,8 +310,7 @@ impl<F: Field, const IS_CREATE2: bool, const S: ExecutionState> ExecutionGadget<
                     }
 
                     cb.require_step_state_transition(StepStateTransition {
-                        // +1: move to new context
```
hero78119, line 313: Just question asking : why original got `1.expr()` and remove here. I guess it's because the removal of `(CallContextField::Value, callee.value)`  ?

zkevm-circuits/src/evm_circuit/execution.rs
KimiWu123: yes, it only supports 1 copy event and will add a `TODO` and warning message as we discussed.
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1482#issuecomment-1600378826

zkevm-circuits/src/evm_circuit/execution/create.rs
KimiWu123: Yes, it's bsc of the removal of `(CallContextField::Value, callee.value)`. I misunderstood the `+1` in the original pr.

KimiWu123:
Refactor for `check_rw_lookup` created, #1489, cc @ChihChengLiang and @hero78119

hero78119 (approved):
LGTM!

Issue #1484, opened 2023-06-19
BufferReaderGadget underconstrained


naure:
### What command(s) is the bug in?

EVM circuit, CALLDATALOAD opcode.

### Describe the bug

There is a BufferReaderGadget which must indicate whether a buffer index is in a range `[start; end)`, and pad with zeros otherwise.  ([Code here](https://github.com/scroll-tech/zkevm-circuits/blob/79caaf4339ab033eea33e4df4b997f5e3d3f1a32/zkevm-circuits/src/evm_circuit/util/memory_gadget.rs#L609)). However, this seems to be underconstrained.

There are two mechanisms that enforces the zero padding, one based on `selectors`, and one based on `bound_dist`. The `selectors` have no constraints relating them to the input bounds. This is unlike the [python code](https://github.com/scroll-tech/zkevm-specs/blob/0f97a3e3843234dd1dcaf35069313f740793ba45/src/zkevm_specs/evm_circuit/util/memory_gadget.py#L10) where selectors are derived from the `bytes_left`, which does not exist in the rust code. Meanwhile, this functionality is not used since the selector witnesses are always true.

The vulnerability: the gadget is used [here](https://github.com/scroll-tech/zkevm-circuits/blob/79caaf4339ab033eea33e4df4b997f5e3d3f1a32/zkevm-circuits/src/evm_circuit/execution/calldataload.rs#L138) for CALLDATALOAD. The value of `read_flag` can be made 0 by setting the selectors to 0, regardless of the addresses. This disables the calldata lookup and reads a 0 byte instead, as an illegal choice of the prover.

See the fix PR #1485.

PR #1485, opened 2023-06-19
fix-bufferreader: constrain the "in-bound" selectors vs the addresses

naure:
### Description

In the BufferReaderGadget, there are two conflicting mechanisms that enforces the zero padding, one based on `selectors`, and one based on `bound_dist`. This causes the issue #1484. This PR fixes this and improves performance at the same time.

### Issue Link

#1484 

### Type of change

### Contents

- Remove the columns `bound_dist` and everything related to it.
- Repurpose the unused `selectors` and add constraints to it.
- The buffer length is taken as the sum of selectors, and compared to the length implied by the start and end addresses.
- The cases where `start < end` and `end <= start` are supported explicitly.
- The length is capped between 0 and the max buffer size using the existing MinMaxGadget. A single MinMaxGadget is used thanks to a `select`.
- Remove the unused methods `has_data` and `num_bytes`.

### Rationale

The simplest fix would be to remove the `selectors` and keep the `bound_dist` mechanism. However, this PR does a more efficient way. Indeed, the existing logic on the selectors is simple and cheap, while the other is costly because it compares every index to the addresses.

### How Has This Been Tested?

`cargo test --package zkevm-circuits --features test -- evm_circuit::execution::calldataload`

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/calldataload.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/calldataload.rs
@@ -261,14 +261,8 @@ impl<F: Field> ExecutionGadget<F> for CallDataLoadGadget<F> {
             }
         }
 
-        self.buffer_reader.assign(
-            region,
-            offset,
-            src_addr,
-            src_addr_end,
-            &calldata_bytes,
-            &[true; N_BYTES_WORD],
-        )?;
+        self.buffer_reader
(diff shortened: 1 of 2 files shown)


roynalnaruto (approved):
Looks good to me

ChihChengLiang (approved):
LGTM. Thanks for the fix.

PR #1492, opened 2023-06-22
MPT new descriptors & benchmark

miha-stopar:
### Description

New descriptors are used which have `Vec<Node>` structure and thus the conversion in `witness_row.rs` from `[][]byte` to `Vec<Node>` was removed. Note that `keccak_data` is now part of the `Node` struct - each node holds the byte streams that need to be hashed.

Additionally, an MPT benchmark was added (some structs needed to be made public).

### Type of change

- [x] This change requires a documentation update

### Contents

- All tests updated.
- `prepare_witness` function removed from `witness_row.rs`.
- MPT test updated to not use `prepare_witness`.
- Benchmark added.


circuit-benchmarks/src/mpt_circuit.rs
```diff
@@ -0,0 +1,110 @@
...
+        };
+
+        // Initialize the polynomial commitment parameters
+        let mut rng = XorShiftRng::from_seed([
+            0x59, 0x62, 0xbe, 0x5d, 0x76, 0x3d, 0x31, 0x8d, 0x17, 0xdb, 0x37, 0x32, 0x54, 0x06,
+            0xbc, 0xe5,
+        ]);
+
+        // Bench setup generation
+        let setup_message = format!("Setup generation with degree = {}", degree);
```
Brechtpd, line 61: Nitpick: the other circuits use `SETUP_PREFIX` etc... for some of these strings.

circuit-benchmarks/src/mpt_circuit.rs
```diff
@@ -0,0 +1,110 @@
...
+        let start2 = start_timer!(|| proof_message);
+
+        create_proof::<
+            KZGCommitmentScheme<Bn256>,
+            ProverSHPLONK<'_, Bn256>,
+            Challenge255<G1Affine>,
+            XorShiftRng,
+            Blake2bWrite<Vec<u8>, G1Affine, Challenge255<G1Affine>>,
+            MPTCircuit<Fr>,
+        >(&general_params, &pk, &[circuit], &[], rng, &mut transcript)
```
Brechtpd, line 84:
```suggestion
        >(&general_params, &pk, &[circuit], &[&[]], rng, &mut transcript)
```

I was running the benchmark and the prover time was too fast (especially because some optimizations are not yet on this branch)!

I think it's related to the input here for the instances. If I for example do the same for the keccak circuit it also gets proven much faster. I looked shortly at why in halo2 and it seems like the instance length shouldn't be 0 or some prover work is skipped (though I could be wrong about that). The fact that this doesn't match the input expected by the MockProver input is pretty confusing.

However, if I change it as suggested I get a `ConstraintSystemFailure`... Not sure what's up yet.

circuit-benchmarks/src/mpt_circuit.rs
```diff
@@ -0,0 +1,110 @@
...
+            KZGCommitmentScheme<Bn256>,
+            VerifierSHPLONK<'_, Bn256>,
+            Challenge255<G1Affine>,
+            Blake2bRead<&[u8], G1Affine, Challenge255<G1Affine>>,
+            SingleStrategy<'_, Bn256>,
+        >(
+            &verifier_params,
+            pk.get_vk(),
+            strategy,
+            &[],
```
Brechtpd, line 104:
```suggestion
            &[&[]],
```

Probably.

zkevm-circuits/src/mpt_circuit/witness_row.rs
```diff
@@ -154,6 +158,7 @@ pub(crate) enum MptWitnessRowType {
     StorageNonExisting = 19,
 }
 
+/// MPT witness row
```
Brechtpd, line 161: I think `MptWitnessRowType` and `MptWitnessRow` can also completely be removed no?

Brechtpd (commented):
Thanks a lot for doing the witness refactor!

I think there's something up with the benchmark, but otherwise looks great.

Brechtpd (approved):

PR #1503, opened 2023-06-30
Skip invalid txs due to nonce, intrinsic gas and insufficient funds

xiaodino:
### Description

Support skipping the following invalid transactions:
- nonce too low
- nonce too high
- intrinsic gas too low
- insufficient funds for gas * price + value

### Issue Link

https://github.com/taikoxyz/zkevm-circuits/issues/70

### Type of change

### Contents
- Check if a TX is invalid by adding a few auxiliary columns to the TX table.
- The related PR in [zkevm-specs](https://github.com/privacy-scaling-explorations/zkevm-specs) is https://github.com/privacy-scaling-explorations/zkevm-specs/pull/338


eth-types/src/evm_types.rs
```diff
@@ -75,6 +75,10 @@ impl GasCost {
     pub const CALL_WITH_VALUE: u64 = 9000;
     /// Constant cost for turning empty account into non-empty account
     pub const NEW_ACCOUNT: u64 = 25000;
+    /// Gas cost of warming up an account with the access list
+    pub const ACCESS_LIST_ADDRESS: u64 = 2400;
+    /// Gas cost of warming up a storage with the access list
+    pub const ACCESS_LIST_STORAGE: u64 = 1900;
```
hero78119, line 81: Can't find in this PR for where uses this 2 variables

hero78119 (commented):
Hi, firstly thanks for the contribution!

1st review pass and raise few questions below

- Only find `TxInvalid` tag in read lookup, but no place to write to TxTable. The write supposed to be happened in pi circuit ?
- Also Only find sth like `when `TxInvalid` == true/false, then what are the derived consequence, but seems not find the place to constraint the pre-condition which cause `TxInvalid` = true/false, should we constraint in in tx circuit?

Other concern & suggested action items:
- https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/geth-utils/src/lib.rs#L34C43-L34C43 `result.starts_with("Failed")` can be removed as right now all tx will be included
- fix existing failed unittest.
- more unittests to cover block with txs `invalid: true`
- block hash computation logic need to take noted as need to excluded TxInvalid = false transaction.

Is this PR ready for review or still WIP ~?

zkevm-circuits/src/table/tx_table.rs
```diff
@@ -183,6 +187,14 @@ impl TxTable {
                             TxContextFieldTag::CallDataGasCost,
                             word::Word::from(tx.call_data_gas_cost()),
                         ),
+                        (
+                            TxContextFieldTag::TxInvalid,
+                            word::Word::from(tx.is_create()),
```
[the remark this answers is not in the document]
xiaodino: it's caused by merge conflicts which I'm fixing now.

eth-types/src/evm_types.rs
xiaodino: Good catch. I will remove them.

ChihChengLiang:
@xiaodino note that the unit tests are still failing. Let us know when the PR is ready for review.

xiaodino:
> @xiaodino note that the unit tests are still failing. Let us know when the PR is ready for review.

Hi @ChihChengLiang All unit tests pass. Could you take a look? Thanks.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -357,7 +442,11 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
 
         // 3. Call to account with empty code.
         cb.condition(
-            and::expr([not::expr(tx_is_create.expr()), no_callee_code.clone()]),
+            and::expr([
```
hero78119, line 445: I think there is mismatch with specs https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/evm_circuit/execution/begin_tx.py#L200
condition

zkevm-circuits/src/pi_circuit.rs
```diff
@@ -1509,6 +1509,11 @@ impl<F: Field> SubCircuit<F> for PiCircuit<F> {
                                     .collect_vec(),
                             ),
                             (TxFieldTag::IsCreate, tx.is_create.to_le_bytes().to_vec()),
+                            (TxFieldTag::TxInvalid, tx.invalid_tx.to_le_bytes().to_vec()),
+                            (
+                                TxFieldTag::AccessListGasCost,
+                                tx.access_list_gas_cost.to_le_bytes().to_vec(),
```
hero78119, line 1515: Looks like `access_list_gas_cost` is under-constraints and prover can fill any value, because this pseudo field is not in normal specs so its validity is not assure.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -712,24 +888,46 @@ mod test {
...
+    #[test]
+    fn begin_tx_gadget_simple_enable_skipping_invalid_tx() {
+        begin_tx_gadget_simple(true);
+    }
+
+    #[test]
+    fn begin_tx_gadget_simple_disable_skipping_invalid_tx() {
+        begin_tx_gadget_simple(false);
+    }
+
```
hero78119, line 930: I think above two test can grouped into single test with true/false because both of them doen't include any invalid transaction

Also below some test seems can be group as well

hero78119 (changes requested):
Thanks for the update!

Just went through 2nd reviews:

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -10,7 +10,7 @@ use std::collections::HashMap;
 // Step dimension
 pub(crate) const STEP_WIDTH: usize = 128;
 /// Step height
-pub const MAX_STEP_HEIGHT: usize = 19;
+pub const MAX_STEP_HEIGHT: usize = 24;
```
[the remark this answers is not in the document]
xiaodino: It's not from resolving conflict. We have to increase MAX_STEP_HEIGHT from 19 to 24. Otherwise, it returns 
```
running 21 tests
thread 'evm_circuit::execution::begin_tx::test::begin_tx_deploy_nonce_6bytes' panicked at 'CMFixedWidthStrategy: max_height reached (19)', zkevm-circuits/src/util/cell_placement_strategy.rs:134:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7f94b314cead7059a71a265a8b64905ef2511796/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7f94b314cead7059a71a265a8b64905ef2511796/library/core/src/panicking.rs:67:14
   2: <zkevm_circuits::util::cell_placement_strategy::CMFixedWidthStrategy as zkevm_circuits::util::cell_manager::CellPlacementStrategy>::place_cell
             at ./src/util/cell_placement_strategy.rs:134:13
   3: zkevm_circuits::util::cell_manager::CellManager<S>::query_cell
             at ./src/util/cell_manager.rs:324:25
   4: zkevm_circuits::util::cell_manager::CellManager<S>::query_cells::{{closure}}
             at ./src/util/cell_manager.rs:360:22
```

external-tracer/src/lib.rs
```diff
@@ -68,6 +68,21 @@ pub fn trace(config: &TraceConfig) -> Result<Vec<GethExecTrace>, Error> {
         },
     )?;
 
-    let trace = serde_json::from_str(&trace_string).map_err(Error::SerdeError)?;
+    let trace: Vec<GethExecTrace> =
+        serde_json::from_str(&trace_string).map_err(Error::SerdeError)?;
+    // Don't throw only for specific invalid transactions we support.
+    println!("Creates a trace for the specified config");
```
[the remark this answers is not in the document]
xiaodino: Removed.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
xiaodino: Updated.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -346,11 +354,12 @@ impl<F: Field, const N_ADDENDS: usize, const INCREASE: bool>
 /// unconditionally if must_create is true.  This gadget is used in BeginTx.
 #[derive(Clone, Debug)]
 pub(crate) struct TransferWithGasFeeGadget<F> {
-    sender_sub_fee: UpdateBalanceGadget<F, 2, false>,
-    sender_sub_value: UpdateBalanceGadget<F, 2, false>,
-    receiver: UpdateBalanceGadget<F, 2, true>,
+    pub sender_sub_fee: UpdateBalanceGadget<F, 2, false>,
+    pub sender_sub_value: UpdateBalanceGadget<F, 2, false>,
+    pub receiver: UpdateBalanceGadget<F, 2, true>,
```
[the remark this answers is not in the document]
xiaodino: Updated. Thanks.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
xiaodino: Grouped those tests.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -225,17 +248,79 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
         });
 
         // Transfer value from caller to callee, creating account if necessary.
+        // For invalid transactions we do not do any transfers
```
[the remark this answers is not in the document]
xiaodino: Updated. Used `require_equal_word` and avoided `Word::from_lo_unchecked(word.lo())`

bus-mapping/src/circuit_input_builder/tracer_tests.rs
```diff
@@ -37,13 +37,14 @@ impl CircuitInputBuilderTx {
         let block = crate::mock::BlockData::new_from_geth_data(geth_data.clone());
         let mut builder = block.new_circuit_input_builder();
         let tx = builder
-            .new_tx(0, &block.eth_block.transactions[0], true)
+            .new_tx(0, &block.eth_block.transactions[0], true, false)
```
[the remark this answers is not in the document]
xiaodino: Add a new field in Mock transaction class https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1503/files#diff-a85105b516c1c726c249026c0ad2796499304c5a185aa49ba24e874efb3c9fa5R141 and add unit tests for `is_success = true and is_invalid == true` https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1503/files#diff-8f5984af7a71d8fdfb2862e9655564673740a1d72df468ede0f01dbbd3ffba98R971

geth-utils/gethutil/trace.go
```diff
@@ -235,15 +236,23 @@ func Trace(config TraceConfig) ([]*ExecutionResult, error) {
 
 		result, err := core.ApplyMessage(evm, &message, new(core.GasPool).AddGas(message.GasLimit))
 		if err != nil {
```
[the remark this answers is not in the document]
xiaodino: The implementation for whitelist and enumerating what kind of errors are invalid transactions is here https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1503/files#diff-9c5c9eb105d122a2c4ddaabcb69d3aabd161c2b0ab7dcb7528ae29b3abd71c2cR74

zkevm-circuits/src/pi_circuit.rs
xiaodino: @hero78119  Could you provide more details for "access_list_gas_cost is under-constraints and prover can fill any value" and suggest a change?

xiaodino:
> Thanks for the update!
> 
> Just went through 2nd reviews:

Hi @hero78119  I have addressed your comments. Please take a look. Many Thanks.

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -357,7 +442,7 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
 
         // 3. Call to account with empty code.
         cb.condition(
-            and::expr([not::expr(tx_is_create.expr()), no_callee_code.clone()]),
+            and::expr([no_callee_code.clone(), tx_is_invalid.expr()]),
```
KimiWu123, line 445: I don't quite understand this change. The original one is `not create & no_callee_code` but it becomes `invalid tx & no_callee_code`. Why doest it become `invalid tx` and if it's for invalid tx case, is the tx still persistent [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1503/files#diff-8f5984af7a71d8fdfb2862e9655564673740a1d72df468ede0f01dbbd3ffba98R446-R451)?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -576,17 +701,59 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
             call.rw_counter_end_of_reversion,
             call.is_persistent,
         )?;
-        self.sufficient_gas_left
-            .assign(region, offset, F::from(tx.gas() - step.gas_cost))?;
+
+        let init_code_gas_cost = if tx.is_create() {
+            // Calculate gas cost of init code for EIP-3860.
+            (tx.call_data.len() as u64 + 31) / 32 * eth_types::evm_types::INIT_CODE_WORD_GAS
```
KimiWu123, line 707: The length of calldata, you could get from the return value of `self.tx_call_data_word_length.assign(...)`

KimiWu123: a nitpick, you could import `INIT_CODE_WORD_GAS` in L31, `use eth_types::{....}`, then the code here looks cleaner

KimiWu123 (commented):
@xiaodino  wondering if you have considered to move invalid tx case to single gadget, just like our error_xxx for opcode error states?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
xiaodino: Good point. Updated.

xiaodino:
> @xiaodino wondering if you have considered to move invalid tx case to single gadget, just like our error_xxx for opcode error states?

We have considered to move invalid tx case to a single gadget. The main reason we choose the current implement is to support skip invalid tx and keep other valid tx pass.

zkevm-circuits/src/pi_circuit.rs
hero78119: Because the logic to define `access_list_gas_cost` is not constrainted in plonk circuit. 

Here attach an example of [init_code_gas_cost](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs#L153-L158) which is constrainted in circuit as constant. It seems to me that `access_list_gas_cost` can remove from transaction field/table and just derived it in begin_tx circuit logic?

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
xiaodino: If it's invalid tx case, the tx won't be persistent and will be skipped if enable_skipping_invalid_tx is set to True.

KimiWu123: But the logic here is if it's a invalid tx, then do the following checks which include `require(is_persistent, 1)` as L449.

zkevm-circuits/src/evm_circuit/param.rs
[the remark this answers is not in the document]
Brechtpd: Each execution state only uses the amount of rows necessary, so increasing the max here will only use more rows for execution states that actually need them, in this case just `begin_tx`. This won't have any significant impact on what can be proven because a tx already costs at a minimum 21k gas, so `begin_tx` is already "expensive". There may be some extra rotations, but I don't think that will have a significant impact on performance either.

Not sure if 24 is indeed still needed because I think that value may have been found before the hi/lo split, so it could be that a lower value is now possible because words could take up fewer cells in some cases.

ChihChengLiang:
Hi @xiaodino,
Do we plan to close this PR if we're to move to another approach?

xiaodino:
@ChihChengLiang Yes. Closed the PR. I will open a new one for another approach.

KimiWu123:
@xiaodino , created a issue https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1611 to follow up. 
cc @ChihChengLiang

Issue #1499, opened 2023-06-29
[word lo/hi] investigate circuit degree issue


hero78119:
### What command(s) is the bug in?

super circuit degree unittest

### Describe the bug

https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1496

hero78119:
Few issues found in pr are unresolved yet https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1508

Also relevant to https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1401

PR #1508, opened 2023-07-03
[word-lo-hi] fine tune copy circuit to reduce degree

hero78119:
### Description

[_PR description_]

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1499

### Type of change

### Contents
This pr is to rewrite/split copy circuit constrains to trade more gates with less degree.
copy table `tag` field is binary number decomposition with degree N (N represent number of bits, in copy table case N = 3). Once put in `or::expr([#num_or_condition])`, gate degrees will be exploded quickly by `N*#num_or_condition`.  

Before PR, copy circuit max is degree 12, 
After PR,  copy circuit max 6 degree with 2 more gates. 

However, circuit degree upper bound still increase by 1 (9 -> 10) for word-lo-hi refactor, due to state-circuit `batch_is_zero-"is_zero is 1 if values are all zero">`  gate degrees from 6 -> 10 Due to we fold 2 extra column, each contribute 2 degrees
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/word-lo-hi/zkevm-circuits/src/state_circuit.rs#L129-L141.

State circuit need some discussion, as I thought rewrite it might be a bit over-engineering.

### DIFF
--- a/zkevm-circuits/src/copy_circuit.rs
+++ b/zkevm-circuits/src/copy_circuit.rs
@@ -419,14 +419,33 @@ impl<F: Field> SubCircuitConfig<F> for CopyCircuitConfig<F> {
             .collect()
         });
 
-        meta.create_gate("id_hi === 0 when Momory, TxLog, and TxCalldata", |meta| {
+        meta.create_gate("id_hi === 0 when Momory", |meta| {
             let mut cb = BaseConstraintBuilder::default();
 
-            let cond = or::expr([
-                tag.value_equals(CopyDataType::Memory, Rotation::cur())(meta),
-                tag.value_equals(CopyDataType::TxLog, Rotation::cur())(meta),
-                tag.value_equals(CopyDataType::TxCalldata, Rotation::cur())(meta),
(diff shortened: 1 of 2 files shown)


hero78119:
+ @adria0, might have thought on state-circuit case

ed255 (approved):
This solution looks good to me!

ChihChengLiang (approved):
LGTM

hero78119:
Will merge it first to unblock word-lo-hi back to main. state circuit case can fine tune later

hero78119:
> > `+` @adria0, might have thought on state-circuit case
> 
> Nice improvment! What do you have in mind (overengineered?) for the 4-batched zero check in state circuit?

Hi @adria0 my option is th quick fix is separating into 4 zero check then batch it. However it implies `BatchedIsZeroChip` usage easily to lead to degree booming pitfall.  Maybe we need to remove or redesign `BatchedIsZeroChip`

PR #1525, opened 2023-07-06
Rm deadcode of EVM circuit

ChihChengLiang:
### Description

- Removed #[allow(dead_code)]  and `#![allow(unused_imports)]`.
- Removed most of the dead code. Kept a few with the "_" initial name.
- Reorganize dev modules and feature flags for "zkevm-circuits" crate. 
   - ~~"test" flag is only used when test utils are shared within "zkevm-circuits"~~ "test" flag is gone.
   - "test-util" flag is used when utils are shared with testool crate.
   - "test-circuits" flag is reserved for test circuits that are shared with integration tests.

### Type of change

### Rationale

These are the reason I keep a method:

- The method is a sensible API and is tested. Although it is unused but it makes sense to keep it. The `remainder` method of the Constant Division Gadget is such an example.
- Variants in Enums. We use the `match` macro for them so it is important we don't miss any variant. It makes sense to keep them even when we don't use them.
- Work in progress and recent works. Cell manager and  _error_depth are examples.

I remove a method or a gadget for these reasons.

- If it has been there for a long time but never used.
- Default removing them to question us hard if the gadget really sparks joy. batched_is_zero and binary number gadgets are such cases.

### How Has This Been Tested?

```
cargo build --all-features
cargo test // just for build, without running the full tests
```


ed255 (commented):
Nice clean up!  I think it's great to remove unused code to keep the codebase clean :)

At first when reviewing the PR I was a bit puzzled by all the `_` prefixes added, then I realized that this is the way Rust stops complaining about unused stuff.  I would suggest a different solution where possible for these two cases:
- Methods/Fields/Enums that are only used in testing, instead of prefixing them with underscore, annotate them with `#[cfg(test)]`.  This applies to some getters, debug functions, etc.
- Enums that have unused options:  annotate just the enum with `allow_dead_code`.  I believe this will look cleaner.

I think we should be more tolerant with allowing testing dead code than non-testing dead code.  For example, I think it's ok to have a testing enum with options that are unused (like the StateCircuit enum to overwrite assignments).  But then if an enum is used in non-testing setting and has a non-used case, perhaps it's better to remove the case (like the EVM Circuit `StepLast`).

What do you think about this?

ed255:
For future reference and indexing purposes (so that we can search via github), here's a list of deleted EVM math gadgets in this PR:
- BatchedIsZeroGadget
- BinaryNumberGadget
- MinMaxWordGadget

ChihChengLiang (commented):
> I would suggest a different solution where possible for these two cases:
>
> - Methods/Fields/Enums that are only used in testing, instead of prefixing them with underscore, annotate them with #[cfg(test)]. This applies to some getters, debug functions, etc.
> - Enums that have unused options: annotate just the enum with allow_dead_code. I believe this will look cleaner.
>
> I think we should be more tolerant with allowing testing dead code than non-testing dead code. For example, I think it's ok to have a testing enum with options that are unused (like the StateCircuit enum to overwrite assignments). But then if an enum is used in non-testing setting and has a non-used case, perhaps it's better to remove the case (like the EVM Circuit StepLast).

There are reasons I like the underscore a lot more than `#[cfg(test)].` and `allow(dead_code)`. Generally speaking,

- underscore is a less verbose way and a very visible way to tell devs and compilers that a variable is unused. 
- `#[cfg(test)]` comes with annoying side effects if a method depends on some imports, we have to mark that imports with `#[cfg(test)]`. For example, the `overrides` property in StateCircuit depends on `HashMap`, which requires us to mark `#[cfg(test)]` for `use std::collections::HashMap;`. The code would look noisy in this way. Generally, I see the scattering of `#[cfg(test)]` in the codebase as an anti-pattern. The desired way should be confining all test dependencies inside a module and we only mark the `mod testdeps` with the `#[cfg(test)]`. 
- "allow(dead_code)" is bad because it is overkill. When we mark an enum as dead_code, we don't know if the variants are dead code or if the whole enum is dead code. Again I believe like the cfg test, the allow(dead_code) should be only applied to the module level. An unstable module that is in active development is a suitable place for allow(dead_code).

zkevm-circuits/src/state_circuit/multiple_precision_integer.rs
```diff
@@ -112,21 +112,14 @@ pub struct Chip<F: Field, T, const N_LIMBS: usize, const N_VALUES: usize>
 where
     T: ToLimbs<N_LIMBS>,
 {
-    config: Config<T, N_LIMBS>,
+    _config: Config<T, N_LIMBS>,
```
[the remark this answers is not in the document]
ChihChengLiang: On second thought, I'm inclined not to touch `_config` in this PR. 

The State circuit modules shared many duplicated functionalities with existing constraint builders and word gadgets. A rewrite of the State circuit in existing architecture in a future pr is a cleaner way to go. Working them partially in this PR is out of the scope and difficult to track the changes.

zkevm-circuits/src/keccak_circuit/table.rs
```diff
@@ -167,15 +167,11 @@ mod tests {
     fn chi_table() {
         // Check the base pattern for all combinations of bits.
         for i in 0..16_usize {
-            let (a, b, c, d) = (i & 1, (i >> 1) & 1, (i >> 2) & 1, (i >> 3) & 1);
+            let (a, b, c) = (i & 1, (i >> 1) & 1, (i >> 2) & 1);
```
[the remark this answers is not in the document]
ed255: Thanks for the explanation!

zkevm-circuits/src/state_circuit/multiple_precision_integer.rs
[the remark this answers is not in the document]
ed255: Makes sense!

ed255:
> There are reasons I like the underscore a lot more than `#[cfg(test)].` and `allow(dead_code)`. Generally speaking,
> 
>     * underscore is a less verbose way and a very visible way to tell devs and compilers that a variable is unused.
> 
>     * `#[cfg(test)]` comes with annoying side effects if a method depends on some imports, we have to mark that imports with `#[cfg(test)]`. For example, the `overrides` property in StateCircuit depends on `HashMap`, which requires us to mark `#[cfg(test)]` for `use std::collections::HashMap;`. The code would look noisy in this way. Generally, I see the scattering of `#[cfg(test)]` in the codebase as an anti-pattern. The desired way should be confining all test dependencies inside a module and we only mark the `mod testdeps` with the `#[cfg(test)]`.
> 
>     * "allow(dead_code)" is bad because it is overkill. When we mark an enum as dead_code, we don't know if the variants are dead code or if the whole enum is dead code. Again I believe like the cfg test, the allow(dead_code) should be only applied to the module level. An unstable module that is in active development is a suitable place for allow(dead_code).

Thanks for the detailed reasoning!  I didn't think about the messy imports coming from having `#[cfg(test)]` on things that are not modules; the code gets very messy to manage.

I'm now convinced about this approach :)

ed255 (approved):
LGTM! Thanks for taking care of this clean up :D

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -311,8 +311,8 @@ pub struct ExecutionConfig<F> {
     error_oog_code_store: Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorOutOfGasCodeStore }>>,
     error_invalid_jump: Box<ErrorInvalidJumpGadget<F>>,
     error_invalid_opcode: Box<ErrorInvalidOpcodeGadget<F>>,
-    error_depth: Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorDepth }>>,
-    error_contract_address_collision:
+    _error_depth: Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorDepth }>>,
```
hero78119, line 314: For example, instead of add underscore prefix, which somehow still make reader confused about this, what do you think we have `deadcode` specificly on this field with explicit `reason`. e.g.

```rust
    #[allow(dead_code, reason = "under active development")]
    error_depth: Box<DummyGadget<F, 0, 0, { ExecutionState::ErrorDepth }>>,
``` 

It both caring about readibility and clean up real unused code for now.

zkevm-circuits/src/util/cell_manager_strategy.rs
```diff
@@ -17,7 +17,7 @@ impl CMFixedWidthStrategyDistribution {
         }
     }
 
-    pub(crate) fn get(&self, cell_type: CellType) -> Option<&Vec<Column<Advice>>> {
+    pub(crate) fn _get(&self, cell_type: CellType) -> Option<&Vec<Column<Advice>>> {
```
hero78119, line 20: If we prefix with underscore, reader might still wondering the reason.

So I suggest we turn to have explicitly `allow(dead_code, reason = "xxx")` in this method.

```rust
#[allow(dead_code, reason = "this method will be used in XXX")]
pub(crate) fn get(&self, cell_type: CellType) -> Option<&Vec<Column<Advice>>> {
```

zkevm-circuits/src/evm_circuit/util/math_gadget/constant_division.rs
```diff
@@ -133,7 +122,7 @@ mod tests {
 
             cb.require_equal(
                 "correct reminder",
```
hero78119, line 124: Just spike a typo "remainder" :)

zkevm-circuits/src/evm_circuit/util/math_gadget/constant_division.rs
```diff
@@ -60,7 +60,7 @@ impl<F: Field, const N_BYTES: usize> ConstantDivisionGadget<F, N_BYTES> {
         self.quotient.expr()
     }
 
-    pub(crate) fn remainder(&self) -> Expression<F> {
+    pub(crate) fn _remainder(&self) -> Expression<F> {
```
hero78119, line 63: Think it will be better by

```rust
#[allow(dead_code, reason = "reserve method for extract remainder expression")]
pub(crate) fn remainder(&self) -> Expression<F>
```

hero78119 (commented):
It's good we cleanup quite lot of unused methods. 
However, I think there are better interpretation for methods left for now comparing with just purely prefix with "_"

We can add `#[allow(dead_code, reason = "XXXX")]` specificly on field/method to explain why we allow it's there. I randomly select few typical case in comments.

Add-on, to use `reason=""` in allow linter, we just need to add 
`#![feature(lint_reasons)]`
in [src/lib.rs](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/lib.rs#L21)

wdyt~?

zkevm-circuits/src/evm_circuit/util/math_gadget/constant_division.rs
ChihChengLiang: Nice catch, fixed in [26b3527](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1525/commits/26b3527c028c204a758ab14222ff02e04b6b91d5)

ChihChengLiang:
Hi @hero78119, please see my previous reasoning with Edu about why I refrain from using `allow(dead_code)`

https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1525#pullrequestreview-1518913830

I like the idea of having the lint reasons feature `#![feature(lint_reasons)]`. But if we removed all `allow(dead_code)`, then we don't need it in this PR.

ChihChengLiang:
@hero78119 Thanks for expanding on the reasoning.

For public or public(crate) methods, I think `#[allow(dead_code, reason = "useful debug function")]` makes sense. It clarifies why the method is unused more than the underscore way.

I'm still not convinced the same can be applied to Enum's variants. For the `the variants are dead code or if the whole enum is dead code` problem, I want the compiler to determine it. The reason field only explains to humans.

hero78119 (approved):
Thanks for the review adoption on public method:) It looks really nice now 💯

Enum case is indeed overkill, and underscore prefix make more sense yes.

Thanks again for the big effort 👍

Issue #633, opened 2022-07-25
Fully implement `PUSHn`


han0110:
Currently `PUSHn` is not fully implemented, the case "push out of code range" is not taken care. It should first lookup the bytecode's length and have the data selectors summing to `min(length - program_counter, n)`.

PR #1527, opened 2023-07-11
Fully implement PUSHn

rrtoledo:
### Description
We intend here to support PUSHn with bytecode length inferior to n.

### Issue Link
Issue #633 
Linked to specs issue https://github.com/privacy-scaling-explorations/zkevm-specs/issues/463

### Type of change

### Contents

- Update push.rs to support bytecode of size smaller than n
- Update stop.rs similarly

### Rationale

In pushn, "selectors" are switches to enable or disable lookup arguments on the bytcode bytes as well as to check whether these are non-null if needs be. They directly depend on "n" of PUSHn.
We add to these "selectors", "counters" which have similar role but are a subset of them depending on the bytecode length. We added these extra switches to be able to easily check the various identities at the risk of some redundancy.

To support these changes, modifications in stop.py also had to be done. More precisely, the check "is_out_of_range had to be extended to support small bytecodes.

### How Has This Been Tested?

Extra tests have been added in push.rs (only).

### DIFF
--- a/zkevm-circuits/src/evm_circuit/execution/push.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/push.rs
@@ -41,7 +41,7 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
         let opcode = cb.query_cell();
         let code_size = cb.query_cell();
-        let code_size_left = code_size.expr() - cb.curr.state.program_counter.expr();
+        let code_size_left = code_size.expr() - cb.curr.state.program_counter.expr() - 1.expr();
 
         let value = cb.query_word32();
         // Query selectors for each opcode_lookup to know whether the current byte needs to be
@@ -65,8 +65,8 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
             LtGadget::construct(cb, code_size_left.clone(), num_pushed.clone());
 
         // Expected number of padding
-        let num_padding: halo2_proofs::plonk::Expression<F> = 32.expr() - num_pushed.clone()
-            + is_out_of_range.expr() * (num_pushed.clone() - code_size_left + 1.expr());
+        let num_padding: halo2_proofs::plonk::Expression<F> =
+            is_out_of_range.expr() * (num_pushed.clone() - code_size_left);
 
         // The pushed bytes are viewed as left-padded big-endian, but our random
         // linear combination uses little-endian, so we lookup from the LSB
@@ -86,8 +86,8 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
         let mut pa_prev = 1.expr();
         for (idx, (pu, pa)) in is_pushed.iter().zip(is_padding.iter()).enumerate() {
             let byte = &value.limbs[idx];
-            let index = cb.curr.state.program_counter.expr() + opcode.expr()
-                - (OpcodeId::PUSH1.as_u8() - 1 + idx as u8).expr();
+            let index: halo2_proofs::plonk::Expression<F> =
+                cb.curr.state.program_counter.expr() + num_pushed.clone() - idx.expr();
 
             // is_pushed can transit from 1 to 0 only once
             // as 1 - [1, 1, 1, ..., 1, 0, 0, 0]
@@ -96,13 +96,12 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
                 pu_prev - pu.expr(),
             );
 
-            // when is_pushed = 0, is_padding = 1
-            // otherwise, is_padding can transit from 1 to 0 only once
-            // as 1 - [1, 1, 0, ..., 0, 1, 1, 1] (out of bound)
-            // as 1 - [0, 0, 0, ..., 0, 1, 1, 1] (not out of bound)
+            // is_padding can transit from 1 to 0 only once
+            // as 1 - [1, 1, 0, ..., 0, 0, 0, 0] (out of bound)
+            // as 1 - [0, 0, 0, ..., 0, 0, 0, 0] (not out of bound)
             cb.require_boolean(
-                "Constrain is_padding can only transit from 1 to 0 when is_pushed = 1, else is 1",
-                pu.expr() * (pa_prev - pa.expr()) + (1.expr() - pu.expr()) * (pa.expr() - 1.expr()),
+                "Constrain is_padding can only transit from 1 to 0",
+                pa_prev - pa.expr(),
             );
 
             // byte is 0 if it is either not pushed or padding
@@ -183,7 +182,7 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
         self.code_size
             .assign(region, offset, Value::known(F::from(code_size)))?;
 
-        let code_size_left = (code_size - step.pc) as usize;
+        let code_size_left = (code_size - step.pc - 1) as usize;
         self.is_out_of_range.assign(
             region,
             offset,
@@ -195,11 +194,9 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
         self.value.assign_u256(region, offset, value)?;
 
         let is_out_of_bound = code_size_left < npushed;
-        // let mut num_padding = 32 - n;
         let mut out_of_bound_padding = 0;
         if is_out_of_bound {
-            out_of_bound_padding = npushed - code_size_left + 1;
-            // num_padding += out_of_bound_padding
+            out_of_bound_padding += npushed - code_size_left
         };
         for (i, (pu, pa)) in self
             .is_pushed
@@ -208,7 +205,7 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
             .enumerate()
         {
             let pushed = i < npushed;
-            let padding = !pushed || (is_out_of_bound && i < out_of_bound_padding);
+            let padding = i < out_of_bound_padding;
             pu.assign(region, offset, Value::known(F::from(pushed as u64)))?;
             pa.assign(region, offset, Value::known(F::from(padding as u64)))?;
         }
--- a/zkevm-circuits/src/evm_circuit/execution/push.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/push.rs
@@ -1,6 +1,7 @@
 use crate::{
     evm_circuit::{
         execution::ExecutionGadget,
+        param::N_BYTES_PROGRAM_COUNTER,
         step::ExecutionState,
         util::{
             common_gadget::SameContextGadget,
@@ -8,6 +9,7 @@ use crate::{
                 ConstrainBuilderCommon, EVMConstraintBuilder, StepStateTransition,
                 Transition::Delta,
             },
+            math_gadget::LtGadget,
             sum, CachedRegion, Cell,
         },
         witness::{Block, Call, ExecStep, Transaction},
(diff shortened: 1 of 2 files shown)


rrtoledo:
I added a commented `num_padding` in `assign_exec_step` as I thought it could be nice to see the parallel with`configure`. Do you think it's worth keeping it or that I should remove it?

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -35,10 +40,33 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
 
     fn configure(cb: &mut EVMConstraintBuilder<F>) -> Self {
         let opcode = cb.query_cell();
+        let code_size = cb.query_cell();
+        let code_size_left = code_size.expr() - cb.curr.state.program_counter.expr();
```
ed255, line 44: In the specs this is `code_length_left = code_length - instruction.curr.program_counter - 1`.  I wonder why the `-1` is not here.

Imagine the `code_size=100`.  And where at `program_counter=99` (the last byte in the bytecode).  Then `code_size_left` should be `0`.

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -35,10 +40,33 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
...
+        // Deduce the number of bytes to push.
+        let num_pushed = opcode.expr() - OpcodeId::PUSH1.as_u64().expr() + 1.expr();
+
+        // If the number of bytes that needs to be pushed is greater
+        // than the size of the bytecode left, then we are out of range.
+        let is_out_of_range: LtGadget<F, N_BYTES_PROGRAM_COUNTER> =
+            LtGadget::construct(cb, code_size_left.clone(), num_pushed.clone());
+
+        // Expected number of padding
+        let num_padding: halo2_proofs::plonk::Expression<F> = 32.expr() - num_pushed.clone()
```
ed255, line 68: In the specs this is `num_padding = is_out_of_bound * (num_pushed - code_length_left)`.  The `+1` in this PR probably comes from the difference of how `code_size_left` is calculated.

But what about the `32 - num_pushed`?

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -53,48 +81,58 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
...
+            );
+
+            // when is_pushed = 0, is_padding = 1
+            // otherwise, is_padding can transit from 1 to 0 only once
+            // as 1 - [1, 1, 0, ..., 0, 1, 1, 1] (out of bound)
+            // as 1 - [0, 0, 0, ..., 0, 1, 1, 1] (not out of bound)
             cb.require_boolean(
-                "Constrain selector can only transit from 1 to 0",
-                selector_prev - selectors[idx].expr(),
+                "Constrain is_padding can only transit from 1 to 0 when is_pushed = 1, else is 1",
```
ed255, line 104: My impression from the specs was that both `is_padding` and `is_pushed` followed the pattern `[1 1 ... 0 0]`.  But here `is_padding` can be `[ 11 0 ... 0 1 1]`.  Could you comment on this difference?

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -53,48 +81,58 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
         //                           ▼                     ▼
         //   [byte31,     ...,     byte2,     byte1,     byte0]
         //
-        for idx in 0..32 {
+        // First is_pushed (resp. is_padding) will always be 1 (resp. 1)
+        let mut pu_prev = 1.expr();
+        let mut pa_prev = 1.expr();
+        for (idx, (pu, pa)) in is_pushed.iter().zip(is_padding.iter()).enumerate() {
             let byte = &value.limbs[idx];
             let index = cb.curr.state.program_counter.expr() + opcode.expr()
```
ed255, line 89: Suggestion: we have
```rust
let num_pushed = opcode.expr() - OpcodeId::PUSH1.as_u64().expr() + 1.expr();
```
So this would be simplified like:
```rust
let index = cb.curr.state.program_counter.expr() + num_pushed - idx.expr();
```

ed255:
> I added a commented `num_padding` in `assign_exec_step` as I thought it could be nice to see the parallel with`configure`. Do you think it's worth keeping it or that I should remove it?

I didn't see this comment, maybe you didn't push it?

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: Perhaps the terminology is not the best (as well as the consistency sorry!).
Basically, I called by padding here anything that is not pushed, so that corresponds to the 32 - n bytes that are not pushed. Likewise, when we are out of bound we have `(num_pushed - code_length_left)` extra bytes that are set to 0.

The +1 comes exactly from this reason, I made a mistake before and will fix them both.

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: Sure! Does the explanation in the previous post satisfy you?

rrtoledo:
> > I added a commented `num_padding` in `assign_exec_step` as I thought it could be nice to see the parallel with`configure`. Do you think it's worth keeping it or that I should remove it?
> 
> I didn't see this comment, maybe you didn't push it?

It is line 198 and 202
```
        let is_out_of_bound = code_size_left < n;
        // let mut num_padding = 32 - n;
        let mut out_of_bound_padding = 0;
        if is_out_of_bound {
            out_of_bound_padding = n - code_size_left + 1;
            // num_padding += out_of_bound_padding
        };
```

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: I am not sure why I did not write or removed the -1, sorry about that!

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: To add a bit more:  We want to check that either the bytes that are pushed are in the bytecode table and the non pushed ones are set to 0s. Finally we check that the number of byte (potentially) pushed is indeed n (`Sum is_pushed = n`).

The number of byte we actually push is not n but `n - is_out_of_bound * (num_pushed - code_length_left)`. We could thus change the previous constraint by `Sum (is_pushed * (1 - is padding)) = n - is_out_of_bound * (num_pushed - code_length_left)`. The problem I see with such equation is that the 1) identity has a higher degree, and 2) you could still set is_pushed to 1 when is_padding = 1. Not major problems, but it's not very robust.
Hence, I decided to keep the check on "n" (not the number of bytes pushed), `Sum is_pushed = n`, as well as add an extra check that the `Sum is_padding = num_padding`.

rrtoledo: ... I see a flaw in this reasoning now when adding the `32 - n`, in `num_padding` so I will remove it!

ed255:
> It is line 198 and 202
> 
> ```
>         let is_out_of_bound = code_size_left < n;
>         // let mut num_padding = 32 - n;
>         let mut out_of_bound_padding = 0;
>         if is_out_of_bound {
>             out_of_bound_padding = n - code_size_left + 1;
>             // num_padding += out_of_bound_padding
>         };
> ```

Oh sorry, for some reason I was thinking it would be in the `bus-mapping` and I didn't see any changed file there, my bad!
I think the comment helps to follow along!  Alternatively you could remove those commented lines and in the circuit add
```rust
let out_of_bound_padding = is_out_of_range.expr() * (num_pushed.clone() - code_size_left + 1.expr());
let num_padding = 32.expr() - num_pushed.clone() + out_of_bound_padding;
```
So that the circuit explicitly matches the logic in the assignment for `out_of_bound_padding`.  But feel free to ignore this :)

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: The flaw (if anyone interested): is padding because it is the concatenation of both the `32-n` and `(num_pushed - code_length_left)` bool has the following form: 11... 000... 111  where the first 1s should represent the out of bound padding and the last the normal 32-n.  Because we do not enforce the size of each sub list of 1s, one could move some 1s from one sublist to the others, hence we would not perform some of the bytecode table lookups.

zkevm-circuits/src/evm_circuit/execution/push.rs
rrtoledo: I will chagne it back to 11000 as said previously

rrtoledo:
> > It is line 198 and 202
> > ```
> >         let is_out_of_bound = code_size_left < n;
> >         // let mut num_padding = 32 - n;
> >         let mut out_of_bound_padding = 0;
> >         if is_out_of_bound {
> >             out_of_bound_padding = n - code_size_left + 1;
> >             // num_padding += out_of_bound_padding
> >         };
> > ```
> 
> Oh sorry, for some reason I was thinking it would be in the `bus-mapping` and I didn't see any changed file there, my bad! I think the comment helps to follow along! Alternatively you could remove those commented lines and in the circuit add
> 
> ```rust
> let out_of_bound_padding = is_out_of_range.expr() * (num_pushed.clone() - code_size_left + 1.expr());
> let num_padding = 32.expr() - num_pushed.clone() + out_of_bound_padding;
> ```
> 
> So that the circuit explicitly matches the logic in the assignment for `out_of_bound_padding`. But feel free to ignore this :)

It was originally not commented but after some changes it wasn't used any longer so I had to either remove or comment it :)

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -35,10 +40,33 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
...
+        // pushed
+        let is_pushed = array_init(|_| cb.query_bool());
+
+        // Query selectors for each opcode_lookup to know whether the current byte is actually
+        // padding
+        let is_padding = array_init(|_| cb.query_bool());
+
+        // Fetch the bytecode length from the bytecode table.
+        let code_hash = cb.curr.state.code_hash.clone();
+        cb.bytecode_length(code_hash.to_word(), code_size.expr());
```
ed255, line 57: Suggestion: The spec uses `code_length` instead of `code_size`.  Since the lookup function is called `bytecode_length` I suggest updating all `code_size` to `code_length` to unify the variable names in the spec and here.

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -35,10 +40,33 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
...
+
+        // Deduce the number of bytes to push.
+        let num_pushed = opcode.expr() - OpcodeId::PUSH1.as_u64().expr() + 1.expr();
+
+        // If the number of bytes that needs to be pushed is greater
+        // than the size of the bytecode left, then we are out of range.
+        let is_out_of_range: LtGadget<F, N_BYTES_PROGRAM_COUNTER> =
```
ed255, line 64: Suggestion: Rename `is_out_of_range` to `is_out_of_bound` to match the specs.

ed255 (approved):
LGTM!  The implementation looks nice and clear!  It's really nice that you keep the code well commented :)

I've left a few suggestions regarding variable name consistency with the specs, please take a look!  Since they are minor comments I'm approving already.

zkevm-circuits/src/evm_circuit/execution/push.rs
```diff
@@ -124,23 +164,50 @@ impl<F: Field> ExecutionGadget<F> for PushGadget<F> {
...
-            selector.assign(
-                region,
-                offset,
-                Value::known(F::from((idx < num_additional_pushed as usize) as u64)),
-            )?;
+        let is_out_of_bound = code_size_left < npushed;
+        let mut out_of_bound_padding = 0;
+        if is_out_of_bound {
+            out_of_bound_padding += npushed - code_size_left
+        };
```
lightsing, line 200:
```suggestion
        let is_out_of_bound = code_size_left < npushed;
        let out_of_bound_padding = if is_out_of_bound {
            out_of_bound_padding = npushed - code_size_left
        } else {
            0
        };
```
remove mut only used by underling if stmt

lightsing (approved):
General OK, several code style fix needed

han0110 (approved):
LGTM!

PR #1528, opened 2023-07-12
Copy: replace broken LtChip with IsEqualChip

naure:
LtChip is used incorrectly, and to use it correctly would be costly. Replace it with a simple equality comparison.


lispc:
oh wrong repo and wrong base branch?

naure:
(fat finger, please ignore)

PR #1537, opened 2023-07-20
Clean up github workflows

ed255:
### Description

- Remove bitrot task which was building the benchmarks and examples but we don't have any
- Merge cargo fmt, clippy and doc links into a single job to reduce noise
- Split heavy tests and light tests into two jobs to parallelize the work
- Merge all feature tests into a single job to reduce noise.
- Disable integration tests on pull requests because they are taking a long time currently.  Enable them only when the trigger-integration-test label is set, when they are scheduled via cron, or when pushing to main.
---
- Times before: 57 minutes
  - Longest: Integration tests: 57 min
  - Second longest: Tests: 28 min
- Times now: 19 minutes
  - Longest: Light unit tests: 19 min
  - Second longest: Heavy unit tests: 16 min

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -28,7 +28,10 @@ concurrency:
 
 jobs:
   set-outputs:
-    if: github.event.pull_request.draft == false
+    if: github.event_name == 'schedule' ||
+        github.event_name == 'push' ||
+        github.event_name == 'workflow_dispatch' ||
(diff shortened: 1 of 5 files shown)


ChihChengLiang (approved):
LGTM.
The CI checks look cleaner! The testing time is also improved by not running the integration tests in every PR.

hero78119 (approved):
LGTM! Thanks for solving the painful waiting time when PR is submit

PR #1544, opened 2023-07-25
Hardcode build target in workflow

ed255:
### Description

Remove the matrix strategy in the build job and instead hardcode the build target.  Since we only have 1 target we don't need the matrix strategy.

This is a small PR aimed to clean up the output of jobs running from commits merged to main.  Our configuration skips jobs that have already run, and when this happens, the matrix strategy is not resolved and we see this:
image

By removing the matrix strategy we should see a cleaner output.


ChihChengLiang (approved):
LGTM. I think this can be merged with 1 approval.

PR #1550, opened 2023-08-03
Worst case benches - deploy wc contracts

AronisAt79:
### Description

This is the first of a PRs series aiming to implement worst case block proof bench-marking tests within zkevm-circuits/integration-tests.
The changes submitted here implement the blockchain data generation functionality* needed for the wc bencmarks.

* compilation and deployment of new SCs

### Type of change

✅ New feature (non-breaking change which adds functionality)

### Contents

~~1. wcerrors.rs : Recreate the solc::SolcError and solc::SolcIoError types with necessary methods and private functions
2. worst_case.rs: Recreate ethers::solc::Solc with necessary methods~~
3. gen_blockchain_data.rs

### Rationale

Smart contracts written for generation of blocks with increased proof computational intensity use nested opcode calls in the inline assembly block that violate serde recursion limits. ~~Modules in Contents section 1 and 2 are created to manipulate the compilation behavior and disable recursion depth limits during deserialization~~

Error handling has also been added in gen_blockchain_data.rs in order to catch solc compilation errors hidden in the CompilerOutput struct

### How Has This Been Tested?

~~https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/5749057365/job/15583140857~~
https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/5753344249/job/15596341036


integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -90,11 +90,24 @@ async fn main() {
...
             .evm_version(EvmVersion::London);
 
-        let compiled = solc
-            .compile(&input)
-            .unwrap_or_else(|_| panic!("solc compile error {:?}", path_sol));
-        if !compiled.errors.is_empty() {
-            panic!("Errors compiling {:?}:\n{:#?}", &path_sol, compiled.errors)
+        // .compile() call will either fail with Err variant or return Ok(CompilerOutput)
+        // which may contain Errors or Warnings
+        let compiled = match solc.compile(&input) {
```
ChihChengLiang, line 95: I noticed that we only need to add `disable_recursion_limit` to the Deserializer. Using the public method `compile_output` from Solc can achieve that directly.

I think we don't need the Solcwc wrapper because much of the content in the `Solcwc` is about solidity version management, which we don't need here.

I have a local build with the following change that can successfully do `cargo run --bin gen_blockchain_data` on those contracts hitting the recursion limit.

```suggestion
        let output = solc.compile_output(&input).unwrap();
        let mut deserializer: serde_json::Deserializer<serde_json::de::SliceRead<'_>> =
            serde_json::Deserializer::from_slice(&output);
        deserializer.disable_recursion_limit();
        let compiled = match CompilerOutput::deserialize(&mut deserializer) {
```

ChihChengLiang (commented):
Hi @AronisAt79, 
I ported some comments from your fork. I also added comments on the solidity wrapper.

integration-tests/src/bin/gen_blockchain_data.rs
AronisAt79: @ChihChengLiang nice! This means that we need neither wcerrors.rs nor worst_case.rs, correct? Since we directly apply the external pub fn compile_output().

integration-tests/src/worst_case.rs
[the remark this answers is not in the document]
AronisAt79: Solcwc net needed based on followin conversation

integration-tests/src/lib.rs
```diff
@@ -27,13 +27,25 @@ use url::Url;
...
+/// compiler optimizations will not allow recursive execution of targeted opcodes
+pub const WARN: &[u64] = &[2018, 5667];
 /// List of contracts as (ContractName, ContractSolidityFile)
 pub const CONTRACTS: &[(&str, &str)] = &[
     ("Greeter", "greeter/Greeter.sol"),
     (
         "OpenZeppelinERC20TestToken",
         "ERC20/OpenZeppelinERC20TestToken.sol",
     ),
+    // worst case block SCs
```
ChihChengLiang, line 45:
```suggestion
    // Contracts to test worst-case usage of opcodes.
```

ChihChengLiang (commented):
I added more comments on small details.

leolara:
@AronisAt79 don't be so hard on yourself, I don't think this is the **worse case** PR I have ever seen.

leolara:
The main function is very long, I wonder if we could split it in several functions that represent different steps in the process. I would be much easier to read.

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -90,11 +91,28 @@ async fn main() {
             .clone()
             .evm_version(EvmVersion::London);
 
-        let compiled = solc
-            .compile(&input)
-            .unwrap_or_else(|_| panic!("solc compile error {:?}", path_sol));
-        if !compiled.errors.is_empty() {
-            panic!("Errors compiling {:?}:\n{:#?}", &path_sol, compiled.errors)
+        // .compile() call will either fail with Err variant or return Ok(CompilerOutput)
+        // which may contain Errors or Warnings
```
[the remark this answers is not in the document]
AronisAt79: maybe simple change to 

```suggestion
        // compilation will either fail with Err variant or return Ok(CompilerOutput)
        // which may contain Errors or Warnings
```

to explain the usage of match block plus if condition?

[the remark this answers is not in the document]
ChihChengLiang: Sounds good.

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -211,6 +230,53 @@ async fn main() {
         (block_num.as_u64(), contract.address()),
     );
 
+    // Deploy smart contracts for worst case block benches
+    //
+
+    // CheckMload
+    let contract = deploy(
```
leolara, line 237: I think here there is a clear opportunity for DRY. There is always deploy, get_block_number, blocks.insert, deployments.insert. So I think it can be extracted into a function

AronisAt79: @leolara, correct. I was thinking about it and trying to figure out if i can extract the constructor arguments' type from the compiled contract somehow, cause it is an input to the .deploy() method, and differs depending on the .sol code. So for example, MLOAD.sol does not have a constructor, so deploy() needs () as input. Maybe work on this separately and put it in another PR to not broaden the scope so much? @ChihChengLiang ?

ChihChengLiang: It looks like the refactor some additional work. I think putting the deduplication in the next pull request is a good idea.

ChihChengLiang (approved):
LGTM after the CI issues are fixed. 
Great work fixing the compilation output parsing. 
Can you run `make fmt` to fix the formatting issue?

integration-tests/src/bin/gen_blockchain_data.rs
leolara: OK

leolara (approved):

PR #1553, opened 2023-08-04
Remove assembly in geth-utils

ChihChengLiang:
### Description

Remove geth-utils and examples.

### Type of change

### Rationale

- Assembly was introduced in #91 to build examples to use geth-util. It was not used elsewhere in the codebase.
- The examples are removed as we currently don't use geth-util like the way the examples show. (Note that the examples also fail to run now. The transactions in the examples would fail because the trace config defaults to a 0 block gas limit. It runs after adding a block gas limit.)
- I found those examples unused when reviewing how we use the tracing API. It is much easier to review the tracing code after removing the Assembly.
- The Readme mentioned a TODO CLI, which we never built. I think it might be interesting to create a good first issue for people to extend testool(https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/690b2f143a94b7474f0c99493a99b0f2c53d93b8/testool/README.md#L38) for it.


Brechtpd (approved):
Looks good! If I remember correctly, I copied these files into this project because they were part of Han's code for generating these traces that I basically just copied. I guess even at the time it was pretty unlikely this would ever get used here so I probably should not have included them in the first place, so good that these are finally removed! Thanks for cleaning things up.

ChihChengLiang:
@Brechtpd My pleasure! Thank you for elaborating on the background of the decision.

hero78119 (approved):
LGTM!

PR #1565, opened 2023-08-15
[WIP] Reenable ignored callops tests

ChihChengLiang:
### Description

The callops tests were disabled in #1120 because there was a bug that was non-trivial to solve (https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1120/commits/34f56e143121fb8556585532c5490b4b9d6e711d). 

#1132 was filed to resolve the bug, which was fixed in #1286.

The disabled callops were not enabled after that.

### Type of change

Bug fix (non-breaking change which fixes an issue)

### Test

```
cargo test -p zkevm-circuits evm_circuit::execution::callop::test::callop_simple
```

PR #1582, opened 2023-09-05
use StepRws to make rw index less error prone

ChihChengLiang:
### Description

Manually counting read-write indices is error-prone. We use the `StepRws` struct to automatically track read-write indices.

This PR is motivated by trying to fix #1565. Unfortunately, this PR alone doesn't fix #1565.

### Type of change


ed255 (approved):
LGTM!

hero78119 (approved):
LGTM!

Issue #1566, opened 2023-08-15
Port mpt-witness-generator back to zkevm-circuits


ChihChengLiang:
### Describe the feature you would like

We've merged the MPT circuit, but it has yet to be able to generate witnesses for our EVM computation. So, our next step is to add the witness generation feature in the circuit input builder, as described in #811.

But #811 was written in the context before the [mpt-witness-generator](https://github.com/privacy-scaling-explorations/mpt-witness-generator) (MPTWG hereafter) became an independent repo.

MPTWG as an external dependency would cause issues, and we should move it back to this repo as a crate.

### Additional context

Before we explain why it causes issues, let me share my understanding of the current status.

We currently test the MPT circuit by feeding a list of generated witnesses to the MPT circuit.
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/5d01150f55ae9e877fae92246d82e47f051db71e/zkevm-circuits/src/mpt_circuit.rs#L813-L826

The generated witnesses are in the form of JSON files located at https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/484c1ae8cdb16ce0d1d8a8645f29a968206bb4fe/zkevm-circuits/src/mpt_circuit/tests

We can run this command to run the MPT tests.
```
 cargo test --package zkevm-circuits --lib -- mpt_circuit::tests --nocapture 
```

The JSON witnesses are generated using MPTWG. To generate witnesses, run `cd witness` then `go test gen_witness_from_infura_blockchain_test.go prepare_witness.go leaf.go extension_node.go modified_extension_node.go nodes.go test_tools.go branch.go util.go`.

This causes some problems. Say we want to update the MPTWG. The flow would be 

1. Work on a PR at MPTWG and get it merged.
2. Work on a PR at zkevm-circuits repo and get it merged.

But for step 2, we have to carefully run the MPTWG and generate new JSON files in the zkevm-circuit PR. Note that these JSON files are [matrices of hax compressed in one line](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1563/files#diff-108a6ca57151149ece222826ca066a7e23ba9a61a431fb80b62edc52592e7e4f), which is not easy to review.

The proposed approach is we 

1. Move MPTWG as a crate in zkevm-circuits. The new flow of updating MPTWG would become one PR. 
[...]

PR #1589, opened 2023-09-08
Port MPT witness generator

ChihChengLiang:
### Description

MPT witness generator was developed [here](https://github.com/privacy-scaling-explorations/mpt-witness-generator). We now port them to here.

### Issue Link

#1566 

### Type of change

### Contents

- All repo contents from https://github.com/privacy-scaling-explorations/mpt-witness-generator

### Rationale

- We port the raw repo content here. No modification except removing `.git` folder. This way, we don't hide any modifications in the big diffs.
- We expect further changes in the following PR

### How Has This Been Tested?

No test


ed255 (approved):
LGTM following the rationale described in the PR (which is just moving the external repo to zkevm-circuits, and iterate the implementation in future PRs)

From the quick look here are some comments that could be addressed in future PRs:
1. Make it explicit which files come from go-ethereum and which ones are new.  Also mention the go-ethereum version used (or the commit hash).  For the files that come from go-ethereum, make it explicit which functions/types have been changed or introduced.
2. I noticed lines of commented code.  I think some of that comes from go-ethereum file; maybe the desire was to disable some functionality.  Comments on why that is disabled would be helpful
3. In some of the test files I noticed many functions that look very similar but have different input-ouputs.  I think there's a big opportunity of reusing code there. 
4. Are all the files and functions strictly needed?  For types and public functions that are not modified, could we instead import them from go-ethereum instead of copying them?

Regarding possible formatting of the code in future PRs, I think it would be ideal if the files imported from go-ethereum are not formatted differently than they appear in go-ethereum, so we can quickly diff both sides to see the differences without noise.

PR #1591, opened 2023-09-09
[MPTWG] Update CI for MPTWG

ChihChengLiang:
### Description

Enable the MPT witness generator CI check

### Type of change

### Contents

- Run Go build and Go format on MPTWG
- Ignore main tests CI run if no change on zkevm-circuits
- Ignore MPTWG CI run if no change on MPTWG

### Rationale

We want to enable automatic checks on MPTWG


.github/workflows/mpt-witness-generator.yml
```diff
@@ -0,0 +1,47 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.20'
+
+    - name: Format
+      working-directory: ./mpt-witness-generator
+      run: go fmt ./...
```
ed255, line 37: I think this will not report formatting issues.  Instead it will try to format all the code and return success (no matter if there were changes or not).

Maybe we can use this action, which fails when the code is not formatted https://github.com/Jerome1337/gofmt-action

ChihChengLiang: Nice catch! Fixed in [b84da75](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1591/commits/b84da758cc3d38aed68d8db47c074419518cfe0f)

ed255 (approved):
LGTM!

PR #1594, opened 2023-09-10
fix(callop): fix LastCalleeId, and other fixes

lispc:
### Description

1. for callop, when there is no bytecode (empty account or non-existed account) or pre check fails, the return data should be flushed to empty, and the LastCalleeId (used in RETURNDATACOPY opcode) should be set to correct callee id instead of 0.
2. precompile can fail. (eg: out of gas)
3. "precompile account exists" is only the current status of eth mainnet. It is not part of evm spec. Following spec, precompile account can be non-existed.
4. when ErrDepth, there is no transfer
5. (not 100% sure) I also changed the 'is_static' check for only call opcode.   It was a fix i made long ago during fixing each fail case of testool. I guess maybe we can construct this test case to check this behavior: delegatecall inside callcode(with non 0 value) inside staticcall. It is a valid tx, but the 'value' is not zero?

### Issue Link

[_link issue here_]

### Type of change


lispc:
oh i cannot see the err logs from github actions... everything ok with CI?

ed255:
> oh i cannot see the err logs from github actions... everything ok with CI?

I can't see them either in https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6135267810/job/16648876291 if this happens again we should investigate it.
On the last run I can see the full logs (all the checks passed).

zkevm-circuits/src/evm_circuit/execution/callop.rs
```diff
@@ -145,7 +145,7 @@ impl<F: Field> ExecutionGadget<F> for CallOpGadget<F> {
             );
         });
 
-        cb.condition(call_gadget.has_value.clone(), |cb| {
+        cb.condition(is_call.expr() * call_gadget.has_value.clone(), |cb| {
```
ed255, line 148: When reviewing this I tested the behavior of a CALLCODE with non-zero value in a static context with geth, and there's no error.  I guess this is because CALLCODE sends value to the caller, which is a no-op, so the state isn't changed.  I'm not sure what's the point of the value parameter in CALLCODE.  This would mean that the only callop that can change the state is CALL with non-zero value; all other combinations can be executed successfully in a static context.  Is this your interpretation as well?

ed255 (approved):
LGTM! Thanks for the various fixes!

zkevm-circuits/src/evm_circuit/execution/callop.rs
lispc: "This would mean that the only callop that can change the state is CALL with non-zero value" i also think so.

ChihChengLiang (approved):
LGTM.
Thank you @lispc for upstreaming this fix.
Thank @ed255 for investigating the CALLCODE with non-zero value.

For the CI issue https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6135267810/job/16648876291, it happened from time to time roughly after #1537. It usually works after a rerun, so we didn't investigate it.

PR #1605, opened 2023-09-14
Automatic compilation of mptwitness

adria0:
### Description

The PR adds the mpt-witness-generator into the workspace, and automatically bulding the go library via rust.

### Type of change


mpt-witness-generator/rustlib/build.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    // Link
+    println!("cargo:rustc-link-search=native={}", out_dir);
+    println!("cargo:rustc-link-lib=static={}", lib_name);
+}
+
+fn fail(message: String) {
+    let _ = writeln!(
+        io::stderr(),
+        "\n\nError while building geth-utils: {}\n\n",
```
ed255, line 43: This message should be updated with `mpt-witness-generator` :P

mpt-witness-generator/rustlib/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+        pub fn GetWitness(str: *const c_char) -> *const c_char;
+    }
+}
+
+pub fn get_witness(data: &str) -> String {
+    let c_config = CString::new(data).expect("invalid config");
+    let result = unsafe { witness::GetWitness(c_config.as_ptr()) };
+    let c_str = unsafe { CStr::from_ptr(result) };
+    let c_str_string = c_str.to_str().expect("Error translating from library");
+    c_str_string.to_string()
```
ed255, line 18: If my understanding is correct, the CStr is a type that holds a reference to an existing string, so this code doesn't own it and so Rust will not free it.  This string was allocated by go, but it will not be freed, leading a memory leak.
You can see how we resolved this in `geth-utils`: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/1bc1eef6344fba33105b018216d7af5f04550cd8/geth-utils/src/lib.rs#L31

Feel free to move this to a new issue if you'd like to fix this in a new PR :)

ed255 (approved):
LGTM!

mpt-witness-generator/rustlib/src/lib.rs
miha-stopar: But shouldn't be fixed also here? We will continue the development of the generator here, right?

miha-stopar (approved):
LGTM!

PR #1610, opened 2023-09-18
Fixing problems in MPT found by light client

miha-stopar:
### Description

Witness generator fixes:

-  When branch is hashed, its children are stored as hashes. This causes problems when computing the neighbour node which is needed for the case when a node is deleted from a branch with two nodes (the branch dissappears, the remaining node moves one level up).
- Placeholder storage leaf RLP - when the leaf value is set to zero, the RLP of the placeholder leaf RLP needs to be updated to reflect the shortened value.
- `CreateObject` called when there is no storage trie for an account.
- Setting code hash fixed.

Circuit fix:
When value is set to 0 (which deletes the key) and there are two nodes in the branch, then the node is deleted and the other node moves up one level (replaces the branch). The returned node is then this moved node which doesn't have the value 0 and the lookup into MPT tables fails.
Now the circuit is fixed to ensure the new value in the table is 0.

Besides that, some other problems are fixed which were found by @adria0 when integrating the light client and MPT (these were temporary hacks in MPT for testing purposes that were forgotten to be removed):

-  `SetState` calls removed in `prepare_witness.go` calls from
-  `oracle.PrefetchStorage` call uncommented in `prepare_witness.go`
-  `oracle.PrefetchStorage` call uncommented in `state/state_object.go updateTrie`
- Account nodes were not appended to all nodes in `prepare_witness.go`.

### Type of change

### How Has This Been Tested?

No test until now covered this scenario, now `TestNeighbourNodeInHashedBranch` has been added.


rrtoledo:
Could you add some unit tests to check that adding/updating/removing is done correctly perhaps? 
We could find some test vectors from other MPT libs

miha-stopar:
> Could you add some unit tests to check that adding/updating/removing is done correctly perhaps?
> We could find some test vectors from other MPT libs

You mean to the trie? The geth code is used, so it should be ok I think?

rrtoledo:
> > Could you add some unit tests to check that adding/updating/removing is done correctly perhaps?
> > We could find some test vectors from other MPT libs
> 
> You mean to the trie? The geth code is used, so it should be ok I think?

Yes, I meant for the trie. I thought not all cases might be covered an error was found (hence the PR) but I guess it is not trie based per se then.

miha-stopar:
> > > Could you add some unit tests to check that adding/updating/removing is done correctly perhaps?
> > > We could find some test vectors from other MPT libs
> > 
> > 
> > You mean to the trie? The geth code is used, so it should be ok I think?
> 
> Yes, I meant for the trie. I thought not all cases might be covered an error was found (hence the PR) but I guess it is not trie based per se then.

Yeah, the trie should be ok I think. What this PR solves are some bugs related to the conversion of the getProof result to the MPT witness + MPT table fix for a case when the leaf is deleted and there were just two leaves in the branch.

mpt-witness-generator/witness/leaf.go
```diff
@@ -474,21 +476,23 @@ func prepareStorageLeafInfo(row []byte, valueIsZero, isPlaceholder bool) ([]byte
 			// [227,48,161,160,187,239,170,18,88,1,56,188,38,60,149,117,120,38,223,78,36,235,129,201,170,170,170,170,170,170,170,170,170,170,170,170]
 			key[0] = row[0]
 			key[1] = row[1]
+			keyRlpLen = 1
```
rrtoledo, line 479: You can put the following in between the else L471 and the if line 472 and remove from the next lines:
```
keyRlpLen = 1
keyRlp = make([]uint8, keyRlpLen)
copy(keyRlp, row[:keyRlpLen])
valueRlpLen = 1
```

mpt-witness-generator/witness/leaf.go
```diff
@@ -418,42 +418,46 @@ func prepareStorageLeafInfo(row []byte, valueIsZero, isPlaceholder bool) ([]byte
...
+			// - 1 is the number of value bytes
+			keyRlp[0] = 192 + keyLen + 2
 		}
 	}
 
 	if len(row) < 32 { // the node doesn't get hashed in this case
+		keyRlpLen = 1
+		keyRlp = make([]uint8, keyRlpLen)
+		copy(keyRlp, row[:keyRlpLen])
+
```
rrtoledo, line 442: you can add `valueRlpLen = 1` here and remove it from the next lines

rrtoledo: The line `setKeyValue(keyLen, offset)` could also be moved at the very end of the function (and removed from all if/else conditions).

mpt-witness-generator/witness/prepare_witness.go
```diff
@@ -116,7 +115,10 @@ func obtainAccountProofAndConvertToWitness(i int, tMod TrieModification, tModsLe
 	} else if tMod.Type == BalanceChanged {
```
rrtoledo, line 115: Line 99: adding "be"  `	// This needs to be called before oracle.PrefetchAccount, otherwise oracle.PrefetchAccount
`

mpt-witness-generator/witness/prepare_witness.go
```diff
@@ -116,7 +115,10 @@ func obtainAccountProofAndConvertToWitness(i int, tMod TrieModification, tModsLe
...
 		statedb.SetBalance(addr, tMod.Balance)
 	} else if tMod.Type == CodeHashChanged {
-		statedb.SetCode(addr, tMod.CodeHash)
+		statedb.SetCodeHash(addr, tMod.CodeHash)
+		// For cases when the wrong account is obtained by PrefetchAccount:
```
rrtoledo, line 119: How do we know we are in that case?

rrtoledo (changes requested):

rrtoledo:
I'm not sure to understand everything you're doing, but here are a few fixes and questions I have.

mpt-witness-generator/witness/prepare_witness.go
miha-stopar: That's very good observation! I still need to handle this (for most cases it's ok, but the problems appears when code hash is set after setting nonce/balance)!

rrtoledo: Ok! Do you intend to handle it in another issue or this PR?

miha-stopar: Will be in this PR.

rrtoledo (approved):
LGTM!

Issue #1557, opened 2023-08-07
Port integration tests from zkevm-chains -- create necessary Txs/blocks for benchmarking


AronisAt79:

AronisAt79:
following issue #1552 and PR #1550 ,
the purpose here is to submit the necessary Txs for creation of blocks optimized for benchmarking

AronisAt79:
For this next step we will create a build script in order to generate the rust contract binding/s needed to interact with the benchmark-optimized smart contract/s

AronisAt79:
1. Add build dependencies to cargo.toml
2. Create build.rs - compiles smart contracts and created rust bindings
3. mv all contracts directly under contracts folder and replace all benchmark ones with a single contract (BENCHMARKS.sol)
4. remove compilation steps from gen_blockchain_data.rs and rewrite deploy function
5. Submit transactions for circuit benchmarks (MLOAD for EVM and SDIV for STATE)

AronisAt79:
CI checks are now passing : https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1615
Integration Tests are also successfully executed: https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6455922264/job/17524363132#logs

PR #1615, opened 2023-09-19
Integration tests benchmarks

AronisAt79:
### Description

Move SC compilation step to a build script. Rewrite deployment function based on SC rust bindings generated with 
[ethers_contract_abigen::Abigen](https://docs.rs/ethers-contract-abigen/latest/ethers_contract_abigen/)  and submit Txs for proof generation benchmarking.

### Issue Link

#1557 

### Type of change

- New feature (non-breaking change which adds functionality)

### Contents

- build.rs : Compile smart contracts and generate rust bindings
- gen_blockchain_data: 
       - Remove compilation step and rewrite fn deploy. New one uses the bindings generated at build step to deploy SCs and also instantiates SC instances to interact with.
      - Create a dump TxTrace function [needed for testing future implementations of benchmark contracts - needs attribute allow(dead_code)] 
      - Submit Txs optimized for maximum calls of MLOAD and SDIV up to 300k gas 
### How Has This Been Tested?
```
https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6264433154
```

~~https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6234076847/job/16920593180~~
~~>> Workflow build step is currently failing~~


AronisAt79:
Seems that introduction of build.rs causes build step to fail. 
Logs show that compilation of OpenZeppelinERC20TestToken.sol returns the following error:
```
"ParserError: Source \"contracts/vendor/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol\" not found: File not found. Searched the following locations: \"\".\n --> contracts/OpenZeppelinERC20TestToken.sol:5:1:\n  |\n5 | import \"./vendor/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol\";\n
```
I assume the problem is associated with the git submodule pointing to OpenZeppelin ERC20.sol but i have yet not managed to figure out and fix it. Could we maybe introduce a worflow step that adds/init the submodule instead of doing that in run.sh? (if i locally directly execute run.sh with a prior cargo build step, tests proceed)

ChihChengLiang:
I unassigned @leolara since Leo recently focused on the DSL effort.

AronisAt79:
@ChihChengLiang @ed255 
i moved the git submodule update step from run.sh to the action workflow definition. Now setup and gendata steps are passing. Will now work on the panic shown here https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/6260891594/job/16999835172

ChihChengLiang (commented):
First round of review. 
My main wish is to improve the error handling.
But if that takes time, we can also create an issue and defer it to future PR.

integration-tests/src/lib.rs
```diff
@@ -151,5 +149,14 @@ pub struct CompiledContract {
...
 
+#[allow(missing_docs)]
+#[allow(clippy::all)]
+pub mod bindings_benchmarks;
+#[allow(missing_docs)]
+#[allow(clippy::all)]
+pub mod bindings_greeter;
+#[allow(missing_docs)]
+#[allow(clippy::all)]
+pub mod bindings_openzeppelinerc20testtoken;
```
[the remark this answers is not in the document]
AronisAt79: Nope, i get an error that private item was used outside of scope.

integration-tests/build.rs
```diff
@@ -0,0 +1,154 @@
...
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    let solc: Solc = Solc::default();
+    let _solc_version = solc.version().expect("Version Works");
+
+    for (name, contract_path) in CONTRACTS {
+        let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
+        let inputs = CompilerInput::new(&path_sol).expect("CompilerInput Created");
```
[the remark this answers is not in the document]
AronisAt79: thanks for the suggestion @ChihChengLiang! will now work on this.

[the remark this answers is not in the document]
AronisAt79: @ChihChengLiang please have a look at commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1615/commits/2c61487b7c595f300929f52b756b2c635cf14b30

ChihChengLiang:
Hi @ed255, do you still have time to review this?

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+            }
+            Err(e) => eprintln!("{:#?}", e),
+        }
+    }
+    Ok(())
+}
+
+fn generate_rust_contract_bindings(bindings_dir: &str, file: &Path) -> Result<(), BuildError> {
+    const SLASH: char = '/';
+    const DOT: char = '.';
```
ed255, line 220: Nitpick: The PathBuf type has methods to "split the path by /" with https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.components
And to extract the base of the file https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.file_stem
I think using those methods would simplify a bit this function.

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// ABI
+    pub abi: Contract,
+    /// Bytecode
+    pub bin: Bytes,
+    /// Runtime Bytecode
+    pub bin_runtime: Bytes,
+}
+
+#[derive(Error, Debug)]
+enum BuildError {
```
ed255, line 29: I think you did some really nice work with this error type here!  Nevertheless I see that there's no error recovery or error handling in here: whenever an error is found, the build process stops with the error message.  For this reason I think it's also OK to just use `expect()` with a nice message of the failure, so that a panic happens on an error.
This is just a personal thought to reduce the amount of work in future code!  I'm not saying to change this (I think using the error type here is really clear and nice), but maybe you can skip some work just using `expect()` or even `panic()` in the context of `build.rs`.

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -1,40 +1,69 @@
...
+    (c_abi, contract_address, block_number, contract_instance)
+}
+
+#[allow(dead_code)]
+async fn dump_tx_trace(prov: Provider<Http>, receipt: TransactionReceipt, name: &str) {
+    let options = GethDebugTracingOptions::default();
+    let trace = prov
+        .debug_trace_transaction(receipt.transaction_hash, options)
+        .await
+        .expect("SOME REASON");
```
ed255, line 63: Maybe this was meant to be replaced with some reason? :laughing:

integration-tests/src/bin/gen_blockchain_data.rs
```diff
@@ -1,40 +1,69 @@
...
+    let filename = format!("{}_trace.json", name);
+    serde_json::to_writer(&File::create(filename).expect("cannot create file"), &trace)
+        .expect("reason");
```
ed255, line 66: Same here

ed255 (approved):
LGTM!  Everything looks really good and really clean!  Also I found it really well commented an organized!  I left a few nitpicks but they are minor, just some comments that popped up in my min while reviewing; feel free to read them and apply changes or not :)

integration-tests/build.rs
ChihChengLiang: Sorry, it was my bad to suggest changing to this style. The PR originally handled errors with `expect`s and `panic`s.
I agree with you that `build.rs` is a good place just to use panics, and I also agree that the error type looks clearer, so no change is required.

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+            .compile_output(&input)
+            .map_err(|err| BuildError::CompileOutputFailure(err.to_string()))?;
+        let mut deserializer: serde_json::Deserializer<serde_json::de::SliceRead<'_>> =
+            serde_json::Deserializer::from_slice(&output);
+        // The contracts to test the worst-case usage of certain opcodes, such as SDIV, MLOAD, and
+        // EXTCODESIZE, generate big JSON compilation outputs. We disable the recursion limit to
+        // avoid parsing failure.
+        deserializer.disable_recursion_limit();
+
+        let p = path_sol.to_str().ok_or_else(|| {
```
ChihChengLiang, line 133: In line 112 We have `path_sol.to_string_lossy().to_string()`, and here we use `path_sol.to_str()`.
We can handle this error before line 112, so that we can use a well-formed path string in the rest of the place.

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    for (name, contract_path) in CONTRACTS {
+        let path_sol = Path::new(CONTRACTS_PATH).join(contract_path);
+        let input = CompilerInput::new(&path_sol)
+            .map_err(|err| BuildError::FailedToComposeCompilerInputs {
+                path: path_sol.to_string_lossy().to_string(),
+                reason: err.to_string(),
+            })?
+            .pop()
+            .ok_or_else(|| BuildError::ArtifactError("No Artifact Found".to_string()))?
```
ChihChengLiang, line 116: The purpose of having a String argument in the error type is to add additional context for debugging. 
For example, we can add the path of the failure or some other error string propagated from the low-level library. The arguments should add more context for the developers to debug.

We use string arguments like "No Artifact Found" in many places in this PR, which doesn't provide additional context than the error type "ArtifactError" itself. 

We can remove string arguments from these error types.

```suggestion
            .ok_or_else(|| BuildError::ArtifactError)?
```

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+            // ethers-solc: explicitly indicate the EvmVersion that corresponds to the zkevm
+            // circuit's supported Upgrade, e.g. `London/Shanghai/...` specifications.
+            .evm_version(EvmVersion::London);
+
+        // compilation will either fail with Err variant or return Ok(CompilerOutput)
+        // which may contain Errors or Warnings
+        let output: Vec<u8> = solc
+            .compile_output(&input)
+            .map_err(|err| BuildError::CompileOutputFailure(err.to_string()))?;
```
ChihChengLiang, line 125: This is a helpful string argument for the error type, because it provides the error string from the solc side.

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        let _ = error_free.then_some(|| ()).ok_or_else(|| {
+            BuildError::CompilerOutputContainsErrors {
+                path: p.to_string(),
+                errors: compiled.errors,
+            }
+        })?;
+
+        let contract: CompactContractRef = compiled_binding.get(p, name).ok_or_else(|| {
+            BuildError::FailedToLoadCompactContractRef("FailedToLoadCompactContractRef".to_string())
```
ChihChengLiang, line 153:
```suggestion
            BuildError::FailedToLoadCompactContractRef
```

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+        })?;
+
+        let abi = contract
+            .abi
+            .ok_or_else(|| {
+                BuildError::ErrorLoadingContractAbi("Failed to get contract Abi".to_string())
```
ChihChengLiang, line 159:
```suggestion
                BuildError::ErrorLoadingContractAbi
```

ChihChengLiang (approved):
LGTM after Edu's and my comments are addressed.

integration-tests/build.rs
AronisAt79: I think we can let it be for now. In any case CC's suggestion was a very useful direction for me on using the error type for cases were we want to actually handle the error itself!

integration-tests/build.rs
AronisAt79: Yes that makes sense.

AronisAt79:
@ChihChengLiang proposed removals of redundant string on BuildError variants plus enum adaptations and replacement of ok_or_else() with ok_or() according to clippy output addressed with commits: d80a1b4dc0964765527b7816fc5654baf11c9c1c & 3a0a27f1d40300340eae7bf345853fb6c2306361

integration-tests/build.rs
AronisAt79: resolution: declare p:  &str and catch the conversion error:
```
        let p = path_sol.to_str().ok_or_else(|| {
            BuildError::StringConversionError(
                "Failed to convert provided path to &str".to_string(),
            )
        })?;
```
Then use   p for creating  CompilerInputs and CompactContractRefs
Commit:

integration-tests/src/bin/gen_blockchain_data.rs
AronisAt79: thanks :). >>> .expect("Failed to get transaction debug trace");

integration-tests/src/bin/gen_blockchain_data.rs
AronisAt79: >>>>  .expect("Could not seralize trace as json data");

integration-tests/build.rs
```diff
@@ -0,0 +1,247 @@
...
+        let compiled_binding = compiled.clone();
+
+        let error_free = !{ compiled.has_error() || compiled.has_warning(WARN) };
+
+        let _ = error_free.then_some(|| ()).ok_or_else(|| {
+            BuildError::CompilerOutputContainsErrors {
+                path: p.to_string(),
+                errors: compiled.errors,
+            }
+        })?;
```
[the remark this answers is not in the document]
AronisAt79: fair enough :). commit ec58e85648a03bde654bfcd98555279b119e9d30

integration-tests/build.rs
AronisAt79: thanks for the suggestion. Commit 8c61fd3cd7563827af6b8009d53f0af20b7debf6

ChihChengLiang (approved):
LGTM

Issue #1513, opened 2023-07-04
Improve BufferReaderGadget


ChihChengLiang:
### Describe the feature you would like

We should add some tests for `BufferReaderGadget`.

- The tests should cover different start addresses and end addresses

See #1485 for more insights.

NOOMA-42:
Can I pick up this issue? 

p.s. This seems also relate to https://github.com/privacy-scaling-explorations/zkevm-specs/issues/440

ChihChengLiang:
Hi @NOOMA-42, 
Yes, feel free to pick up this issue! The spec 440 issue you mentioned is related too. Feel free to take that issue too!

PR #1621, opened 2023-09-21
Feat/#1513 buffer reader gadget unit test

NOOMA-42:
### Description

Solve #1513 by adding a testutil.rs and test cases

### Issue Link

#1513 

### Type of change

### Contents

- Add a unit test to buffer reader gadget in memory gadget and test util.
- The testcases cover completeness and soundness.

### Rationale

I followed test util in `math_gadget` and the design of test container. I removed lookup table part in the `test_util.rs` of math_gadget, since buffer reader gadget didn't use lookup table.

### How Has This Been Tested?

`cargo test --package zkevm-circuits --features test-circuits -- evm_circuit::util::memory_gadget`


NOOMA-42:
This PR is ready for review, I have a few questions though:
1. Should I squash to just 1 commit?
2. The `test_util` is mostly the same as it in `math_gadget` but I remove the lookup table part. There's still a 3 phase challenge in [current design](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1621/files#diff-8545dbd217ceb05401c329e057a7645643ab170670928e59a0270b01929f8168R84) What should I go about it?

ChihChengLiang:
Hi @NOOMA-42,
Thanks for your contribution!

> Should I squash to just 1 commit?

No need. The commits on this branch you sent are perfect.

We merge pull requests with the squashing option enabled. But the squashing is performed automatically by Github. You don't need to do any manual action for squashing.

> The test_util is mostly the same as it in math_gadget but I remove the lookup table part. There's still a 3 phase challenge in [current design](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1621/files#diff-8545dbd217ceb05401c329e057a7645643ab170670928e59a0270b01929f8168R84) What should I go about it?

We can remove the 3 phase challenge if we don't intend to do lookups.

ChihChengLiang:
Hi @NOOMA-42, please fix the linting.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -686,3 +686,154 @@ impl<F: Field, const MAX_BYTES: usize, const ADDR_SIZE_IN_BYTES: usize>
...
+
+    #[test]
+    fn test_buffer_reader_gadget_soundness(){
+        // buffer len < data len
+        try_test!(
+            BufferReaderGadgetTestContainer<Fr, 32, 10>,
+            vec![
+                Word::from(0),
+                Word::from(1), 
+                Word::from(1024),
```
KimiWu123, line 812: nit: suggest to change the 3rd input to 256.

The data size in your first two cases are 2 bytes (256) and 1 byte (255) and this one is 2 bytes (1024). If you want to represent 2 bytes data, it's better to use either 256 or 1024 (just one fixed value). If so, reviewers could understand what you want to test easily. Different valus could represent different things. So, I would suggest to change this value to 256 (2 bytes data and the len is larger than the given buffer size) and change 255 to 256 in above case either (bcs your buffer is 31 bytes, way to larger than 2 bytes data)

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
```diff
@@ -686,3 +686,154 @@ impl<F: Field, const MAX_BYTES: usize, const ADDR_SIZE_IN_BYTES: usize>
...
+                    byte.assign(region, offset, Value::known(F::from(*input_byte as u64)))
+                )
+                .collect::<Result<Vec<_>, _>>()?;
+            self.buffer_reader_gadget.assign(region, offset, addr_start, addr_end, &input_bytes[..])?;
+
+            Ok(())
+        }
+    }
+    
+    #[test]
```
KimiWu123, line 779: perhaps you could have a set of cases which has smaller `MAX_BYTES`, smaller than given buffer size or data size and to see how it goes.

KimiWu123 (commented):
First round pass, testing looks good to me.

However, I was thinking the need to introduce a container trait here. Currently, it's only used by `BufferReaderGadget`.  It's fine with me if we have plan to complete testing of other memory gadget. But is it the case? Another concern of mine is the code in `test_util.rs` is 90% the same as the one in `math_gadget`. If we decide to keep it, I would recommend to create separated issue for refactoring. Would like to hear your opinion @ChihChengLiang.

ChihChengLiang:
@NOOMA-42 @KimiWu123

It's okay to duplicate the code in this PR. We can de-duplicate in future PR.

zkevm-circuits/src/evm_circuit/util/memory_gadget.rs
NOOMA-42: I can test MAX_BYTES < buffer size returning false but can't test against MAXBYTES < data size, because of this [assert](https://github.com/NOOMA-42/zkevm-circuits/blob/feat/%231513-BufferReaderGadget-unittest/zkevm-circuits/src/evm_circuit/util/memory_gadget.rs#L652) panic

KimiWu123 (approved):
LGTM! Nice work!

ChihChengLiang (approved):
LGTM. Thanks for the contribution

PR #1624, opened 2023-09-22
fix: parse "TLOAD/TSTORE" correctly in trace

lispc:
### Description

0xb3 / 0xb4 are rendered as "TLOAD/TSTORE" in geth trace json.

### Issue Link

[_link issue here_]

### Type of change


han0110 (commented):
Seems like `TLOAD/TSTORE` are updated to `0x5c` and `0x5d`

See https://github.com/ethereum/go-ethereum/pull/27613 and https://eips.ethereum.org/EIPS/eip-1153

Also perhaps we could handle `MCOPY (0x5e)` and `BLOBBASEFEE (0x4a)` in the same way in this PR.

lispc:
the geth inside geth-utils is v1.11.5, earlier than https://github.com/ethereum/go-ethereum/pull/27613.

So either temporarily keep using  0xb3/0xb4, or upgrade geth and change TSTORE/TLOAD opcode.

ChihChengLiang (approved):

Issue #1622, opened 2023-09-22
testool issues


lispc:
### Describe the bug

1. `Ignored_SkipTestDifficulty` and `Ignored_SkipTestBalanceOverflow` not needed.  They are related limitation of scroll l2 evm. I guess these tests should work well for standard evm.
2. i noticed the html only shows "circuit was not satisfied" without more details. I remembered previously detailed failure msg is shown on the web page?

PR #1630, opened 2023-09-27
Address various testool issues

ChihChengLiang:
### Description

- [x] `SkipTestDifficulty` and `SkipTestBalanceOverflow` should not be ignored.
- [ ] `creationTxInitCodeSizeLimit_d1(invalid)_g0_v0` should be parsed as exception and return ok.
- [x]  skip precompile tests
- [x] Use shanghai fork

Not addressed

- Fix "circuit was not satisfied"

### Issue Link

#1622

### Type of change

### DIFF
--- a/testool/src/statetest/executor.rs
+++ b/testool/src/statetest/executor.rs
@@ -4,7 +4,7 @@ use bus_mapping::{
     circuit_input_builder::{CircuitInputBuilder, FixedCParams},
     mock::BlockData,
 };
-use eth_types::{geth_types, Address, Bytes, GethExecTrace, ToBigEndian, U256, U64};
+use eth_types::{geth_types, Address, Bytes, GethExecTrace, U256, U64};
 use ethers_core::{
     k256::ecdsa::SigningKey,
     types::{transaction::eip2718::TypedTransaction, TransactionRequest},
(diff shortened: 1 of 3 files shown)


testool/src/statetest/executor.rs
```diff
@@ -32,34 +32,28 @@ pub enum StateTestError {
...
+        num_failure: usize,
+        first_failure: String,
+    },
     #[error("SkipTestMaxGasLimit({0})")]
     SkipTestMaxGasLimit(u64),
     #[error("SkipTestMaxSteps({0})")]
     SkipTestMaxSteps(usize),
     #[error("SkipTestSelfDestruct")]
     SkipTestSelfDestruct,
-    #[error("SkipTestDifficulty")]
```
ed255, line 41: Removing this code will mean the diff between our testool and scroll version will increase; which means that more effort is needed when porting or upstreaming changes.

What about leaving this this code but introducing some flag to trigger one behavior or the other?  This can be a rust feature flag, or a variable that is read from the environment.  Then in the `check_geth_traces` method, depending on this flag the behavior would change.

For example we could use an enum like this:
```rust
enum FeatureSet {
  EthereumL1,
  Scroll,
  Taiko,
}
```
and set a global variable with a value chosen at runtime.

Maybe this suits better another PR, but I think the testool would be a nice candidate to try this approach.

testool/src/statetest/executor.rs
```diff
@@ -350,7 +339,12 @@ pub fn run_test(
         builder = _builder;
 
         let prover = MockProver::run(k, &circuit, instance).unwrap();
-        prover.assert_satisfied_par();
+        prover
+            .verify()
```
ed255, line 343:
```suggestion
            .verify_par()
```
Suggestion to get parallelized MockProver verification

ed255 (approved):
LGTM!  Please take a look at my first comment, if you like the idea, feel free to implement it in this PR, or in a future PR.

testool/src/statetest/executor.rs
ChihChengLiang: I like the idea of adding a feature set but want to name it like "ZeroDifficulty," "L2Specific", or some names that refer to the feature's functions but the team's name. I'll create an issue for this.

For now, I'll keep the `SkipTestDifficulty` and `SkipTestBalanceOverflow` so that they can be feature-flagged later.

testool/src/statetest/executor.rs
ChihChengLiang: @adria0 Before working on this PR, I thought that `assert_satisfied_par` dumps richer output, too.

However, the [report](http://testool-public.s3-website.eu-central-1.amazonaws.com/nightly.1695857921.ee97ab1.html#all) shows only "circuit was not satisfied" but not the rich output. My theory is that `assert_satisfied_par` panicked so the output was not captured by `StateTestError`, that's why I customized a StateTestError to capture the VerifyFailure.

PR #1632, opened 2023-09-28
move bytecode circuit to conform with the codebase style

ChihChengLiang:
### Description

We move zkevm-circuits/src/bytecode_circuit/circuit.rs to zkevm-circuits/src/bytecode_circuit.rs

### Issue Link

#1607

### Type of change

Bug fix (non-breaking change which fixes an issue)

Before

```
wc -l zkevm-circuits/src/*circuit.rs
      12 zkevm-circuits/src/bytecode_circuit.rs
     887 zkevm-circuits/src/copy_circuit.rs
     528 zkevm-circuits/src/evm_circuit.rs
     550 zkevm-circuits/src/exp_circuit.rs
    1069 zkevm-circuits/src/keccak_circuit.rs
     796 zkevm-circuits/src/mpt_circuit.rs
    1653 zkevm-circuits/src/pi_circuit.rs
     190 zkevm-circuits/src/root_circuit.rs
     614 zkevm-circuits/src/state_circuit.rs
     493 zkevm-circuits/src/super_circuit.rs
     361 zkevm-circuits/src/tx_circuit.rs
    7153 total
```

After

```
wc -l zkevm-circuits/src/*circuit.rs
     828 zkevm-circuits/src/bytecode_circuit.rs
     887 zkevm-circuits/src/copy_circuit.rs
     528 zkevm-circuits/src/evm_circuit.rs
     550 zkevm-circuits/src/exp_circuit.rs
    1069 zkevm-circuits/src/keccak_circuit.rs
     796 zkevm-circuits/src/mpt_circuit.rs
    1653 zkevm-circuits/src/pi_circuit.rs
     190 zkevm-circuits/src/root_circuit.rs
     614 zkevm-circuits/src/state_circuit.rs
     491 zkevm-circuits/src/super_circuit.rs
     361 zkevm-circuits/src/tx_circuit.rs
    7967 total
```

Issue #1607, opened 2023-09-16
refactor(bytecode): unify circuit style


lispc:
### Describe the feature you would like

due to historical reasons, the codes of bytecode circuit is a bit different than other circuits:

1. circuit implementation file location.  Other circuits puts circuit/config implementation directly inside `src`, while bytecode circuit has another level of hierarchy.

```
 $ wc -l zkevm-circuits/src/*circuit.rs
      12 zkevm-circuits/src/bytecode_circuit.rs
     854 zkevm-circuits/src/copy_circuit.rs
     528 zkevm-circuits/src/evm_circuit.rs
     550 zkevm-circuits/src/exp_circuit.rs
    1069 zkevm-circuits/src/keccak_circuit.rs
     858 zkevm-circuits/src/mpt_circuit.rs
    1653 zkevm-circuits/src/pi_circuit.rs
     190 zkevm-circuits/src/root_circuit.rs
     614 zkevm-circuits/src/state_circuit.rs
     493 zkevm-circuits/src/super_circuit.rs
     361 zkevm-circuits/src/tx_circuit.rs
```

2. the use of `minimum_rows: meta.minimum_rows()`. Other circuits don't use this.

ChihChengLiang:
It looks like the actions to take are:

1. Move the bytecode circuit to the correct file. 
2. Remove the `minimum_rows: meta.minimum_rows()`, and fix the tests that depends on `minimum_rows`.

Note that `minimum_rows` is the number of reserved rows for blinding factors and other stuff. It helps the verifier compute the maximum possible number of rows we can utilize. 

Here's how it works. The bytecode circuit takes the parameter "size" to configure, usually, the size is two to the power of degree, as the example in the circuit benchmark shows.

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/bb0d9e479a8081c5e7d903b93d68b97c66b46bce/circuit-benchmarks/src/bytecode_circuit.rs#L49-L53

The bytecode circuit then subtracts `minimum_rows` from size, resulting in the maximum number of usable rows. The verifier then sets the last row selector `q_last` there. 

However, I realized that's not how the current supercircuit works. We didn't configure the bytecode circuit size in the super circuit with degrees but instead with the number of bytes (technically the number of bytes plus the number of headers) to prove.

[...]

PR #1633, opened 2023-09-28
Remove minimal_rows in the Bytecode circuit

ChihChengLiang:
### Description

- Remove minimal_rows 
- Configure the bytecode circuit with the circuit degree instead of size.

### Issue Link

#1607

### Type of change

[x] Bug fix (non-breaking change which fixes an issue)

### Note

Please review this after #1632 is merged.

### DIFF
--- a/zkevm-circuits/src/bytecode_circuit.rs
+++ b/zkevm-circuits/src/bytecode_circuit.rs
@@ -16,7 +16,7 @@ use crate::{
     },
     table::{BytecodeFieldTag, BytecodeTable, KeccakTable, LookupTable},
     util::{
-        self, get_push_size,
+        self, get_push_size, log2_ceil,
         word::{empty_code_hash_word_value, Word, Word32, WordExpr},
         Challenges, Expr, SubCircuit, SubCircuitConfig,
     },
@@ -162,7 +162,6 @@ impl<F: Field> Deref for BytecodeCircuitAssignment<F> {
 #[derive(Clone, Debug)]
 /// Bytecode circuit configuration
 pub struct BytecodeCircuitConfig<F> {
-    minimum_rows: usize,
     q_enable: Column<Fixed>,
     q_first: Column<Fixed>,
     q_last: Column<Fixed>,
@@ -555,7 +554,6 @@ impl<F: Field> SubCircuitConfig<F> for BytecodeCircuitConfig<F> {
         let index_length_diff_is_zero = IsZeroChip::construct(index_length_diff_is_zero);
 
         BytecodeCircuitConfig {
-            minimum_rows: meta.minimum_rows(),
             q_enable,
             q_first,
             q_last,
@@ -576,19 +574,17 @@ impl<F: Field> BytecodeCircuitConfig<F> {
     pub(crate) fn assign_internal(
         &self,
         layouter: &mut impl Layouter<F>,
-        size: usize,
+        degree: u32,
+        unusable_rows: usize,
         witness: &BytecodeCircuitAssignment<F>,
         challenges: &Challenges<Value<F>>,
     ) -> Result<(), Error> {
-        // Subtract the unusable rows from the size
-        assert!(size > self.minimum_rows);
-
-        let last_row_offset = size - self.minimum_rows + 1;
+        let last_row_offset = 2usize.pow(degree) - unusable_rows - 1;
 
         trace!(
-            "size: {}, minimum_rows: {}, last_row_offset:{}",
-            size,
-            self.minimum_rows,
+            "degree: {}, unusable_rows: {}, last_row_offset:{}",
+            degree,
+            unusable_rows,
             last_row_offset
         );
         if witness.len() > last_row_offset {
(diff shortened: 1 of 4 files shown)


zkevm-circuits/src/bytecode_circuit.rs
```diff
@@ -576,19 +574,17 @@ impl<F: Field> BytecodeCircuitConfig<F> {
     pub(crate) fn assign_internal(
         &self,
         layouter: &mut impl Layouter<F>,
-        size: usize,
+        degree: u32,
+        unusable_rows: usize,
         witness: &BytecodeCircuitAssignment<F>,
         challenges: &Challenges<Value<F>>,
     ) -> Result<(), Error> {
-        // Subtract the unusable rows from the size
```
hero78119, line 583: Nitpick: notice `unusable_rows` and `degree` is just to compute `last_row_offset`. Do you think is it good idea to compute `last_row_offset` outside and just pass `last_row_offset` as assign_internal argument?

zkevm-circuits/src/bytecode_circuit.rs
```diff
@@ -576,19 +574,17 @@ impl<F: Field> BytecodeCircuitConfig<F> {
...
-        assert!(size > self.minimum_rows);
-
-        let last_row_offset = size - self.minimum_rows + 1;
+        let last_row_offset = 2usize.pow(degree) - unusable_rows - 1;
```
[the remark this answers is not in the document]
ChihChengLiang: I'm leaning toward leaving the computation of last_row_offset inside assign_internal for the separation of concern reason. 

The SubCircuit trait should know only the degree and unusable rows but not the last_row_offset. last_row_offset is only used inside assign_internal, so the function should compute them.

zkevm-circuits/src/bytecode_circuit.rs
```diff
@@ -779,17 +775,17 @@ pub struct BytecodeCircuit<F: Field> {
...
     pub(crate) rows: BytecodeCircuitAssignment<F>,
     /// Circuit size
-    pub size: usize,
+    pub degree: u32,
 }
 
 impl<F: Field> BytecodeCircuit<F> {
     /// new BytecodeCircuitTester
-    pub fn new(bytecodes: CodeDB, size: usize) -> Self {
+    pub fn new(bytecodes: CodeDB, degree: u32) -> Self {
```
ed255, line 783: I think that parametrizing the circuit with just the degree has some downsides, in particular for testing with the MockProver.
Most of the times when running the MockProver we want the verification to only apply to rows that contain actual data.  The only case where we want more than that would be to test the padding strategy that the circuit applies to support a fixed size circuit with a smaller than max capacity data; but that's usually a single test.

By configuring with degree, when running the MockProver we'll be checking more rows than strictly necessary.  Let's say we have some bytecodes that need 1200 rows.  We would need to verify 2048 rows with the current approach.

Also this is specially important when testing the SuperCircuit with the MockProver.  Let's imagine that for a block we need 4000 rows for the TxCircuit and 100 for the block circuit.  Then the SuperCircuit degree would be k=12 (4096 rows).  Would you pass degree=12 to the TxCircuit?  Maybe you'd pass degree=7 (128 rows) to have a better verification time, but that feels strange because that degree is not the real one.

We have other circuits that don't receive the degree, and instead receive how many rows to enable, or the maximum number of operations to verify.
- `CopyCircuit` takes `max_copy_rows`
- `KeccakCircuit` takes `num_rows`
- `ExpCircuit` takes `max_exp_steps`

To avoid dynamic circuits on real proofs, we have the `DynamicCParams` and `FixedCParams` types in `bus-mapping/src/circuit_input_builder.rs`.  The first allows the circuit input builder to chose the circuit parameters dynamically based on a specific block input; and the second allows us to pick static parameters regardless of the input (so we get fast MockProver verification, and constant circuits for the real prover).

ed255: On the other hand: I agree that saying:
```rust
let last_row_offset = 2usize.pow(degree) - unusable_rows - 1;
```
Is correct and clear.
Whereas if we only have the parameters `max_rows` and `unusable_rows`, I would argue that
```rust
let last_row_offset = max_rows - 1;
```
But then we don't know if `last_row_offset <= 2^k - unusable_rows - 1`, which is a nice check to have as early as possible.

As a possible solution, 2 ideas come to my mind:
- a) Pass the `max_rows` and `degree` to each circuit so that we can have the full view and do all the early checks, while supporting fast MockProver.verify.  Note that for a real proof we would need to set `max_rows = 2^degree - unusable_rows()`, but we already have that API in the SubCircuit trait
- b) Extend halo2 (either the `Layouter` or the `ConstraintSystem`) to expose the circuit degree, so that we can get that value during configuration or synthesis time, and do all the checks we want with it.

ed255 (commented):
Please take a look at my long discussion :P

zkevm-circuits/src/bytecode_circuit.rs
ChihChengLiang: I see the reasoning now.

- To save the testing time on Mock prover, I'll change the parameter from "degree" back to "size." Or even better, rename "size" to "max_rows."
- I'll comment on the rationale for using size.
- I'll comment on DynamicCParams and FixedCParams

For validating max_rows, I like the idea for Halo2 to expose the circuit degree.

ChihChengLiang: Hi @ed255,
I've addressed the feedback, can you check again?

ed255 (approved):
LGTM!

Issue #1601, opened 2023-09-14
[proof-chunk] super circuit implement new permutation gadget for evm rwtable lookup


hero78119:
This issue focus on super circuit change scope. 

Based on the design https://hackmd.io/NPDfrFfjQoa5ebzH6bS39w 
There will be 2 RwTable
- [new] chronologically sorted rwtable
- [existing] multi-sorted rwtable.

> multi-sorted rwtable https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/1bc1eef6344fba33105b018216d7af5f04550cd8/zkevm-circuits/src/witness/rw.rs#L141-L150

EVM circuit will switch to lookup into new chronologically-sorted rwtable, and at the end we need to constraints multi-sorted rwtable vs chronologically sorted rwtable are permutation of each others.

To prove permutation of 2 tables, we can introduce a new permutation gadget to layout the custom gates to constraint permutation fingerprints updated of 2 rwtable, and at the last row we constraints the permutation fingerprints equality of 2 rwtable fingerprints.

## Permutation fingerprint challenge

To update the permutation fingerprints there need a challenge. In super circuit scope we can trust challenge bypass from public-input first, and constraint its consistency from row 1 to the end row. The validation of challenge will be outside of super circuit, which is in the scope of root circuit + snark-verifier in another issues. https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1603

## Last chunk (reminded by @ed255 )
Introduce new public input `is_last_chunk: bool` to toggle some extra custom gates check for last chunk
- Last step should be `EnbBlock`
- permutation fingerprint LHS/RHS equality
- .... 

Besides to keep scope smaller, this issue keep single supercircuit proof (chunk) as usual. Multiple supercircuit proof chunk can be defered into https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1604

ed255:
Status: this issue has already been resolved via https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1641 and is now in the branch https://github.com/privacy-scaling-explorations/zkevm-circuits/tree/proof-chunk but not yet in main

PR #1641, opened 2023-10-03
[proof chunk]  refactor evm_circuit/state_circuit to support proof chunk

hero78119:
### Description

[_PR description_]

### Issue Link

#1601 

### Type of change

### Highlights
- separate rw_table in evm_circuit/state_circuit and each have its own permutationchip. fingerprints checking will be deferred to root circuit.
- disable rwtable first row check (by disable selector offset=0 in state_circuit, other offset toggle on as usual), since it will be copy constraints via public input.
- keep `Rw::Start` and introduce `Rw::Padding` to support post-padding in address-sorted rwtable in last_chunk. 
- instance (new public input) column are design into two, one for prev_chunk context, another for current_chunk context to be used for next chunk. So in root circuit commitment of instance column can be used in snark-verifier without need to unwrap it. 

> Update on 18 Oct 2023: beginchunk/endchunk virtual step, inner_rw_counter, chunkctx

- add Begin/EndChunk virtual step. BeginChunk is not mandatory in first chunk first step. while EndChunk is not mandatory in last chunk last step.
- add `inner_rw_counter` which is local rw_counter within each chunk. This is used to count valid rw_row and assure Rw::Padding are append consecutively in `end_block.rs` logic => EndChunk should apply similar check later on
- introduce chunkctx->{chunk_index, total_chunks} to tell first chunk (chunk_index==0) and last chunk (chunk_index + 1 == total_chunks) during witness generation/constraints assignment
- add chunkctx_table to able to lookup chunk context (chunk_index, next_chunk_index, total_chunk, initial_rwc, end_rwc..etc) in exec step to allow various conditional check based on current chunk context

### How Has This Been Tested?

[_explanation_]

### More context on Rw::Padding (Not cover in current PR, will be handled in later multiple chunk PR to make scope smaller)

In new logic, `Rw::Start` will be insert in first chunk offset 0, while other holes are filled by `Rw::Padding` in last chunk(s). The address-sorted rwtable layout will be 
```
address-sorted rwtable
## first chunk
[
      Rw::start,  // offset 0, Rw::Start is only in first chunk, and only in offset 0, constrainted by public input
      ....(normal Rw), 
       ...(Rw::Padding),  // padding if there are only one chunk
]

## end chunk (if there are > 1 chunk)
[
      ....(normal Rw),  // offset 0 are carry over from previous chunk, other are address sorted
       ...(Rw::Padding) // padding in end chunk
]
```

For chronologically rwtable, since there is no in-row constraints to check each Rw operation, so theoretically Rw::Padding rows can be filled in any offset. However, we also need to assure there is no malicious insertion other than Rw::Padding. To do that, we will rewrite this logic in later PR https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit/execution/end_block.rs#L86-L90 to lookup consecutive `Rw::Padding` at **chronologically** sorted table, at the END of EACH chunk. 

> A tricks: first Rw::Padding rw_counter need to be set as last (globally) valid row rw_counter + 1. This is to make sure in both chronologically rw_table or address-sorted rw_table it's always append in the end of rw_table.

```
chronologically rwtable, ascending sorted by `rw_counter`.
## first chunk
[
      Rw::start,  // offset 0, Rw::Start is only in first chunk, constrainted by public input
      ...(normal Rw),
      ...(Rw::Padding),   // first Rw::Padding rw_counter need to be set as last (globally) valid row rw_counter + 1, last means from last !!chunk!!. It assures Rw::Padding always append at the end of each chunk
]

## end chunk (if there are > 1 chunk)
[
      ....(normal Rw),  // offset 0 are carry over from previous chunk, other are rw_counter sorted
       ...(Rw::Padding) // padding, also rw_counter sorted
]
```


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -413,11 +556,6 @@ impl<F: Field> Circuit<F> for EvmCircuit<F> {
             block.circuits_params.max_calldata,
         )?;
         block.rws.check_rw_counter_sanity();
-        config.rw_table.load(
```
hero78119, line 416: This was moved to `synthesis_sub` since now evm_circuit as `Circuit`/`SubCircuit` are needed to load rw_table (because it's separated from state_circuit already)

bus-mapping/src/exec_trace.rs
```diff
@@ -14,6 +14,7 @@ impl fmt::Debug for OperationRef {
             "OperationRef{{ {}, {} }}",
             match self.0 {
                 Target::Start => "Start",
```
ed255, line 16: The `Start` tag was already used as padding.  Is `Start` still necessary in the chunking approach?  I think we could remove `Start`.

From the PR description I see you're already aware of the missing checks to make sure that the chronological rw table doesn't contain extra entries not found in the evm circuit.  I think that can be solved with a very similar solution that we did with the Start tag.
Also I want to mention that I think it doesn't matter weather the padding rows are at the beginning or at the end (as long as they are forced to be consecutive).

zkevm-circuits/src/witness/block.rs
```diff
@@ -50,6 +51,24 @@ pub struct Block<F> {
     pub keccak_inputs: Vec<Vec<u8>>,
     /// Original Block from geth
     pub eth_block: eth_types::Block<eth_types::Transaction>,
+
```
ed255, line 54: I'm wondering about the naming of the structs.  This struct is called `Block`, but it has fields that are for chunks.
Would it make sense to take a `Block` and from it generate `n` `BlockChunk` types? and all the fingerprint fields would only appear in the `BlockChunk` struct.

Then a update all the circuits to take a `BlockChunk` instead of a `Block`?

Another option is for the `Block` to contain a field called `chunks` with a vector of `Chunks` which has the chunk specific fields (like the fingerprints).  Then each circuit takes a `Block` and a `chunk_index`.

zkevm-circuits/src/super_circuit.rs
```diff
@@ -122,7 +122,10 @@ impl<F: Field> SubCircuitConfig<F> for SuperCircuitConfig<F> {
         }: Self::ConfigArgs,
     ) -> Self {
         let tx_table = TxTable::construct(meta);
+
+        let chronological_rw_table = RwTable::construct(meta);
```
ed255, line 126: I think it would be more clear if here we see the two rw_tables: `chronological_rw_table` and `by_address_memory_table`.
One is passed to the evm_circuit and the other to the state_circuit.

zkevm-circuits/src/state_circuit.rs
```diff
@@ -149,13 +160,30 @@ impl<F: Field> SubCircuitConfig<F> for StateCircuitConfig<F> {
...
         // annotate columns
         rw_table.annotate_columns(meta);
         mpt_table.annotate_columns(meta);
         u8_table.annotate_columns(meta);
         u10_table.annotate_columns(meta);
         u16_table.annotate_columns(meta);
 
+        <RwTable as LookupTable<F>>::columns(&rw_table)
+            .iter()
+            .for_each(|c| meta.enable_equality(*c));
```
ed255, line 177: From the `enable_equality` I imagine that you're doing the fingerprint in another region and passing the cells via copy constraints.  I think the fingerprint could be implemented as a gadget applied inline at every row of the state circuit, to skip the copy constraints.  But this is an optimization, the current solution should work for now!

ed255 (commented):
I've done a very quick overview and left some comments!  I didn't find anything strange :)  Note that I didn't check the fingerprint chip

bus-mapping/src/exec_trace.rs
hero78119: Agree! Right now It seems to me that we can remove `Start` and only use `Padding`.

Regarding to padding at beginning or end, I prefer `end` for the reason of `better coding style` in chronologically 
 rw_table handling. More elaboration, in later multiple chunk modification, I imagine the rw_table chunk should be pre-process in bus-mapping to chunk into BlockChunk structure
> while BlockChunk is almost the same as current `Block`, so the circuit assigment logic require less change

To literally represent `chronologically_rw_table` meaning and still make padding consecutive appearing, `chronologically_rw_table` should be sorted by `rw_counter` in ascending order **inside** chunk , while other non-padding rw rw_counter still start from uint`1`.  Therefore padding step rw_counter better assign largerly to make it consecutive appear at the end.

hero78119: @ed255 based on my reasoning above, padding at the end make life easier.
Then I realized we need add `Padding` (largest enum), and also need to keep `Start` (smallest enum) and make it appear in first chunk first row. Because first chunk first row is public input and we will design it like constant commitment in root circuit, so we need smallest enum `Start` to assure it can appear in the first row of `address-sorted-rwtable`

zkevm-circuits/src/super_circuit.rs
hero78119: fixed and renamed by by_address_rw_table

zkevm-circuits/src/witness/block.rs
hero78119: Thanks for the suggestion!
Right now the naming and place is naively design  
I plan to refactor this part in later multiple chunks PRs, at that moment we will be more clear on how to design this part. Wdyt :) ?

hero78119:
All unittests is fixed and CI passed.
To avoid PR grows to be too big for review, I think current scope is ok to merge as first PRs, and we left other changes later
Any suggestion is appreciated and welcome!

zkevm-circuits/src/state_circuit/constraint_builder.rs
```diff
@@ -109,12 +112,23 @@ impl<F: Field> ConstraintBuilder<F> {
             .collect()
     }
 
-    pub fn lookups(&self) -> Vec<Lookup<F>> {
```
hero78119, line 112: Offset 0 is constraint by public input, and either it's `Rw::Start` or row propagate from previous chunk.

Therefore, introduce selector in lookup so offset=0 will be skip, due to we already set selector=0 when offset=0

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -50,6 +52,16 @@ pub struct EvmCircuitConfig<F> {
     copy_table: CopyTable,
     keccak_table: KeccakTable,
     exp_table: ExpTable,
+
+    // rw permutation config
+    rw_permutation_config: PermutationChipConfig<F>,
+
+    // pi for carry over previous chunk context
+    pi_pre_continuity: Column<Instance>,
+    // pi for carry over previous chunk context
```
ggkitsas, line 61:
```suggestion
    // pi for carry over chunk context to the next chunk
```
Is this more accurate?

bus-mapping/src/exec_trace.rs
ed255: I see, so there's 1 single `Start` row at the top, and a variable number of `Padding` rows at the bottom, right?

zkevm-circuits/src/witness/block.rs
ed255: Sounds good!

bus-mapping/src/exec_trace.rs
hero78119: > I see, so there's 1 single `Start` row at the top, and a variable number of `Padding` rows at the bottom, right?

Exactly!

zkevm-circuits/src/evm_circuit.rs
hero78119: Right! and fixed. Thanks for point out that :)

zkevm-circuits/src/state_circuit.rs
hero78119: You're right! Previously Rwtable continuity is carried over via a separate copy constraints, because the `PermutationChip` is meant to only cover fingerprints related logic. After your reminding, I think rwtable continuity can be realized via fingerprints as well, because first/last row of rwtable against fingerprint are one to one mapping, so we can leverage `PermutationChip` to do everything. 

I have done the change to simplify design in new commit https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1641/commits/de550345bb65f04250c65d6ac779484f29da4a54 so we can save copy contraints effort on rw_table itself.

han0110 (commented):
First round review, overall LGTM!

zkevm-circuits/src/witness/rw.rs
```diff
@@ -120,34 +125,60 @@ impl RwMap {
...
+        let padding_length = {
+            let length = Self::padding_len(rows_trimmed.len(), target_len);
+            if is_first_row_padding {
+                length.saturating_sub(1)
+            } else {
+                length
+            }
+        };
+        let start_padding_rw_counter =
+            rows_trimmed.last().map(|row| row.rw_counter()).unwrap_or(1) + 1;
```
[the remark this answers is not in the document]
hero78119: There is an existing issue (might be no impact) since day one: rw_counter = 1 in rw_table is not unique. In fact, there will be another with non `Start` field tag also got `rw_counter = 1`. In state_circuit constraints, it seems we only emphasise rw_counter unique within same tag, but not across tag.

So the padding start from 1 is ok, which can be treat as special case

gadgets/src/permutation.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// assign
+    pub fn assign(
+        &self,
+        region: &mut Region<'_, F>,
+        alpha: Value<F>,
+        gamma: Value<F>,
+        acc_fingerprints_prev: Value<F>,
+        col_values: &[Vec<Value<F>>],
+    ) -> Result<PermutationAssignedCells<F>, Error> {
+        self.annotate_columns_in_region(region, "state_circuit");
```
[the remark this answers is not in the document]
hero78119: It's some copy/paste error. Fixed by refactor `prefix` in function arguments so caller can specify annotation.

zkevm-circuits/src/witness/rw.rs
```diff
@@ -37,25 +43,25 @@ impl std::ops::Index<OperationRef> for RwMap {
...
+        for (rw_counter_prev, rw_counter_cur) in self
             .0
             .iter()
-            .filter(|(tag, _rs)| !matches!(tag, Target::Start))
+            .filter(|(tag, _rs)| !matches!(tag, Target::Padding) && !matches!(tag, Target::Start))
             .flat_map(|(_tag, rs)| rs)
             .map(|r| r.rw_counter())
             .sorted()
-            .enumerate()
+            .tuples()
```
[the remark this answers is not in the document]
hero78119: Nice catch! Yes it should be overlap so should be tuple_windows 
Fixed and thanks for point out this!

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -799,6 +834,11 @@ impl<F: Field> ExecutionConfig<F> {
                             ExecutionState::EndBlock,
                             vec![ExecutionState::EndTx, ExecutionState::EndBlock],
                         ),
+                        (
+                            "Only EndChunk or BeginChunk can transit to BeginChunk",
+                            ExecutionState::BeginChunk,
+                            vec![ExecutionState::EndChunk, ExecutionState::BeginChunk],
+                        ),
```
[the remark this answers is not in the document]
hero78119: It's some left over due to some unnessesary flexiblity. 
Fixed it for only `BeginChunk` can transit to `BeginChunk`

gadgets/src/permutation.rs
```diff
@@ -0,0 +1,312 @@
...
+    row_fingerprints: Column<Advice>,
+    alpha: Column<Advice>,
+    power_of_gamma: Vec<Column<Advice>>,
+    // selector
+    q_row_non_first: Selector, // 1 between (first, end], exclude first
+    q_row_enable: Selector,    // 1 for all rows (including first)
+
+    _phantom: PhantomData<F>,
+}
+
```
davidnevadoc, line 37: Maybe it would be helpful to have a comment here to indicate the values of each cell. Something like
`alpha`, `gamma`, `previous fingerprint accumulator`, `last fingerprint`.

gadgets/src/permutation.rs
```diff
@@ -0,0 +1,312 @@
...
+                "GADGETS_PermutationChipConfig_alpha".to_string(),
+            ),
+        ]
+        .iter()
+        .cloned()
+        .chain(
+            self.power_of_gamma
+                .iter()
+                .enumerate()
+                .map(|(i, col)| (*col, format!("GADGETS_PermutationChipConfig_gamma_{}", i))),
```
davidnevadoc, line 156: I find it clearer if we start counting on `1` so that the `ith` column represents `gamma^i`.

gadgets/src/permutation.rs
```diff
@@ -0,0 +1,312 @@
...
+            let alpha_cell = region.assign_advice(
+                || format!("alpha at index {}", offset),
+                self.alpha,
+                offset,
+                || alpha,
+            )?;
+            let gamma_cells = self
+                .power_of_gamma
+                .iter()
+                .zip(power_of_gamma.iter())
```
davidnevadoc, line 104: I believe `power_of_gamma` first element is `gamma^0 = 1` so we should skip the first element here.

gadgets/src/permutation.rs
```diff
@@ -0,0 +1,312 @@
...
+    alpha: Column<Advice>,
+    power_of_gamma: Vec<Column<Advice>>,
+    // selector
+    q_row_non_first: Selector, // 1 between (first, end], exclude first
+    q_row_enable: Selector,    // 1 for all rows (including first)
+
+    _phantom: PhantomData<F>,
+}
+
+type PermutationAssignedCells<F> = (
```
davidnevadoc, line 38: Maybe it would be helpful to have a comment here to indicate the values of each cell. Something like
`alpha`, `gamma`, `previous fingerprint accumulator`, `last fingerprint`.

zkevm-circuits/src/witness/rw.rs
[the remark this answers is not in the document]
han0110: I see, thanks for clarifying!

zkevm-circuits/src/state_circuit.rs
```diff
@@ -227,11 +258,12 @@ impl<F: Field> StateCircuitConfig<F> {
                 log::trace!("state circuit assign offset:{} row:{:#?}", offset, row);
             }
 
+            // disable selector on offset 0 since it will be copy constraints by public input
```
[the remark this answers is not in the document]
han0110: Got it, make sense, we could remove the constraint later when we are going to remove `Start`.

han0110 (approved):
LGTM!

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -36,6 +36,8 @@ pub struct CircuitInputStateRef<'a> {
     pub block: &'a mut Block,
     /// Block Context
     pub block_ctx: &'a mut BlockContext,
+    /// Block Context
```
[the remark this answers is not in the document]
hero78119: Fixed :)

gadgets/src/permutation.rs
hero78119: Nice catch!
Fix in latest commit.
Previous this not trigger unittest error due to gamma is set as `1`
I also update hardcode challenge value to assure it can capture this bug, besides use zip_eq to be more error prune

gadgets/src/permutation.rs
hero78119: documentation added

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -8,7 +8,7 @@ use halo2_proofs::{
 use std::collections::HashMap;
 
 // Step dimension
-pub(crate) const STEP_WIDTH: usize = 128;
```
hero78119, line 11: increase 3 column to make `EndBlock` height remain `1`

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -64,7 +64,7 @@ pub const FIXED_TABLE_LOOKUPS: usize = 8;
 pub const TX_TABLE_LOOKUPS: usize = 4;
 
 /// Rw Table lookups done in EVMCircuit
-pub const RW_TABLE_LOOKUPS: usize = 8;
```
hero78119, line 65: increase 5 columns for rw_lookup to make sure `EndChunk` height to be `1`, same as `EndBlock`

> `EndChunk` consume more rw_lookup than `EndBlock` is because `EndChunk` write StepState fields into `rw_table`

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -252,7 +252,13 @@ impl<F: Field> SubCircuitConfig<F> for EvmCircuitConfig<F> {
     }
 }
 
-type AssignedChunkContextCell<F> = (AssignedCell<F, F>, AssignedCell<F, F>, AssignedCell<F, F>);
```
davidnevadoc: Let's add a comment here to see clearly what each of this cells is
`/// chunk_index, chunk_index_next, total_chunk, initial_rwc, end_rwc `

davidnevadoc (commented):
First pass of the review. Still have some parts to go through ( common_gadget and Chunkctx_table ), the rest LGTM.

zkevm-circuits/src/evm_circuit.rs
hero78119: Done :)

hero78119:
will merge it first and left impromemt to future PR

PR #1643, opened 2023-10-04
[MPTWG] Reformat JSON files

ChihChengLiang:
### Issue Link

#1658

### Type of change

Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Contents

- [x] Use proper Go libraries to dump the json files. This reduces the number of serialization/deserialization codes and errors.
- [x] Regenerate JSON files.
- [x] Rewrite the parsing code in MPT circuit.

### How Has This Been Tested?

See if the new JSON format can be consumed by the MPT circuit and run the test successfully.


miha-stopar (commented):
Nice! I started to review this PR, but later I saw it's still draft.
It requires some changes in the circuit as well, like for example in`witness_row.rs` adding
```
#[derive(Clone, Serialize, Deserialize, Debug)]
#[serde(transparent)]
struct HexVal {
    #[serde(with = "hex::serde")]
    hex: Vec<u8>,
}
```
and changing for `Node` struct
```
pub values: Vec<Vec<u8>>,
```
to
```
pub values: Vec<HexVal>,
```
and then changing the assign function of `MainRlpGadget`. Do you plan to do these changes in this PR?

zkevm-circuits/src/mpt_circuit.rs
```diff
@@ -424,7 +424,7 @@ impl<F: Field> MPTConfig<F> {
                         let rlp_value = self.rlp_item.assign(
                             &mut cached_region,
                             offset + idx,
-                            bytes,
+                            bytes, // highlight for comment
                             *item_type,
                         )?;
```
ChihChengLiang, line 429: The `bytes` passed here is `Hex` struct and the assign takes `&[u8]` for this argument.
The compiler is happy about this because we've implement the Deref trait for Hex. The `bytes` here is a Vec<u8> dereffed from the Hex struct.

ChihChengLiang (commented):
Hi @miha-stopar,
Thanks for the early review! The PR is still a draft, but a quick look from you is helpful!
I think the MainRLPGadget can still take `&[u8]` as inputs. The Deref trait on Hex resolved that issue.

KimiWu123 (approved):
I didn't check json files but others looks good to me.

zkevm-circuits/src/mpt_circuit.rs
miha-stopar: Wonderful!

miha-stopar (approved):
LGTM!

Issue #1639, opened 2023-10-03
bug for create gadget


lispc:
### Describe the bug

```
        // Handle the case where an error of ErrDepth, ErrInsufficientBalance,
        // ErrNonceUintOverflow or ErrContractAddressCollision occurred.
        cb.condition(
            or::expr([
                not::expr(is_precheck_ok),
                not::expr(not_address_collision.expr()),
            ]),
```

no they should not share same branch.   at least, when `!is_precheck_ok`, the step state transition for gas_left should be -Delta(gas_cost), while when address_collision, the remaining gas is only 1/64 current gas.

lispc:
btw i think the collision test is not enabled in fact... this should be true.. 

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4f1f56e41e8b1a724ae141a5e7b27d2e19c07353/zkevm-circuits/src/evm_circuit/execution/create.rs#L888

PR #1657, opened 2023-10-12
fix many bugs of create op

lispc:
### Description

i think this pr can fix many errs of testool report

1. fix bugs of https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1639
2. refactor rw reading using the rw iterator, which is far more friendly to understand compared to the current manually maintained rw offset way.
3. another refactor here, is move the "init code copy event" from callee stage to caller stage (before state.push_call). It is more intuitive. Inside execution, we can think the initcode is "copied from CopyDataType::Memory to CopyDataType::Bytecode" then is executed. So doing the copy event inside callee stage is a bit wired.

Btw i have another double on hi-lo (which i am not very familiar with): is this correct? https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/09651d40afc59462fd943e9789be8f8b0ccb90d9/zkevm-circuits/src/evm_circuit/execution/create.rs#L545 is_address_collision is boolean, which is not equal to the expr used to construct the is_zero gadget?

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1639

### Type of change


bus-mapping/src/evm/opcodes/create.rs
```diff
@@ -159,23 +170,15 @@ impl<const IS_CREATE2: bool> Opcode for Create<IS_CREATE2> {
             state.call_context_write(&mut exec_step, caller.call_id, field, value);
         }
 
+        let (initialization_code, code_hash) = if is_precheck_ok && length > 0 {
```
KimiWu123, line 173: I guess here we missed one condition `!is_address_collision`. In circuit, we check `is_precheck_ok && not_address_collision` (https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1657/files#diff-66e8458ad6b55447e17de7f715e4395be943d682483be4a945b51f98a4cc50aeR261)

KimiWu123 (commented):
LGTM. Just some minor questions.

zkevm-circuits/src/evm_circuit/execution/create.rs
```diff
@@ -263,7 +276,42 @@ impl<F: Field, const IS_CREATE2: bool, const S: ExecutionState> ExecutionGadget<
...
+                program_counter: Delta(1.expr()),
+                stack_pointer: Delta(2.expr() + is_create2.expr()),
+                memory_word_size: To(memory_expansion.next_memory_word_size()),
+                gas_left: Delta(-gas_cost.expr()),
+                ..StepStateTransition::default()
+            });
+        });
 
+        // Case2: Normal create call, precheck is ok and no address collision
+        let transfer = cb.condition(
```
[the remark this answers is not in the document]
KimiWu123: I just saw the description, and I guess the reason why is 
>another refactor here, is move the "init code copy event" from callee stage to caller stage (before state.push_call). It is more intuitive. Inside execution, we can think the initcode is "copied from CopyDataType::Memory to CopyDataType::Bytecode" then is executed. So doing the copy event inside callee stage is a bit wired.

Is it right? If so, could you add some comment to explain why? Bcs it's not straightforward to see the same condition but split into two parts.

bus-mapping/src/evm/opcodes/create.rs
lispc: i would suggest do cb.copy_table_lookup only with condition of is_precheck_ok but not `is_precheck_ok && not_address_collision`? The reason is a bit subtle like this:

Backgound: for normal create2 case, (a new contract is deployed), we should constrain the codehash (which is used to calculate the new callee.address). Currently we do this constrain implicitly using CopyTable with Bytecode as dst, and all bytecodes inside byecode circuit have correct keccak lookup.

For the create2 op, if precheck is ok, we need to hash the initcode to calculate the new address. If there is a collision, then in fact no initcode will be executed so for this collision case, only keccak lookup is needed while copy lookup with bytecode is not needed.

So we have 2 solutions here:
(1) do cb.copy_table_lookup only with condition of is_precheck_ok but not `is_precheck_ok && not_address_collision`. This method, code will be simpler. The only drawback is for collision case, the row usage of bytecode circuit is not optimal. (But i think it is generally ok, since collision is not very common).
(2) for happy case (initcode executed), we still do a single copy lookup with bytecode dst. But for "precheck is ok but collision" case, we need to change to (a) an copy lookup with RlcAcc instead of Bytecode as dst + (b) a keccak lookup. This method can save some bytecode circuit row usage, but make code very complex.

I would prefer solution (1). Any ideas on this?

KimiWu123: Solution (1) works for me and it does make the code simpler. If so, I think you need to change the condition from `is_precheck_ok && not_address_collision` to `is_precheck_ok` only in the circuit [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1657/files#diff-66e8458ad6b55447e17de7f715e4395be943d682483be4a945b51f98a4cc50aeR261).

KimiWu123: Besides, the reason behind is a bit subtle, could you add some comment as well?

KimiWu123 (approved):

rrtoledo (approved):
LGTM :)

PR #1662, opened 2023-10-19
fix soundness: block_table tag/index should be fixed column

hero78119:
### Description

block_table `tag`, `index` those 2 value are not only non-constrained but also be advice column, which means the `tag`, `index` can be witness in any value/order and lead to a situation of lookup blocktag under wrong value. E.g. use `timestamp` value to represent `base_fee`. 

Block_table `value` column is constrained by public input so there is no such issue.

Since `block_table` per row are pre-defined, this PR fixed soundness by move `tag`, `index` to fixed column, so value will be encoded in verify key and avoid malicious manipulation.

Folk Scroll previously has similar fixed https://github.com/scroll-tech/zkevm-circuits/commit/d3922e1d2b3188a04b39d90cdacd22f83cb4dde0#diff-632aa6def09f1668ecdf89781dbf23c2c3cf3bc379209b19cce79abd0a1d0bfe

### Issue Link

N/A

### Type of change

### DIFF
--- a/zkevm-circuits/src/table/block_table.rs
+++ b/zkevm-circuits/src/table/block_table.rs
@@ -1,3 +1,5 @@
+use std::iter;
+
 use super::*;
 
 /// Tag to identify the field in a Block Table row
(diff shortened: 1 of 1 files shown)


hero78119:
Hi @lispc since I notice Scroll got similar fix in April `https://github.com/scroll-tech/zkevm-circuits/commit/d3922e1d2b3188a04b39d90cdacd22f83cb4dde0#diff-632aa6def09f1668ecdf89781dbf23c2c3cf3bc379209b19cce79abd0a1d0bfe` May I ask do you have more background context on this?

han0110 (approved):
LGTM! As for the reason scorll's modification keeps index as advice column, I guess it's because they support multiple block in a single proof, so prover could assign the index up to its wish.

ChihChengLiang (approved):
LGTM! Nice catch

PR #1672, opened 2023-10-25
opt: dedup inputs for keccak circuit

lispc:
### Description

this is a minor PR to remove duplicate entries for keccak circuit.

### Issue Link

[_link issue here_]

### Type of change

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -35,7 +35,10 @@ pub use execution::{
 pub use input_state_ref::CircuitInputStateRef;
 use itertools::Itertools;
 use log::warn;
-use std::{collections::HashMap, ops::Deref};
+use std::{
+    collections::{HashMap, HashSet},
+    ops::Deref,
+};
 pub use transaction::{Transaction, TransactionContext};
 pub use withdrawal::{Withdrawal, WithdrawalContext};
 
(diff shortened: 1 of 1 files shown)


rrtoledo (approved):
LGTM! :)

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -442,19 +445,23 @@ impl CircuitInputBuilder<DynamicCParams> {
 /// Return all the keccak inputs used during the processing of the current
 /// block.
 pub fn keccak_inputs(block: &Block, code_db: &CodeDB) -> Result<Vec<Vec<u8>>, Error> {
-    let mut keccak_inputs = Vec::new();
+    let mut keccak_inputs: HashSet<Vec<u8>> = HashSet::new();
```
hero78119, line 448: How about also adding debug log when detecting duplicate input?  It make thing easy to trace and fix root cause if have.

hero78119 (commented):
Just one advice for debugging improvement, other LGTM :)

bus-mapping/src/circuit_input_builder.rs
lispc: i think "duplicate input" is not something wrong.  in sha3 opcode, in sign verify, in mpt trie, it would be common to produce some duplicate input

hero78119: got it, thanks for clarify!

hero78119 (approved):
LGTM!

PR #1676, opened 2023-10-26
fix return/revert offset/len in handle_return for err cases

lispc:
### Description

geth_step.error is not enough. It only contains error inside geth trace, only stack err and oog err. Exec_step.error contains all types of err we recovered during parsing trace.

This pr will fix some failures of testool, like `CreateOOGFromCallRefunds_d1(SStore_Refund_OoG)_g0_v0`

### Issue Link

[_link issue here_]

### Type of change


ChihChengLiang (approved):
LGTM

Issue #1611, opened 2023-09-19
Support invalid tx


KimiWu123:
### Describe the feature you would like

A follow-up issue of #1503 

----

### Description
Support skipping the following invalid transactions:
- nonce too low
- nonce too high
- intrinsic gas too low
- insufficient funds for gas * price + value

related issue: https://github.com/taikoxyz/zkevm-circuits/issues/70

### Contents
- Check if a TX is invalid by adding a few auxiliary columns to the TX table.
- The related PR in [zkevm-specs](https://github.com/privacy-scaling-explorations/zkevm-specs) is https://github.com/privacy-scaling-explorations/zkevm-specs/pull/338

ed255:
@xiaodino would you like to get this task assigned to you?

CeciliaZ030:
> @xiaodino would you like to get this task assigned to you?

Hi @ed255, I'm currently doing this :) 
@xiaodino has been working on taiko's bridge these days, we reverted his PR from our end and here's a new draft PR https://github.com/taikoxyz/zkevm-circuits/pull/157
Also I ended up adding a new `ExecStep` rather than error state since invalid Tx shouldn't even make it to execution, and hence shouldn't be thrown in error. A separate state that skips them seems more neat I guess.

PR #1677, opened 2023-10-29
Add basic invalid tx support

Brechtpd:
### Description

Resolves https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1611
Upstreamed from https://github.com/taikoxyz/zkevm-circuits/pull/157 with some additional changes.

Adds an additional execution state for transactions that are invalid and need to be skipped. This execution state needs to be enabled instead of the normal BeginTx/EndTx for valid transactions.

Support skipping the following invalid transactions for now:
- nonce too low
- nonce too high
- intrinsic gas too low
- insufficient funds for gas * price + value

More cases will be added in future PRs.

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1611

### Type of change

### Contents

- [_item_]

### Rationale

### How Has This Been Tested?

- Added required unit tests
- make test-all passes


bus-mapping/src/evm/opcodes/begin_end_tx.rs
```diff
@@ -313,17 +314,62 @@ fn gen_end_tx_steps(state: &mut CircuitInputStateRef) -> Result<ExecStep, Error>
...
+    Ok(())
+}
+
+pub(crate) fn end_tx(
+    state: &mut CircuitInputStateRef,
+    exec_step: &mut ExecStep,
+    call: &Call,
+) -> Result<(), Error> {
+    // Write the tx receipt
+    write_tx_receipt(state, exec_step, call.is_persistent)?;
```
CeciliaZ030, line 343: Should we write a tx receipt if it's invalid?

mock/src/transaction.rs
```diff
@@ -342,4 +346,18 @@ impl MockTransaction {
...
+
+    /// Mark the transaction as invalid.
+    pub fn invalid(&mut self) -> &mut Self {
+        self.invalid = true;
+        self
+    }
+
+    /// Force the nonce field for the MockTransaction.
+    /// Can only be used for invalid transactions.
+    pub fn force_nonce(&mut self, nonce: u64) -> &mut Self {
```
CeciliaZ030, line 358: More intuitive to use?
```suggestion
    pub fn set_nonce(&mut self, nonce: u64) -> &mut Self {
```

CeciliaZ030 (commented):
LGTM, cleaner than my original ver :) Maybe util/tx.rs can be renamed to util/tx_gadget.rs to be consistent with the other helper gadgets files?

bus-mapping/src/evm/opcodes/begin_end_tx.rs
Brechtpd: I write the receipt because `end_tx` expects them to be there for the previous transaction:

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/7617394711a0e79beec5ef1f792938242be3dbcb/zkevm-circuits/src/evm_circuit/execution/end_tx.rs#L304-L310

So if they are not there in the same format like with valid transactions, I think it would create a messy situation between valid and invalid transactions.

It's used to track the `CumulativeGasUsed` over multiple transactions, so we still have to do that at least for invalid transactions as well (though no update to the value because no gas is used).

mock/src/transaction.rs
Brechtpd: I called it `force_nonce` to make it more apparent that it normally shouldn't be used because the nonce is set automatically when you need the nonce to be valid.

KimiWu123 (approved):
LGTM! It looks nice and clear.

KimiWu123:
> LGTM, cleaner than my original ver :) Maybe util/tx.rs can be renamed to util/tx_gadget.rs to be consistent with the other helper gadgets files?

@CeciliaZ030 , could you review if your comments were well-addressed?

bus-mapping/src/circuit_input_builder/block.rs
```diff
@@ -23,6 +23,8 @@ pub struct BlockContext {
     pub(crate) call_map: HashMap<usize, (usize, usize)>,
     /// Total gas used by previous transactions in this block.
     pub(crate) cumulative_gas_used: u64,
+    /// Number of invalid transactions
+    pub(crate) num_invalid_tx: usize,
```
hero78119, line 27: Didn't find the place derive logic from this new defined field. is it left over?

mock/src/transaction.rs
```diff
@@ -160,6 +162,8 @@ impl Default for MockTransaction {
             max_priority_fee_per_gas: Word::zero(),
             max_fee_per_gas: Word::zero(),
             chain_id: *MOCK_CHAIN_ID,
+            invalid: false,
+            forced_nonce: None,
```
hero78119, line 166: It seems to me that new defined `force_nonce` is a bit confused.

Can we just manipulate original `nonce` value by new setter `set_nonce` when we are unittest invalid transaction? ( I believe there must be some obstacle so we need to introduce it)

bus-mapping/src/circuit_input_builder/transaction.rs
```diff
@@ -183,6 +183,8 @@ pub struct Transaction {
     pub id: u64,
     /// The raw transaction fields
     tx: geth_types::Transaction,
+    /// Invalid tx
+    pub invalid: bool,
```
hero78119, line 187: A optimisation suggestion: maybe do not need to define new field `invalid`, because at the place we need to derived logic based on `tx.invalid`  can just use `geth_trace.invalid` instead ?

hero78119 (changes requested):
LGTM! Great work to introduce new virtual step to separate invalid tx logic!

I left few comments related to typo and code cosmetics, please kindly check :)

bus-mapping/src/circuit_input_builder/block.rs
Brechtpd: Yes, left it in but indeed better to remove for now.

mock/src/transaction.rs
Brechtpd: The reason I introduced the new variable is because I didn't want to change the existing `nonce` value, but it turns out it's actually a small change, so just made `nonce` an option. Also renamed to `set_nonce` because it seems people like that name better.

external-tracer/src/lib.rs
```diff
@@ -70,6 +70,18 @@ pub fn trace(config: &TraceConfig) -> Result<Vec<GethExecTrace>, Error> {
...
 
-    let trace = serde_json::from_str(&trace_string).map_err(Error::SerdeError)?;
+    let trace: Vec<GethExecTrace> =
+        serde_json::from_str(&trace_string).map_err(Error::SerdeError)?;
+    // Don't throw only for specific invalid transactions we support.
+    for trace in trace.iter() {
+        let error = &trace.return_value;
+        let allowed_cases = error.starts_with("nonce too low")
+            || error.starts_with("nonce too high")
+            || trace.return_value.starts_with("intrinsic gas too low")
```
hero78119, line 80: some left over :))

hero78119 (approved):
Thanks for addressing all the issues so promptly!
LGTM and great job

external-tracer/src/lib.rs
Brechtpd: How did I miss that...

PR #1684, opened 2023-11-15
simplify pi input for proof chunk continity to single column

hero78119:
### Description

simplify public input to single instance column.
in transcript there is no public input commitment, so we can constraint their continuity via raw field value in aggregation circuit

### Type of change


hero78119:
Will be covered in PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1693

Issue #1603, opened 2023-09-14
[proof-chunk] root circuit constraints permutation fingerprint and challenges


hero78119:
Depends on https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1601

### Permutation Fingerprint
Validation of permutation fingerprint (field equality) is on last row of last chunk, imply the equality check gate of rwtable fingerprints should be able to toggle on and off as an option. In root circuit, there need an extra protocol to constraint the last chunk permutation fingerprint need to be on, while the other should be off. 

### Validation of Challenge 
Need to encode poseidon hash statement as protocol after https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1602. and challenge is computed via Poseidon hash with input `commit(m1) || commit(m2) || ...` where commit mN defined as Nth column commitment of rwtable. Commitment of columns belongs to 2 rwtables should all being include to derive global challenge. 

Furthermore, there need an extra protocol to constraint challenge output from poseidon matching with global challenge pass in supercircuit public input.

### Changed Item

New public input in supercircuit
- `LHS` & `RHS` permutation fingerprint 
- `challenge` of permutation fingerprint. We need 2 challenges: alpha & gamma. 
   > $\prod_{i=0}^{m-1} (\alpha - (cell_i[0] + \gamma \times cell_i[1] + \gamma^2 \times cell_i[2] + ...))$

- `is_last_chunk` to enable public input

>it seems `is_last_chunk` is possible to be hints from non-deterministic advice instead of public input. However it involve more ad-hoc change to commit a non-deterministic advice on the fly outside of supercircuit circuit, which involved more hyper-parameter to support this workaround. @han0110 any opinion on that ~?

Extra protocols
- `poseidon hash circuit` to derive challenge from rwtable commitment
-  constraint `is_last_chunk=true` in last chunk
-  constraint `computed_challenge == challenge` in last chunk

hero78119:
Hi @han0110 and @ed255 would you kindly help to review the issue and provide more feedback :) ?

ed255:
Overall sounds good to me!  Here are some notes of things I found missing:
- Each memory operation requires several field elements.  One solution for the fingerprinting is to first do RLC of each element, and then accumulate the result in the fingerprint.  For this we need 2 challenges (alpha and gamma).  See:

[...]

PR #1693, opened 2023-11-29
[proof chunk] root circuit consistency between chunk

hero78119:
### Description

covered item
- [x] supercircuit simplify instance columns related to chunkctx to just one column
- [x] first chunk instance: chunk continuity related fields should be constraints as constant.
- [x] aggregate multiple super circuit proof chunk with chunk consistency check .
- [x] compute permutation fingerprint challenges `alpha,gamma` from rw_table advices commitments and assert its equality
- [x] generalized `challenge([column_indexs], challenge_column_index)` to a new protocol structure so more challenges in the future can be added. 

TODO
- [ ] verify multiple chunk from bus_mapping => rely on https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1690
 
### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1603

### Type of change

### Tests

Light tests pass.
integration test failed due to challenge computation in witness not implemented yet, which will be done in bus mapping chunk PR.

### DIFF
--- a/zkevm-circuits/src/root_circuit.rs
+++ b/zkevm-circuits/src/root_circuit.rs
@@ -8,7 +8,7 @@ use halo2_proofs::{
     poly::{commitment::ParamsProver, kzg::commitment::ParamsKZG},
 };
 use itertools::Itertools;
-use maingate::MainGateInstructions;
+use maingate::{MainGateInstructions, RegionCtx};
 
 use snark_verifier::{
     pcs::{
@@ -27,6 +27,7 @@ mod aggregation;
 mod dev;
 #[cfg(test)]
 mod test;
+
 #[cfg(feature = "test-circuits")]
 pub use self::RootCircuit as TestRootCircuit;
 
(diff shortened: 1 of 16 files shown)


hero78119:
Hi @CeciliaZ030 this PR is for root circuit proof chunk poseidon challenges circuit & chunk consistency check.
Will be appreciate if you can also involve for reviewing, tks!

zkevm-circuits/src/witness/block.rs
```diff
@@ -317,29 +313,43 @@ pub fn block_convert<F: Field>(
...
     Ok(block)
 }
+
+fn get_rwtable_fingerprints<F: Field>(
+    alpha: F,
+    gamma: F,
+    prev_continuous_fingerprint: F,
+    rows: &Vec<Rw>,
+) -> RwTablePermutationFingerprints<F> {
+    // &<dyn ToVec<Value<F>>>::to2dvec(&chronological_rws_rows)
```
han0110, line 333: leftover

zkevm-circuits/src/root_circuit.rs
```diff
@@ -170,18 +244,183 @@ where
...
+                                .assert_equal(&mut loader.ctx_mut(), a, b)
+                                .unwrap();
+                        });
+
+                        first_chunk
+                    })
+                    .expect("error");
+
+                // constraint consistency between chunk
+                let _ = supercircuit_instances.iter().tuple_windows().inspect(
```
han0110, line 338: `inspect` is also lazy so the closure is not called afais, so perhaps using `for_each` here?

zkevm-circuits/src/super_circuit.rs
```diff
@@ -250,6 +264,31 @@ impl<F: Field> SubCircuitConfig<F> for SuperCircuitConfig<F> {
             },
         );
 
+        // constraint chronological/by address rwtable row fingerprint must be the same in first
+        // chunk first row.
+        meta.create_gate(
+            "chronological rwtable row fingerprint == by address rwtable row fingerprint",
```
han0110, line 270: This seems already guaranteed by the `RootCircuit` right (by initializing them both as `1`)?

han0110 (commented):
Overall LGTM!

zkevm-circuits/src/super_circuit.rs
hero78119: There are 2 terms `row_fingerprint` and `acc_fingerprint`

- acc_fingerprint: start from 1
- row_fingerprint: row level rlc value by `alpha/gamma` challenge

by current design: `row_fingerprint(0) v acc_fingerprint_rotation(-1) = acc_fingerprint_rotation(-1)`

Root circuit only guarantee `acc_fingerprint` at offset 0 as `1`. However, root circuit do not know what the first `row_fingerprints` looks like (In our case, first row first chunk in rw_table is `Rw::Start`). This value can only being known in super circuit thats's why this gates is added. 

Besides, `BeginChunk` virtual step will also lookup `Rw::Start`. So

- by address rw table first row must be Start -> guaranteed by BeginChunk lookup 
- chronologically rw_table first row must equal with state_circuit -> guaranteed by this new gate

Thanks for the reminding and I will add more code comments to explain this :)

zkevm-circuits/src/root_circuit.rs
hero78119: Thanks for point out this. Fixed by `for_each` which is more semantics correct

zkevm-circuits/src/super_circuit.rs
han0110: I see I see, now I get it, thanks for clarifying!

han0110 (approved):

zkevm-circuits/src/root_circuit.rs
```diff
@@ -70,15 +139,22 @@ where
 {
     /// Create a `RootCircuit` with accumulator computed given a `SuperCircuit`
     /// proof and its instance. Returns `None` if given proof is invalid.
+    /// TODO support multiple snark proof aggregation
     pub fn new(
         params: &ParamsKZG<M>,
         super_circuit_protocol: &'a PlonkProtocol<M::G1Affine>,
```
CeciliaZ030, line 145: In the actual aggregation situation where we're dealing with multiple chunks, should `protocol`, `instances`, and `proof` all be Vec? So that the `flatten_super_circuit_instances` will actually contain concatenated instances from different chunks.

zkevm-circuits/src/witness/rw.rs
```diff
@@ -1071,3 +1071,28 @@ impl From<&operation::OperationContainer> for RwMap {
         Self(rws)
     }
 }
+
+/// RwTablePermutationFingerprints
+#[derive(Debug, Default, Clone)]
+pub struct RwTablePermutationFingerprints<F> {
```
CeciliaZ030, line 1077: In #1690, I actually make a `Chunk<F>` that holds all chunk-related data, so l guess if this struct is incorporated in `Chunk<F>` then it wouldn't be necessary anymore? 
For now it makes total sense so just keep it, just a heads up I'll probably get rid of it when I merge :)

CeciliaZ030 (approved):
LGTM! The`UserChallenge` is very nice. I already started syncing this branch to #1690 and it doesn't look too hard to merge :) the RootCircuit & SuperCircuit are isolated and the chunk-related data I'll just manually move them to `witness::Chunk<F>`.
The only work integrate bus-mapping's witness generation to`SuperCircuitInstance<T>` which should be easy.

Issue #1700, opened 2023-12-05
Non-deterministic constraint generation in memory.rs


adria0:
Since `self.banks` is declared as HashMap

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/e246d4dbf30ccf8369fc455422129b59137e219a/zkevm-circuits/src/circuit_tools/memory.rs#L22-L25

and in constraint generation is iterated

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/e246d4dbf30ccf8369fc455422129b59137e219a/zkevm-circuits/src/circuit_tools/memory.rs#L77-L85

thus, since HashMap iteration is not deterministic, constraint generation is not also.

Must be changed with something like `BTreeMap`

ed255:
@zemse is currently working on this

PR #1706, opened 2023-12-08
fix nondeterministic constraint generation

zemse:
### Description

Change HashMap to BTreeMap in memory.rs and in cell_placement_strategy.

### Issue Link

#1700

### Type of change

### Contents

- fix in memory.rs
- fix in cell_placement_strategy

### Rationale

Generates different circuit keys making proof verification fail for different witnesses.

### How Has This Been Tested?

Manual testing.

### DIFF
--- a/zkevm-circuits/src/circuit_tools/memory.rs
+++ b/zkevm-circuits/src/circuit_tools/memory.rs
@@ -7,7 +7,7 @@ use halo2_proofs::{
     poly::Rotation,
 };
 use std::{
-    collections::HashMap,
+    collections::BTreeMap,
     marker::PhantomData,
     ops::{Index, IndexMut},
 };
@@ -20,7 +20,7 @@ use super::{
 
 #[derive(Clone, Debug, Default)]
 pub(crate) struct Memory<F: Field, C: CellType, MB: MemoryBank<F, C>> {
-    banks: HashMap<C, MB>,
+    banks: BTreeMap<C, MB>,
     _phantom: PhantomData<F>,
     tag_counter: usize,
 }
@@ -42,7 +42,7 @@ impl<F: Field, C: CellType, MB: MemoryBank<F, C>> IndexMut<C> for Memory<F, C, M
 impl<F: Field, C: CellType, MB: MemoryBank<F, C>> Memory<F, C, MB> {
     pub(crate) fn new() -> Self {
         Self {
-            banks: HashMap::new(),
+            banks: BTreeMap::new(),
             _phantom: PhantomData,
             tag_counter: 0,
         }
--- a/zkevm-circuits/src/mpt_circuit.rs
+++ b/zkevm-circuits/src/mpt_circuit.rs
@@ -766,10 +766,37 @@ pub fn load_proof_from_file(path: &str) -> Vec<Node> {
 mod tests {
     use super::*;
     use halo2_proofs::{dev::MockProver, halo2curves::bn256::Fr};
-    use std::{fs, ops::Deref};
+    use std::{fs, ops::Deref, path::PathBuf};
 
     #[test]
     fn test_mpt() {
+        let degree = 15;
+        get_witnesses()
+            .enumerate()
+            .for_each(|(idx, (path, num_rows, circuit))| {
+                println!("{} {:?}", idx, path);
+                let prover = MockProver::<Fr>::run(degree, &circuit, vec![]).unwrap();
+                assert_eq!(prover.verify_at_rows(0..num_rows, 0..num_rows,), Ok(()));
+                // assert_eq!(prover.verify_par(), Ok(()));
+                // prover.assert_satisfied();
+            });
+    }
+
+    #[test]
+    fn variadic_size_check() {
+        let mut circuits = get_witnesses();
+        let first = circuits.next().unwrap().2;
+        let second = circuits.next().unwrap().2;
+
+        let degree = 15;
+        let prover_1 = MockProver::<Fr>::run(degree, &first, vec![]).unwrap();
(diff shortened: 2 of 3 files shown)


hero78119 (commented):
LGTM and nice catch! Thanks for PR.

Btw, did you catch the issue when testing `mpt_circuit`?  because I didn't find the place to call `Memory utils` other than mpt_circuit.

It will be nice if you can help add unittest to verify `prover_key` are consistent, which is a bit similar as what [evm_circuit.rs](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit.rs#L484-L527) does

zemse:
> LGTM and nice catch! Thanks for PR.
> Btw, did you catch the issue when testing mpt_circuit? because I didn't find the place to call Memory utils other than mpt_circuit.

The memory.rs issue I did not catch it, it is from the github issue #1700 by adria0.

> It will be nice if you can help add unittest to verify prover_key are consistent, which is a bit similar as what [evm_circuit.rs](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/main/zkevm-circuits/src/evm_circuit.rs#L484-L527) does

I am planning to work on this today

zemse:
Currently `cargo test mpt_circuit::tests::variadic_size_check` fails due to `assert_eq!(prover_1.fixed(), prover_2.fixed());` (I've checked that permutations comparison pass). It appears there are just 32 fixed cells (out of over 100,000 of them) that differ. In prover_1 they are assigned with `Fr(0)` while in prover_2 these remain unassigned. 

I don't know where it is coming from. I can imagine adding similar tests for smaller components to nail down the issue. But any suggestions on how this can be approached?

ed255:
> Currently `cargo test mpt_circuit::tests::variadic_size_check` fails due to `assert_eq!(prover_1.fixed(), prover_2.fixed());` (I've checked that permutations comparison pass). It appears there are just 32 fixed cells (out of over 100,000 of them) that differ. In prover_1 they are assigned with `Fr(0)` while in prover_2 these remain unassigned.
> 
> I don't know where it is coming from. I can imagine adding similar tests for smaller components to nail down the issue. But any suggestions on how this can be approached?
> 
> Steps to reproduce:
> 
> ```
> cd zkevm-circuits
> cargo test mpt_circuit::tests::variadic_size_check 
> ```

I have reviewed this error and here is some feedback:
- The variadic test introduced in this PR only runs the MPTCircuit with 2 tests inputs (out of 86 different inputs that can be found in the `mpt_circuit/tests` folder.  So it can be the case that we find pairs of inputs that lead to the same circuit and pairs of inputs that lead to different circuits.  Since we have all these test inputs, I think we should test them all, and make sure they all generate the same circuit
- While picking 2 test inputs, the `fs::read_dir` function is used and the first 2 results are taken.  This result is not deterministic because the output order of reading entries from a directory is platform and filesystem dependent.  So I suggest this list is sorted by filename first to get deterministic order and thus deterministic test results.
- I've investigated the issue.  It fails with these two test inputs: `src/mpt_circuit/tests/OnlyLeafInStorageProof.json` `src/mpt_circuit/tests/FromNilToValue.json` (possibly it also fails for others).
  - The affected columns are `q_enable`, `q_first` and `q_last`.

I've taken a look at the assignment code for the three affected fixed columns and found the issue.  This is the assignment:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/d97258872bbe14547465b984343b1e359217281b/zkevm-circuits/src/mpt_circuit.rs#L507-L511

The assigned rows depend on `height`, and `height` depends on the number of nodes used in the input: https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/d97258872bbe14547465b984343b1e359217281b/zkevm-circuits/src/mpt_circuit.rs#L500
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/d97258872bbe14547465b984343b1e359217281b/zkevm-circuits/src/mpt_circuit.rs#L404

Since the number of nodes depends on the input MPTProofs, different inputs lead to different fixed column assignments.  I think a possible solution is:
1. Introduce a new parameter which is the number of nodes that the MPT circuit can verify: `max_nodes`
2. Pad the vector of nodes with dummy nodes (these are, nodes that by themselves are valid, but we don't care about) until we reach the `max_nodes`
3. If the length of input nodes is bigger than `max_nodes`, panic.

@miha-stopar what do you think about this possible solution?

ed255:
Miha solved the variadic fixed column assignment in the MPT Circuit in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1745
@zemse could you rebase this PR and check if the test now passes?

ed255:
There is one feedback that I had for this PR that has not yet been addressed:
> While picking 2 test inputs, the fs::read_dir function is used and the first 2 results are taken. This result is not deterministic because the output order of reading entries from a directory is platform and filesystem dependent. So I suggest this list is sorted by filename first to get deterministic order and thus deterministic test results.

I think it would be nice to fix that before merging so that we don't encounter surprises in the future, where the test fails in CI but passes locally and we spend unnecessary time figuring out why.

zemse:
Sorry for not addressing it earlier. I have pushed a commit to sort the files. Can you please review it again?

ed255 (approved):
LGTM! Thanks for addressing the feedback!  Now this PR has 2 approvals, so after it's rebased to main it will be ready to merge :)

zemse:
> Update: I'm not 100% sure if rebasing to main is needed

Since we are squash merging and there are no conflicts, I don't think we need rebase. I'll try clicking on the "merge when ready".

PR #1709, opened 2023-12-14
[proof chunk] WIP: rw_table witness commitment to derive challenge

hero78119:
### Contents

- introduce `get_rwtable_cols_commitment` which will be used to get rw_table column witnesses commitment 
- [temporarily] hardcode challenge value in aggregation circuit and remove hardcode later

### Other notes
Need to upgrade  PSE-halo2 api [empty_lagrange_assigned](https://github.com/privacy-scaling-explorations/halo2/blob/9b33f9ce524dbb9133fc8b9638b2afd0571659a8/halo2_proofs/src/poly/domain.rs#L189-L194
) visibility to public.

### Design rationale
Implement commitment follow halo2 `create_proof` to collect witness and commit

### DIFF
--- a/zkevm-circuits/src/root_circuit.rs
+++ b/zkevm-circuits/src/root_circuit.rs
@@ -280,7 +280,7 @@ where
                     config.aggregate::<M, As>(ctx, &key.clone(), &self.snark_witnesses)?;
 
                 // aggregate user challenge for rwtable permutation challenge
-                let (alpha, gamma) = {
+                let (_alpha, _gamma) = {
                     let mut challenges = config.aggregate_user_challenges::<M, As>(
                         loader.clone(),
                         self.user_challenges,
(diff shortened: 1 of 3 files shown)


hero78119:
Hi @CeciliaZ030 this PR introduce new api `get_rwtable_cols_commitment` so we can derive unblinded advice column commitment in the same way to compute permutation challenge in bus-mapping, FYI

CeciliaZ030:
I wonder if there's an easier way to do this other than instantiating a new circuit in `get_rwtable_cols_commitment`, for example changing the existing Challenge API to squeeze challenge from specified `unblinded_advice_column` during the assignment phase? In that way the whole design will change 😅 but one solution could be uniting the Challenge and UserChallenge to one big thing?
I don't know I'm just brainstorming.

hero78119:
Covered in PR https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1773/

PR #1718, opened 2023-12-19
[proof chunk] introduce Padding virtual step

hero78119:
### Description

Introduce this virtual step to carry transition context, which do not incur any extra rw_table lookup.
This also eliminate ambiguous `end_block_non_last` with unified padding virtual step.

### Design Rationale

Why introduce Padding
- Currently we use EndBlock as padding, which still append few unnecessary rw_table lookup, so it make slight hard to control `evm_row` and `max_rws` separate and independently.
- EndBlock as padding only work in single chunk. When comes to multiple chunk, we will have the case that we need to padding execution steps. Padding as [...EndBlock, EndBlock, ... EndChunk] is semantically unpreferred. Another option is [... EndChunk,... EndChunk, EndChunk] however EndChunk introduce > 10 rw_table lookup, so padding like this will easily to hit  `max_rws` bound.

Therefore, introduce virtual step `Padding` to solve it.

#### Padding then following by `EndChunk`
```
[
...
Padding,
Padding,
EndChunk
]
```

#### Padding then following by `EndBlock`
```
[
...
Padding,
Padding,
EndBlock
]
```

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -371,20 +371,27 @@ impl CircuitInputBuilder<FixedCParams> {
             self.set_end_chunk(max_rws);
             return;
         }
+        // set padding
+        let last_step = self
+            .block
+            .txs()
+            .last()
+            .map(|tx| tx.last_step())
+            .unwrap_or_else(|| &self.block.block_steps.padding);
+        let mut padding = last_step.clone(); // padding step need to keep transition context
+        padding.bus_mapping_instance = vec![]; // there is no rw in padding step
+        padding.exec_state = ExecState::Padding;
+        padding.rwc = self.block_ctx.rwc;
+        padding.rwc_inner_chunk = self
+            .chunk_ctx
+            .as_ref()
+            .map_or_else(RWCounter::new, |chunk_ctx| chunk_ctx.rwc);
+        // TODO automatially assign all field
 
         // set end block
-        let mut end_block_not_last = self.block.block_steps.end_block_not_last.clone();
-        let mut end_block_last = self.block.block_steps.end_block_last.clone();
-        end_block_not_last.rwc = self.block_ctx.rwc;
-        end_block_last.rwc = self.block_ctx.rwc;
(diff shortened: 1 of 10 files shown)


bus-mapping/src/circuit_input_builder.rs
```diff
@@ -371,20 +371,27 @@ impl CircuitInputBuilder<FixedCParams> {
...
-            .map_or_else(RWCounter::new, |chunk_ctx| chunk_ctx.rwc);
-        end_block_last.rwc_inner_chunk = self
+        // set padding
+        let last_step = self
+            .block
+            .txs()
+            .last()
+            .map(|tx| tx.last_step())
+            .unwrap_or_else(|| &self.block.block_steps.padding);
+        let mut padding = last_step.clone(); // padding step need to keep transition context
```
hero78119, line 381: Heads up: here use last_step because it's at the end of block.
In chunking, we should use next_tx.first_step instead

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -421,7 +428,7 @@ impl CircuitInputBuilder<FixedCParams> {
         if is_first_chunk {
             push_op(
                 &mut state.block.container,
-                &mut end_block_last,
+                &mut end_block,
```
CeciliaZ030, line 431: why don't we just move `StartOp` and `PaddingOp` here in to the padding `ExecStep`?

bus-mapping/src/circuit_input_builder.rs
CeciliaZ030: Just to make sure `this_tx.last_step` has the same transition witness as `next_tx.first_step` right?

hero78119: No they are different. I thought we discussed this issue in Pr #1690 review section.

hero78119:
This PR was included in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1690

hero78119:
included in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1690

Issue #1707, opened 2023-12-11
soundness: begin_tx tx_id should be read?


lispc:
### Describe the bug

here https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/66788d7902e35d93d07580b6d917a0d922e0a173/zkevm-circuits/src/evm_circuit/util/tx.rs#L42 the tx_id is written into CallContext, but this cell is not constrained.   I know in EndTxGadget, tx_id is increased by one.    I think in begin_tx, an evil attacker can still assign an invalid value to BeginTxHelperGadget.tx_id?   The call_context_lookup_write of TxId inside EndTx, and the call_context_lookup_write of TxId inside BeginTx, are **2 different rws**, so the former cannot constrain the latter?  Unless we modify codes and deliberately make them **share same rwc so they can be same identical rw**?

hero78119:
Hi, Just have a quick look, In end_tx.rs

https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/66788d7902e35d93d07580b6d917a0d922e0a173/zkevm-circuits/src/evm_circuit/util/tx.rs#L127-L131

it constrain next TxId should be current TxId + 1 with under next state (cb.**next**.state) rw_counter  `cb.next.state.rw_counter`. So end_tx and next begin_tx already share the same *rwc* ?

Besides,  rw_counter should be unique under same tag 
> https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/ef143529304544a18416aa96ba0c6dfa9872a77b/zkevm-circuits/src/state_circuit/lexicographic_ordering.rs#L128-L134

It seems to me there is no soundness issue, cmiiw :)

lispc:
> Hi, Just have a quick look, In end_tx.rs
> 
> https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/66788d7902e35d93d07580b6d917a0d922e0a173/zkevm-circuits/src/evm_circuit/util/tx.rs#L127-L131
> 
> it constrain next TxId should be current TxId + 1 with under next state (cb.**next**.state) rw_counter `cb.next.state.rw_counter`. So end_tx and next begin_tx already share the same _rwc_ ?
> 
> Besides, rw_counter should be unique under same tag
> 
[...]

PR #1719, opened 2023-12-19
fix soundness: share txid write lookup across step

hero78119:
### Description

This PR fix soundness by sharing txid write lookup between current step and next step

### Issue Link

#1707 

### Type of change

### Design rationale
Before fix, `cur_end_tx -> callcontext(txid, WRITE)` vs `next_begin_tx -> callcontext(txid, WRITE)` are meant to constraint `next_tx_id = cur_tx_id + 1`. However, because 2 callcontext write are different records in rw_table, so actually second record are under-constraint and tx_id can be manipulated. 

Because so far we do not have negative test framework. So this PR just simply adding a simple positive test to assure there is no consecutive tx_id write with SAME tx_id.

If this fix make sense, will update zkevm-specs accordingly https://github.com/privacy-scaling-explorations/zkevm-specs/blob/master/src/zkevm_specs/evm_circuit/execution/end_tx.py#L71-L79

### DIFF
--- a/bus-mapping/src/evm/opcodes/begin_end_tx.rs
+++ b/bus-mapping/src/evm/opcodes/begin_end_tx.rs
@@ -343,10 +343,6 @@ pub(crate) fn end_tx(
     // Write the tx receipt
     write_tx_receipt(state, exec_step, call.is_persistent)?;
 
-    // Write the next transaction id if we're not at the last tx
-    if !state.tx_ctx.is_last_tx() {
-        // do nothing. as the txid lookup will share the same rwc and txid with next step
-    }
     Ok(())
 }
 
--- a/bus-mapping/src/evm/opcodes/begin_end_tx.rs
+++ b/bus-mapping/src/evm/opcodes/begin_end_tx.rs
@@ -345,12 +345,7 @@ pub(crate) fn end_tx(
 
     // Write the next transaction id if we're not at the last tx
     if !state.tx_ctx.is_last_tx() {
-        state.call_context_write(
-            exec_step,
-            state.block_ctx.rwc.0 + 1,
-            CallContextField::TxId,
-            (state.tx_ctx.id() + 1).into(),
-        )?;
+        // do nothing. as the txid lookup will share the same rwc and txid with next step
     }
     Ok(())
 }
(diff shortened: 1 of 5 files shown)


bus-mapping/src/evm/opcodes/begin_end_tx.rs
```diff
@@ -345,12 +345,7 @@ pub(crate) fn end_tx(
 
     // Write the next transaction id if we're not at the last tx
     if !state.tx_ctx.is_last_tx() {
```
ChihChengLiang, line 347: We can get rid of this if block since the body is empty now.
```suggestion
```

ChihChengLiang (approved):
LGTM. The approach looks good.
I added a nitpick.

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
```diff
@@ -1132,6 +1132,30 @@ impl<'a, F: Field> EVMConstraintBuilder<'a, F> {
         );
     }
 
+    pub(crate) fn call_context_lookup_write_with_counter(
```
KimiWu123, line 1135: suggest to add comment to explain that rwc doesn't need  +1. Like your comment [here](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1719/files#diff-85bb226621e44af9ca0cd097278bcc4aa9c3aeaa80de587a962ed99196bbf7dcR136)

KimiWu123 (approved):
LTGM, only one minor suggestion.

bus-mapping/src/evm/opcodes/begin_end_tx.rs
hero78119: 🙆‍♂️ remove the whole block entirely

zkevm-circuits/src/evm_circuit/util/constraint_builder.rs
hero78119: Done and add comments :)

lispc:
i think the fix is correct. A small comment:  it should be better to change this value from 10 to 9 ( https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/e26176d2990c74677f0b70e18b54ee535c8df475/zkevm-circuits/src/evm_circuit/execution/end_tx.rs#L122 ) and use `rw_counter: Delta(rw_counter_offset.clone()),` instead of `rw_counter: Delta(rw_counter_offset.clone() - 1.expr()),`

ChihChengLiang:
Hi @lispc,
Great catch! Created an issue #1725 and assigned @curryrasul as good first issue.

PR #1721, opened 2023-12-28
chores: add create/create2 in stats

hero78119:
### Description

- add create/create2 in stats
- in stats, separate dynamic gas with constant gas
- sort by height per constant gas to derive worst case

### DIFF
--- a/zkevm-circuits/src/bin/stats/helpers.rs
+++ b/zkevm-circuits/src/bin/stats/helpers.rs
@@ -86,9 +86,13 @@ struct Row {
     opcode: OpcodeId,
     #[table(title = "Height", justify = "Justify::Right")]
     height: usize,
-    #[table(title = "Gas Cost", justify = "Justify::Right")]
+    #[table(title = "Const Gas Cost", justify = "Justify::Right")]
     gas_cost: u64,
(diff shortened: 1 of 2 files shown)


ChihChengLiang (approved):
LGTM.
For other reviewers, you can run `cargo run --bin stats --features=stats -- evm` to see the result.

KimiWu123 (approved):
LGTM! 
Just curious what the benefit is to separate gas_cost into constant one and dynamic one, and the reason why adding a `height_per_dynamic_gas` row?

hero78119:
> LGTM! Just curious what the benefit is to separate gas_cost into constant one and dynamic one, and the reason why adding a `height_per_dynamic_gas` row?

This pr basically
- swap original `gas_cost := constant gas + dynamic gas` with constant gas instead.
- create new field `dynamic_gas_cost` to hold the `gas_cost := constant gas + dynamic gas` value.

Because the gas_cost from geth struct log are including base gas cost (opcode const gas) + dynamic cost (memory expansion etc..) so the row per gas (gas_cost/row) might looks too optimistic. So we separate constant gas cost and sort by it, therefore it equivalent we are care more about the worst case.

Issue #1636, opened 2023-09-28
Add runtime feature set


ChihChengLiang:
### Describe the feature you would like

The Scroll's and Taiko's forks have diverged significantly from this codebase. The cause for the divergence is our codebase has some L1-specific features, and the collaborating teams need to customize for L2 use cases.

We plan to increase the modularity of the codebase so that we can minimize the difference between different forks.

A while ago @Brechtpd proposed the idea of having runtime configs. Teams can include the features they want by configuring the feature set.

@ed255 [proposed](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1630#discussion_r1340265757) that we introduce a feature set that looks like

```rust
enum FeatureSet {
  EthereumL1,
  Scroll,
  Taiko,
}
```

However, the team-specific flag might be confusing because it doesn't convey what are the functionalities to include and why they are included. For example, Scroll skips TestBalanceOverflow and TestDifficulty in Ethereum tests. But it is hard to expect those tests to be skipped by seeing a feature named "scroll". When we see tests are skipped with `if match!(feature_set, FeatureSet::Scroll)`, it is hard to guess why they are skipped. 

I propose we name the FeatureSet with its functionalities.

```rust
enum FeatureSet {
  EthereumL1,
  ZeroDifficulty,
  L2Balance,
}
```
And we define variables to include those features.
```
const ScrollFeatures = [ZeroDifficulty, L2Balance]
```
 `if match!(feature_set, FeatureSet::ZeroDifficulty)` would tell the exact reason why a test is skipped.

### Additional context

Before we work on an actual PR, we need some input on these.

- What feature flags should we include?
- How do we structure the flags? What module should we put the feature set in and where to configure them?

Brechtpd:
Feature set with specific features sounds good to me!

 On the Taiko side I can think of these features:
- No fee payment for 1st transaction. The 1st transaction in Taiko is a generated transaction to help sync data between L1 and L2. It also does some additional checks so we don't have to do them directly in a circuit. Maybe in the future this will a be a dedicated transaction type though.
- Disable standard EIP-1559 calculation to calculate the basefee
- Invalid tx support

I think that's it, will add later if there are more.

PR #1728, opened 2024-01-10
Runtime config

ChihChengLiang:
### Description

We add an initial structure for runtime config. In this PR, we plan to add only the invalid tx configuration for the starter.

### Issue Link

#1636 

### Type of change

### Decision

- Allow Geth to take invalid tx. The config doesn't affect geth util

### TODO

- [x] Fix invalid tx test
- [ ] Add tx validity check. (Will continue #1740)

### DIFF
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -674,14 +674,7 @@ pub fn build_state_code_db(
 impl<P: JsonRpcClient> BuilderClient<P> {
     /// Create a new BuilderClient
     pub async fn new(client: GethClient<P>, circuits_params: FixedCParams) -> Result<Self, Error> {
-        let chain_id = client.get_chain_id().await?;
-
-        Ok(Self {
-            cli: client,
-            chain_id: chain_id.into(),
-            circuits_params,
-            feature_config: FeatureConfig::default(),
-        })
+        Self::new_with_features(client, circuits_params, FeatureConfig::default()).await
     }
 
     /// Create a new BuilderClient
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -47,10 +47,18 @@ pub use withdrawal::{Withdrawal, WithdrawalContext};
 /// Default to mainnet block
 #[derive(Debug, Clone, Copy)]
 pub struct FeatureConfig {
-    zero_difficulty: bool,
-    free_first_tx: bool,
-    enable_eip1559: bool,
-    invalid_tx: bool,
+    /// Zero difficulty
+    pub zero_difficulty: bool,
+    /// Free first transaction
+    pub free_first_tx: bool,
+    /// Enable EIP1559
+    pub enable_eip1559: bool,
+    /// Allow invalid transactions to be included in a block
+    ///
+    /// Transactions with mismatched nonce, insufficient gas limit, or insufficient balance
+    /// shouldn't be included in a mainnet block. However, rollup developers might want to
+    /// include invalid tx in the L2 block to support forced exit feature.
+    pub invalid_tx: bool,
 }
 
 impl Default for FeatureConfig {
@@ -172,6 +180,7 @@ pub struct CircuitInputBuilder<C: CircuitsParams> {
     pub circuits_params: C,
     /// Block Context
     pub block_ctx: BlockContext,
+    /// Feature config
     pub feature_config: FeatureConfig,
 }
 
@@ -310,7 +319,7 @@ impl<'a, C: CircuitsParams> CircuitInputBuilder<C> {
                 ExecState::InvalidTx,
             )?;
             tx.steps_mut().push(invalid_tx_step);
-        }else {
+        } else {
             panic!("invalid tx support not enabled")
         }
 
@@ -489,6 +498,7 @@ impl CircuitInputBuilder<DynamicCParams> {
             block: self.block,
             circuits_params: c_params,
             block_ctx: self.block_ctx,
+            feature_config: self.feature_config,
         };
 
         cib.set_end_block(c_params.max_rws)?;
@@ -773,7 +783,13 @@ impl<P: JsonRpcClient> BuilderClient<P> {
         prev_state_root: Word,
     ) -> Result<CircuitInputBuilder<FixedCParams>, Error> {
         let block = Block::new(self.chain_id, history_hashes, prev_state_root, eth_block)?;
-        let mut builder = CircuitInputBuilder::new(sdb, code_db, block, self.circuits_params);
+        let mut builder = CircuitInputBuilder::new(
+            sdb,
+            code_db,
+            block,
+            self.circuits_params,
+            FeatureConfig::default(),
+        );
         builder.handle_block(eth_block, geth_traces)?;
         Ok(builder)
     }
--- a/bus-mapping/src/circuit_input_builder.rs
+++ b/bus-mapping/src/circuit_input_builder.rs
@@ -72,6 +72,13 @@ impl Default for FeatureConfig {
     }
 }
 
+impl FeatureConfig {
+    /// Check if we are mainnet config
+    pub fn is_mainnet(&self) -> bool {
+        self.zero_difficulty && !self.free_first_tx && self.enable_eip1559 && !self.invalid_tx
+    }
+}
+
 /// Circuit Setup Parameters
 #[derive(Debug, Clone, Copy)]
 pub struct FixedCParams {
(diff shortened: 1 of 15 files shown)


zkevm-circuits/src/evm_circuit.rs
```diff
@@ -390,12 +403,17 @@ impl<F: Field> Circuit<F> for EvmCircuit<F> {
...
                     u16_table,
+                    feature_config: params,
                 },
             ),
             challenges,
         )
     }
 
+    fn configure(meta: &mut ConstraintSystem<F>) -> Self::Config {
+        Self::configure_with_params(meta, Default::default())
```
ed255, line 414: halo2 will not use this method when `configure_with_params` is implemented.  I would suggest adding an `unreachable!()` here to make sure this method is not called by accident (and thus guarantee that we're always using the features passed by the user).

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -457,7 +478,7 @@ mod evm_circuit_stats {
     fn evm_circuit_unusable_rows() {
         assert_eq!(
             EvmCircuit::<Fr>::unusable_rows(),
-            unusable_rows::<Fr, EvmCircuit::<Fr>>(()),
+            unusable_rows::<Fr, EvmCircuit::<Fr>>(FeatureConfig::default()),
```
ed255, line 481: I wonder if different feature configs could lead to different unusable rows :thinking:

ed255, line 481: I wonder if different feature configs could lead to different unusable rows :thinking:

zkevm-circuits/src/evm_circuit/param.rs
```diff
@@ -173,7 +174,14 @@ lazy_static::lazy_static! {
 }
 fn get_step_height_map() -> HashMap<ExecutionState, usize> {
     let mut meta = ConstraintSystem::<Fr>::default();
-    let circuit = EvmCircuit::configure(&mut meta);
+    let circuit = EvmCircuit::configure_with_params(
+        &mut meta,
+        FeatureConfig {
+            // Enable invalid_tx to get ExecutionState height
+            invalid_tx: true,
```
ed255, line 181: What happens when we use this `step_height_map` with the feature `invalid_tx: false`?  Does any state height change?  Or is it only that we have an extra one for the InvalidTx state?
I think we could have the situation where a a state height changes, maybe not now?

I think it's worth adding a check to make sure that all common states between `step_height_map` built out of different `FeatureConfig`s have the same height, to detect bugs as soon as possible.

As an example. with the `invalid_tx` feature, the EndTx constraints change a bit.  Imagine this change causes an extra cell to be allocated, and then we have that `height[EndTx]=5` with `invalid_tx=false`, and `height[EndTx]=6` with `invalid_tx=true`!  Provably this is not the case right now, but I imagine this situation could happen with other features.

zkevm-circuits/src/super_circuit.rs
```diff
@@ -407,19 +401,12 @@ impl<F: Field> Circuit<F> for SuperCircuit<F> {
...
-                max_txs: params.max_txs,
-                max_withdrawals: params.max_withdrawals,
-                max_calldata: params.max_calldata,
-                mock_randomness: params.mock_randomness,
-            },
-        )
+        Self::Config::new(meta, params)
     }
 
     fn configure(_meta: &mut ConstraintSystem<F>) -> Self::Config {
```
ed255, line 412: Note: Previously I suggested doing `unreachable!()` in the `configure` of the `EvmCircuit`, it's following this pattern.

zkevm-circuits/src/test_util.rs
```diff
@@ -175,12 +188,19 @@ impl<const NACC: usize, const NTX: usize> CircuitTestBuilder<NACC, NTX> {
...
 
-        let circuit = EvmCircuitCached::get_test_circuit_from_block(block);
-        let prover = MockProver::<Fr>::run(k, &circuit, vec![]).map_err(|err| {
-            CircuitTestError::SynthesisFailure {
-                circuit: Circuit::EVM,
-                reason: err,
-            }
+        // Mainnet EVM circuit constraints can be cached for test performance.
+        // No cache for EVM circuit with customized features
+        let prover = if block.feature_config.is_mainnet() {
```
ed255, line 193: Very nice! This resolves my question about using the cached version in non-mainnet features :)

ed255 (approved):
Overall looks good to me!  I think the way the runtime config is implemented is very nice; in particular it's not very intrusive!  And the features are now available anywhere where they are needed; to allow the implementation to take different paths as necessary without making thinks too complex :)

I'm approving already even though I left some comments and suggestions; as I think they are minor and don't affect the current implementation.  The most serious one is the state height map discussion.

zkevm-circuits/src/evm_circuit.rs
```diff
@@ -310,6 +314,7 @@ pub(crate) mod cached {
         /// Circuit configuration.  These values are calculated just once.
         static ref CACHE: Cache = {
             let mut meta = ConstraintSystem::<Fr>::default();
+            // Cached EVM circuit is configured with Mainnet FeatureConfig
```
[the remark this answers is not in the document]
ed255: Self resolved with https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1728#discussion_r1464777767

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -741,7 +790,13 @@ impl<P: JsonRpcClient> BuilderClient<P> {
         prev_state_root: Word,
     ) -> Result<CircuitInputBuilder<FixedCParams>, Error> {
         let block = Block::new(self.chain_id, history_hashes, prev_state_root, eth_block)?;
-        let mut builder = CircuitInputBuilder::new(sdb, code_db, block, self.circuits_params);
+        let mut builder = CircuitInputBuilder::new(
+            sdb,
+            code_db,
+            block,
+            self.circuits_params,
+            FeatureConfig::default(),
```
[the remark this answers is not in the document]
ChihChengLiang: I think this is a good idea. Resolved in [f9e197d](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1728/commits/f9e197dbdcba9400564f81f78e0c3f2c52b52a87)

zkevm-circuits/src/evm_circuit.rs
ChihChengLiang: Fixed here. [8787b3c](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1728/commits/8787b3c321647294fc03bf58d61c1b8bcb613dbd)

bus-mapping/src/circuit_input_builder.rs
```diff
@@ -630,6 +680,23 @@ impl<P: JsonRpcClient> BuilderClient<P> {
             cli: client,
             chain_id: chain_id.into(),
             circuits_params,
+            feature_config: FeatureConfig::default(),
```
Brechtpd, line 683: Maybe this should just call through to `new_with_features` with `FeatureConfig::default()`?

zkevm-circuits/src/evm_circuit/execution.rs
```diff
@@ -1259,7 +1276,11 @@ impl<F: Field> ExecutionConfig<F> {
             ExecutionState::BeginTx => assign_exec_step!(self.begin_tx_gadget),
             ExecutionState::EndTx => assign_exec_step!(self.end_tx_gadget),
             ExecutionState::EndBlock => assign_exec_step!(self.end_block_gadget),
-            ExecutionState::InvalidTx => assign_exec_step!(self.invalid_tx),
+            ExecutionState::InvalidTx => {
+                if let Some(invalid_tx) = &self.invalid_tx {
+                    assign_exec_step!(invalid_tx)
+                }
```
Brechtpd, line 1282: Maybe can just unwrap it here and let it error out if it's not set up so it doesn't silently skip it?

Brechtpd (approved):
Very nice! Only some really unimportant comments from me, the step heights issue pointed out by Edu I think indeed potentially problematic.

bus-mapping/src/circuit_input_builder.rs
ChihChengLiang: Good idea. Fixed in [949c228](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1728/commits/949c2284162023737ee90d681fe33d71bd220920)

zkevm-circuits/src/evm_circuit/execution.rs
ChihChengLiang: Nice catch! I fixed it here. [e4be4d0](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1728/commits/e4be4d03736fcd9f80848b4db131181b8b5d1b00)

ed255 (approved):
I've reviewed the latest changes on this PR and all looks good to me!

PR #1730, opened 2024-01-13
chore(mpt-witness-generator): bit of proofreading

Ursulafe:
Find few typos while studying the code, in comments. Hope I can be of any help, even if tiny.
Have a nice weekend.


ChihChengLiang (approved):
LGTM. Thanks for the quick fix.

PR #1731, opened 2024-01-13
chore: set resolver v2 at the workspace level

bishopcheckmate:
### Description

Cargo defaults to `resolver = "2"` in edition 2021 only for singular crates. Workspace tho defaults to the old `resolver = "1"` and will override this setting for all crates in the workspace. [The guideline](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-workspace) is to manually set the `resolver = "2"` on workspace level.
This was documented recently and since a few releases produces a warning on building. You don't see it due to the `rust-toolchain` setting.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue) (not really a bugfix but that's the closest match)

### Contents

- specifies `resolver = "2"` on workspace level
- removes `resolver = "2"` from light-client-poc/Cargo.toml because it was ignored anyway

This also fixes this warning:

```
warning: resolver for the non root package will be ignored, specify resolver at the workspace root:
package:   /data/bishop/zkevm-circuits/light-client-poc/Cargo.toml
workspace: /data/bishop/zkevm-circuits/Cargo.toml
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
     "zkevm-circuits",
     "bus-mapping",
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -40,7 +40,7 @@ mock_prover = []
 
 [build-dependencies]
 env = "0.0.0"
-ethers = "=2.0.10"
+ethers = { version = "=2.0.10", features = ["ethers-solc"] }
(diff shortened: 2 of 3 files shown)


bishopcheckmate:
I pushed the commit which should fix the builds. With resolver = 2, selected features are not propagated through different dependency types and targets so `ethers-solc` had to be explicitly enabled in build deps

ChihChengLiang (approved):
LGTM. Thanks for the fix.

PR #1735, opened 2024-01-15
Add comment regarding RwTableTag

zemse:
Adds a comment so that it makes it clear what is RwTableTag when going through the code.


han0110 (approved):
LGTM

PR #1745, opened 2024-01-29
MPT max nodes

miha-stopar:
### Description

This PR resolves the test failure in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1706.

The parameter `max_nodes` has been added to the MPT to have always the same fixed column `q_enable`.

The fixed column `q_last` has been removed because it is not necessary - if some nodes are added to the MPT proof after the account/storage leaf to enable some non-valid lookups, the proof would fail because of address/key RLC constraints (the RLC is being built node by node).

### Issue Link

https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1700

### Type of change

### DIFF
--- a/zkevm-circuits/src/mpt_circuit.rs
+++ b/zkevm-circuits/src/mpt_circuit.rs
@@ -769,8 +769,7 @@ mod tests {
 
     #[test]
     fn test_mpt() {
-        // let path = "src/mpt_circuit/tests";
-        let path = "/Users/miha/projects/2appliedzkp/variadic-main/zkevm-circuits/zkevm-circuits/src/mpt_circuit/tests";
+        let path = "src/mpt_circuit/tests";
         let files = fs::read_dir(path).unwrap();
         files
             .filter_map(Result::ok)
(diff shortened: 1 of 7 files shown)


zkevm-circuits/src/mpt_circuit.rs
```diff
@@ -497,24 +493,32 @@ impl<F: Field> MPTConfig<F> {
...
+                        self.params.max_nodes,
+                        2 * HASH_WIDTH + 1,
+                    );
+                }
+                if offset > self.params.max_nodes {
+                    panic!(
+                        "The parameter max_nodes is set too low, max_nodes: {}, offset: {}",
+                        self.params.max_nodes,
+                        offset,
+                    );
```
ChihChengLiang, line 509: Nitpick: use assert here for brevity.
```
                assert!( self.params.max_nodes < (2 * HASH_WIDTH + 1),
                        "The parameter max_nodes is set too low for the mult table: {}, mult table height: {}",
                        self.params.max_nodes,
                        2 * HASH_WIDTH + 1,
                    );
                assert!( offset > self.params.max_nodes,
                        "The parameter max_nodes is set too low, max_nodes: {}, offset: {}",
                        self.params.max_nodes,
                        offset,
                    );
```

ChihChengLiang (approved):
LGTM with a nitpick.

Issue #1736, opened 2024-01-17
Two `Word` types in zkEVM codebase


zemse:
### Describe the feature you would like

Currently, there are two `Word` types in the codebase

- [zkevm_circuits::util::word::Word](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4dc76ebd2c1776abb15ee5bc8064bd13b000f78c/zkevm-circuits/src/util/word.rs#L199) has two limbs.
- [eth_types::lib::Word](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/4dc76ebd2c1776abb15ee5bc8064bd13b000f78c/eth-types/src/lib.rs#L152) which is alias of U256.

It would be great if one of these is renamed.

rrtoledo:
There were some proposals for Circuit's words at an online meeting including:
- ZWord
- CWord
- WordLimbs
- WordCell

rrtoledo:
There are also some types with the same names in bus mapping and zkevm-circuits such as `Block` (defined in bus-mapping/src/circuit_input_builder/block.rs and zkevm-circuits/src/witness/block.rs).
Do you think we should open an issue per homonyms or turn this one here for all?

ChihChengLiang:
@rrtoledo Let's address the renaming of the Blocks in a new issue.

Feel free to choose any name you see fit for Word.

PR #1747, opened 2024-01-31
Replace word::Word with WordLoHi

zemse:
### Description

Replacing Word with WordHiLo.

### Issue Link

#1736

### Type of change

- [x] Refactor

### Contents

- Replace word::Word with WordLoHi
- Replace word::WordCell with WordLoHiCell

### Rationale

- WordLimbs and WordCell is used somewhere so couldn't go with that. 
- ZWord and CWord could be used, however, I saw that Word is actually an alias for Word2 ([source](https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/7f356548dd6dd614289b9700a40fbc80c4949b47/zkevm-circuits/src/util/word.rs#L203)). Which is an alias of WordLimbs<2>. So Word2 could be better than ZWord or CWord, but I think WordHiLo is even better and makes the code more obvious, since we have hi-lo terminology in many places.

### How Has This Been Tested?

Built all packages locally.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,6 +12,7 @@
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] This change requires a documentation update
+- [ ] Refactor (no updates to logic)
 
 ### Contents
 
--- a/zkevm-circuits/src/circuit_tools/cell_manager.rs
+++ b/zkevm-circuits/src/circuit_tools/cell_manager.rs
@@ -4,7 +4,7 @@ use crate::{
     circuit_tools::cached_region::CachedRegion,
     evm_circuit::util::rlc,
     table::LookupTable,
-    util::{query_expression, word::Word, Expr},
+    util::{query_expression, word::WordLoHi, Expr},
 };
 use eth_types::Field;
 use halo2_proofs::{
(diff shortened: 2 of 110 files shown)


.github/pull_request_template.md
```diff
@@ -12,6 +12,7 @@
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] This change requires a documentation update
+- [ ] Refactor (no updates to logic)
```
ChihChengLiang, line 15: I like this category 👍🏽

zkevm-circuits/src/circuit_tools/cell_manager.rs
```diff
@@ -103,11 +103,11 @@ impl<F: Field> Expr<F> for &Cell<F> {
...
 
-pub(crate) type WordCell<F> = Word<Cell<F>>;
+pub(crate) type WordHiLoCell<F> = WordHiLo<Cell<F>>;
 
-impl<F: Field> WordCell<F> {
-    pub fn expr(&self) -> Word<Expression<F>> {
-        Word::new([self.lo().expr(), self.hi().expr()])
+impl<F: Field> WordHiLoCell<F> {
+    pub fn expr(&self) -> WordHiLo<Expression<F>> {
+        WordHiLo::new([self.lo().expr(), self.hi().expr()])
```
ChihChengLiang, line 110: I notice a problem naming it `WordHiLo`. 
We put the first argument "Low" and the second argument "High," which is the reverse order of "HiLo."
I propose we name it `WordLoHi` to reflect the correct order of the argument.
Or even we can drop the `Word` and call it `LoHi` for brevity. And it seems "Lohi" is [salmon in Finnish](https://fi.wikipedia.org/wiki/Lohi).

ChihChengLiang (commented):
Nice renaming! After this PR, we no longer need to write `word::Word` everywhere.
I left a comment.

zkevm-circuits/src/circuit_tools/cell_manager.rs
zemse: Yes thanks for pointing out, LoHi is more logical.

> And it seems "Lohi" is [salmon in Finnish](https://fi.wikipedia.org/wiki/Lohi).

😄

ChihChengLiang (approved):
LGTM. Thanks for the fix ❤️

KimiWu123 (approved):
LGTM! Feel free to merge it after light testing is fixed.

Issue #1645, opened 2023-10-05
Report halo2 stats of each sub circuit


ed255:
Write a binary (or utility function triggered by a test) that reports: 
- number of gates
- max degree of gates
- number of advice, instance and public columns
- number of rotations?
- number of lookups
- number of copy constraints (or number of column pairs involved in copy constraints)

lispc:
I have this https://github.com/scroll-tech/zkevm-circuits/blob/7d9bc181953cfc6e7baf82ff0ce651281fd70a8a/zkevm-circuits/src/util.rs#L294 haha

ed255:
> I have this https://github.com/scroll-tech/zkevm-circuits/blob/7d9bc181953cfc6e7baf82ff0ce651281fd70a8a/zkevm-circuits/src/util.rs#L294 haha
> 
> https://github.com/scroll-tech/zkevm-circuits/blob/7d9bc181953cfc6e7baf82ff0ce651281fd70a8a/zkevm-circuits/src/super_circuit.rs#L257 And print these nums after each sub circuit

That's awesome!  We can just port that, as it fulfills our exact needs :D

PR #1763, opened 2024-02-09
SubCircuit halo2 stats

ed255:
### Description

Add a new command to the stats binary that reports halo2 circuit stats for each circuit, and the supercircuit.

### Issue Link

Resolve https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1645

### Type of change

### Contents

- Added new module in `stats` that records halo2 circuit stats from the configured `ConstraintSystem`, based on Scroll's code from https://github.com/scroll-tech/zkevm-circuits/blob/7d9bc181953cfc6e7baf82ff0ce651281fd70a8a/zkevm-circuits/src/util.rs#L294
  - The collection of stats works in two phases, first it records per-circuit stats and then total stats (to get super circuit stats).  The per-circuit stats works by configuring all the shared tables into the same `ConstraintSystem` (and obtaining stats via deltas), and then configuring each sub-circuit starting from the shared tables `ConstraintSystems` each time.
  - In order to separate the shared column configuration from the sub-circuit and constraints configuration, I had to extract the column definition of the BinaryNumberChip into BinaryNumberBits so that the CopyTable can be constructed without any constraint.
- Added peak memory estimation based on @han0110's analysis

### Results

These results are for `k = 26`.  At the current worst-case estimation of At 0.0139 gas/row this gives us 2^26 * 0.0139 = ~900k gas.  For 30M gas we would need 33 chunks like that.

|circuit|constraints|rots|min/max(rots)|fix_cols|sels|advs|perms|lookups|degree|mem_gb|
|:----|:----|:----|:----|:----|:----|:----|:----|:----|:----|:----|
|tx_table|0|0|0/0|1|0|4|0|0|3|58|
|wd_table|0|0|0/0|0|0|5|0|0|3|56|
|rw_table|0|0|0/0|0|0|14|0|0|3|110|
|mpt_table|0|0|0/0|0|0|12|0|0|3|98|
|bytecode_table|0|0|0/0|0|0|6|0|0|3|62|
|block_table|0|0|0/0|2|0|2|0|0|3|54|
|copy_table|0|0|0/0|1|0|12|0|0|3|106|
|exp_table|0|0|0/0|1|0|5|0|0|3|64|
|keccak_table|0|0|0/0|0|0|5|0|0|3|56|
|sig_table|0|0|0/0|1|0|9|0|0|3|88|
|u8_table|0|0|0/0|1|0|0|0|0|3|34|
|u10_table|0|0|0/0|1|0|0|0|0|3|34|
|u16_table|0|0|0/0|1|0|0|0|0|3|34|
|keccak|2523|105|-89/207|18|0|198|0|123|4|3000|
|pi|21|2|0/1|3|7|10|15|3|9|754|
|tx|2|2|0/1|13|4|6|9|6|6|780|
|bytecode|23|2|0/1|5|0|6|0|2|8|342|
|copy|34|3|0/2|1|1|14|0|12|9|466|
|state|203|3|-1/1|3|0|49|0|36|10|2224|
|exp|44|11|0/10|0|1|10|0|0|5|142|
|evm|42318|21|0/20|5|3|131|3|53|7|2976|
|super|45168|106|-89/207|57|16|498|25|235|10|21736|

PD: Yes, the estimation is 21TB of memory for a super circuit proof :(  Nevertheless there are various things we can do to improve this.


ed255:
Marking as draft because it seems I broke the CopyCircuit or CopyTable.

zkevm-circuits/src/super_circuit.rs
```diff
@@ -128,20 +130,29 @@ impl<F: Field> SubCircuitConfig<F> for SuperCircuitConfig<F> {
...
+        #[cfg(not(feature = "mock-challenge"))]
+        let challenges = Challenges::construct(meta);
+
+        #[cfg(feature = "mock-challenge")]
+        let challenges_exprs = Challenges::mock(
             power_of_randomness[0].clone(),
             power_of_randomness[0].clone(),
         );
+        #[cfg(not(feature = "mock-challenge"))]
+        let challenges_exprs = { challenges.exprs(meta) };
```
KimiWu123, line 149: nitpick: I think we don't need `{ }`

KimiWu123 (commented):
First round pass. Will take a deeper look at `halo2_stats.rs`.

zkevm-circuits/src/bin/stats/halo2_stats.rs
```diff
@@ -0,0 +1,373 @@
...
+        let c_pg = (c_p + self.degree - 3) / (self.degree - 2);
+        let e = (self.degree - 1).next_power_of_two();
+        // number of "n elliptic curve points"
+        let m_c = 2;
+        // number of "n field elements"
+        let m_s = 1 + 2 * c_f + 2 * c_p + 2 * c_i + c_a + 3 * c_l + c_pg;
+        // number of "e * n field elements"
+        let m_e = 4 + c_f + c_p + c_pg + c_i + c_a + 3 * (c_l > 0) as usize;
+        let unit = m_c + m_s + e * m_e;
+        unit * 2usize.pow(k) * field_bytes
```
KimiWu123, line 249: `unit * 2usize.pow(k) * field_bytes` looks like to calculate the memory usage as above formula `M = 32 * n * (2 * M.C + M.S + e * M.E)` for me. But it seems not exact the same.

The implementation here is
unit * 2usize.pow(k) * field_bytes --> (m_c + m_s + e * m_e) * n * 32
however M is
(`2 * M.C` + M.S + e * M.E) * n * 32

Wondering which one is correct?

KimiWu123 (commented):
Overall looks good! Have a question about formula and implementation mismatch.

zkevm-circuits/src/bin/stats/halo2_stats.rs
```diff
@@ -0,0 +1,373 @@
...
+    //
+    // The memory usage M.C and M.S stands for:
+    //
+    // - M.C: number of "n elliptic curve points" (with symbol ◯)
+    // - M.S: number of "n field elements"        (with symbol △)
+    // - M.E: number of "e * n field elements"    (with symbol ⬡)
+    //
+    // So the actual memory usage in terms of bytes will be:
+    //
+    //   M = 32 * n * (2 * M.C + M.S + e * M.E)
```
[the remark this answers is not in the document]
ed255: I'll update the documentation to explain `M.E` as:
```
 - M.E: number of "n field elements" that will be extended by "* e"
```
So now `M.E` should be treated as independent of `e` but in the formula to calculate `M` we have `e * M.E`.

> So, I can see this formula as M = 32 * n * (2 * M.C + M.S + `e^2 * M.S`), right?

Hm, I don't follow this.  The way you wrote this it implies that `e^2 * M.S = e * M.E <=> e * M.S = M.E`, but  I don't think this is the case.  Maybe the confusion is that we have `n = 1<< k` number of rows, but also `M.C, M.S, M.E` are defined with `n`?  I'll update it like this then:

```
    // - M.C: number of "num elliptic curve points"                         (with symbol ◯)
    // - M.S: number of "num field elements"                                (with symbol △)
    // - M.E: number of "num field elements" that will be extended by "* e" (with symbol ⬡)
```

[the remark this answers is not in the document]
ed255: I also noticed that I was missing a `2 *` next to `m_c` in the formula here:  
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/2905d4e9654181e16c24e74f9bcdd609bc7825e5/zkevm-circuits/src/bin/stats/halo2_stats.rs#L248
I'll fix that as well.

zkevm-circuits/src/bin/stats/halo2_stats.rs
ed255: Yes you're correct, the `2 *` is missing!  Thanks for pointing this out!  I also saw this when checking your review message https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1763#discussion_r1493946938 :smiley:

zkevm-circuits/src/super_circuit.rs
ed255: Thanks!  Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1763/commits/4fe16a01d0f6cad3199633038a45d27cc118253c

zkevm-circuits/src/bin/stats/halo2_stats.rs
```diff
@@ -0,0 +1,373 @@
...
+#[derive(Debug, Default)]
+pub(crate) struct CircuitStats {
+    pub num_constraints: usize,
+    pub num_fixed_columns: usize,
+    pub num_lookups: usize,
+    pub num_shuffles: usize,
+    pub num_advice_columns: usize,
+    pub num_instance_columns: usize,
+    pub num_selectors: usize,
+    // num_simple_selectors: usize,
```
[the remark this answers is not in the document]
ed255: Thanks!  This was a leftover from copying the implementation from Scroll.  Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1763/commits/4fe16a01d0f6cad3199633038a45d27cc118253c

zkevm-circuits/src/bin/stats/halo2_stats.rs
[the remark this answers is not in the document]
ed255: Fixed in https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1763/commits/4fe16a01d0f6cad3199633038a45d27cc118253c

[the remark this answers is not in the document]
KimiWu123: It looks clear for me now! Thanks.

KimiWu123 (approved):

zkevm-circuits/src/bin/stats/halo2_stats.rs
```diff
@@ -0,0 +1,374 @@
...
+        let e = (self.degree - 1).next_power_of_two();
+        // The estimated peak memory formula comes from point 9 of the analysis, which is the step
+        // of proving that needs the most memory (after that step, allocations start getting freed)
+
+        // number of "elliptic curve points"
+        let m_c = 2;
+        // number of "field elements"
+        let m_s = 1 + 2 * c_f + 2 * c_p + 2 * c_i + c_a + 3 * c_l + c_pg;
+        // number of "field elements" that will be extended by "* e"
+        let m_e = 4 + c_f + c_p + c_pg + c_i + c_a + 3 * (c_l > 0) as usize;
```
miha-stopar, line 249: These formulas for `m_s` and `m_e` are from step 10. Should they be from step 11?

zkevm-circuits/src/bin/stats/halo2_stats.rs
```diff
@@ -0,0 +1,374 @@
...
+        min_rotation: rotations.first().cloned().unwrap_or_default(),
+        max_rotation: rotations.last().cloned().unwrap_or_default(),
+        num_verification_ecmul: num_advice_columns
+            + num_instance_columns
+            + num_permutation_columns
+            + num_shuffles
+            + num_selectors
+            + num_fixed_columns
+            + 3 * num_lookups
+            + rotations.len(),
```
han0110, line 314: I think this part is not accurate:
- `num_instance_columns` will not contribute when we are using KZG commitment scheme
- grand products in permutaton argument are not counted in (`div_ceil(num_permutation_columns, degree - 2)`)
- quotients are not counted in (`degree - 1`)
- `num_selectors` might be ambiguous cause it might be compressed
- batch open scheme matters (GWC as `rotations.len()`, SHPLONK as `2`).

zkevm-circuits/src/bin/stats/halo2_stats.rs
ed255: The formulas for `m_s` and `m_e` should correspond to step 9 (which is the step that has the peack memory usage).  I think the current code is correct:
https://github.com/privacy-scaling-explorations/zkevm-circuits/blob/7020216f02985a29232df398032b94d4b8d6109b/zkevm-circuits/src/bin/stats/halo2_stats.rs#L190-L202

gadgets/src/binary_number.rs
```diff
@@ -104,10 +148,11 @@ where
     /// Annotates columns of this gadget embedded within a circuit region.
     pub fn annotate_columns_in_region<F: Field>(&self, region: &mut Region<F>, prefix: &str) {
         let mut annotations = Vec::new();
-        for (i, _) in self.bits.iter().enumerate() {
+        for (i, _) in self.bits.0.iter().enumerate() {
             annotations.push(format!("GADGETS_binary_number_{}", i));
         }
         self.bits
+            .0
```
ChihChengLiang, line 155: We can implement a Deref for BinaryNumberBits to prevent all the `.0` diff.

ChihChengLiang (commented):
I don't intend to do a full review. Please exclude me from the final approval.
I just want to give a quick comment on the Deref.

zkevm-circuits/src/bin/stats/halo2_stats.rs
han0110: > `num_selectors` might be ambiguous cause it might be compressed

Yes, for this bullet I'm talking about the link you share.

han0110 (approved):
LGTM!

zkevm-circuits/src/bin/stats/halo2_stats.rs
rrtoledo: So we might have extra fixed columns, but I guess some imprecision in the nbs or just having order of magnitudes may still be fine

han0110: yeah make sense, Edu also added [the comment](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1763/commits/d9fc61b7bff34c6c7b11905f309c1d18488cd42d#diff-0e9ee29187a7ec01aa350218b105d413b8f63ff6ac04747eafe63003a4a6ec6bR299) so we know this is an estimated upper bound

gadgets/src/binary_number.rs
ed255: Thanks!  This was a great suggestion!  I've applied it and with it reduced the diff :D

PR #1764, opened 2024-02-13
Simplify pair_select

ChihChengLiang:
### Description

Pair select depends only on expressions. We don't have to implement a gadget and assign values to it.

### Issue Link

This is a quick refactor

### Type of change

Refactor (no updates to logic)


ChihChengLiang:
This PR makes no sense now. I made an embarrassing mistake. 
`is_a` still needs to be assigned.
It might be possible to derive `is_a` from `a` expression and `b` expression, but that loses the point of pair_select. pair_select minimizes constraints by witnessing  `is_a`.

PR #1774, opened 2024-02-21
Remove mod.rs files

curryrasul:
### Description

Refactor the module system to be the same within a project

### Type of change

- [X] Refactor (no updates to logic)

### Rationale

We use two different ways to address the module system - without `mod.rs` (only with directories/files) and with `mod.rs` in some places. Unless there's a specific reason for that - one OR the other must be used. 
I refactored the code and removed `mod.rs` files.

### How Has This Been Tested?

I did `cargo build --release`


zemse (approved):
LGTM!

ChihChengLiang (approved):
LGTM. Thanks for the fix.

PR #1775, opened 2024-02-21
[Doc] Cleanup  Bus-Mapping, ExecutionSteps, and ExecutionTrace

ChihChengLiang:
### Description

-  Bus-Mapping has two library docs, one is inside the lib.rs, another locates in README.md, untested. We remove the one in lib.rs and make the README one checked.
- ExecutionSteps and ExecutionTrace were 

We hope the update can reduce confusion.

### Type of change

- [x] This change requires a documentation update
- [ ] Refactor (no updates to logic)

### How Has This Been Tested?

``` 
cargo test --doc
```


ChihChengLiang:
CI can catch the bug https://github.com/privacy-scaling-explorations/zkevm-circuits/actions/runs/7992225793/job/21825261087?pr=1775#step:10:2042

eth-types/src/lib.rs
```diff
@@ -476,14 +476,14 @@ impl<'de> Deserialize<'de> for GethExecStep {
 }
 
 /// Helper type built to deal with the weird `result` field added between
-/// `GethExecutionTrace`s in `debug_traceBlockByHash` and
+/// [`GethExecTrace`]s in `debug_traceBlockByHash` and
```
KimiWu123, line 479: Wondering the reason why you add `[ ]` here? Forgot to add `(eth_types::GethExecTrace)`? L486 is the same as well.

bus-mapping/README.md
```diff
@@ -167,18 +187,13 @@ On that way, we would get something like this for the Memory ops:
...
+| `0xa0` | `0xcafeb0ba`  | `Write` | 34   | `83 MSTORE`                              |
 ```
 
 Where as you see, we group by `memory_address` and then order by
 `global_counter`.
 
-- Iterate over the `ExecutionTrace` itself over
-each `ExecutionStep`'s is formed by and check which Stack/Memory&Storage operations are linked to each step.
+- Iterate over the [`GethExecTrace`](eth_types::GethExecTrace) itself over
+each [`ExecStep`](crate::circuit_input_builder::ExecStep)'s is formed by and check which Stack/Memory&Storage operations are linked to each step.
```
KimiWu123, line 197: nit: not changing line here.

KimiWu123 (approved):
LGTM! Left a question and a nitpick.

eth-types/src/lib.rs
ChihChengLiang: We mark a term as a keyword with `[ ]`. The `cargo doc` checks whether the referenced term in the brackets exists. Let's say if we rename `GethExecTrace` to something else, it will raise a compilation error and remind us to update the name here.

 For `(eth_types::GethExecTrace)`, if the referenced struct is already in the same file, we can drop the import path. Cargo doc raises a warning if we add an unnecessary explicit link.

bus-mapping/README.md
ChihChengLiang: I fixed all the line breaks. I also added a GPT round to improve the readability.

eth-types/src/lib.rs
KimiWu123: Thanks for explanation! Learnt something new. If so, do we need to add `[ ]` for `debug_traceBlockByNumber`in L480?

ChihChengLiang: Note that `debug_traceBlockByNumber` is a name from Geth RPC API; it is not a variable defined in our codebase, so the Cargo doc knows nothing about it.

hero78119 (commented):
Overall LGTM!

I am checking whether we should add `-D warnings` along with `cargo test --doc` to assure even warning can be captured.

hero78119:
> Overall LGTM!
> 
> I am checking whether we should add `-D warnings` along with `cargo test --doc` to assure even warning can be captured.

Since this PR targeting on docs clean up, I suggest we can further restrict [warnings on docs test](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#showing-warnings-in-doctests)

Example:
https://github.com/privacy-scaling-explorations/zkevm-circuits/compare/doc-fix...hero78119:doc_deny_on_warning?expand=1

hero78119 (approved):
LGTM! appreciate for taking care of this

PR #1777, opened 2024-02-21
Merge Transfer and TransferWithGas gadget

ChihChengLiang:
### Description

- Transfer and TransferWithGas gadgets are mostly the same. We can merge them and duplicate them.
- There are duplicated logics for creating the receiver account (writing empty hash). We remove the duplicated one.
- The update balance gadget is vague, it is hard to understand whether we want to add or subtract balance. We extract a helpful method for that.
- deduplicate the transfer_to logic in busmapping.
- Fix the receiver creation condition on busmapping.
- Fix the order of the sender account read-write in begin tx assignment.

Thank @curryrasul for initializing this conversation.

### Type of change

Refactor (no updates to logic)

### Test

```
cargo test -p zkevm-circuits create
cargo test -p zkevm-circuits begin_tx
cargo test -p zkevm-circuits end_tx 
```

### DIFF
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -623,7 +623,14 @@ impl<'a> CircuitInputStateRef<'a> {
                 value_prev: sender_balance_prev,
             },
         )?;
-        self.transfer_to(step, receiver, receiver_exists, opcode_is_create, value, true)?;
+        self.transfer_to(
+            step,
+            receiver,
+            receiver_exists,
+            opcode_is_create,
+            value,
+            true,
+        )?;
 
         Ok(())
     }
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -565,7 +565,7 @@ impl<'a> CircuitInputStateRef<'a> {
         sender: Address,
         receiver: Address,
         receiver_exists: bool,
-        opcode_is_create: bool,
+        is_create: bool,
         value: Word,
         fee: Option<Word>,
     ) -> Result<(), Error> {
@@ -621,14 +621,7 @@ impl<'a> CircuitInputStateRef<'a> {
             )?;
         }
 
-        self.transfer_to(
-            step,
-            receiver,
-            receiver_exists,
-            opcode_is_create,
-            value,
-            true,
-        )?;
+        self.transfer_to(step, receiver, receiver_exists, is_create, value, true)?;
 
         Ok(())
     }
@@ -639,11 +632,11 @@ impl<'a> CircuitInputStateRef<'a> {
         step: &mut ExecStep,
         receiver: Address,
         receiver_exists: bool,
-        opcode_is_create: bool,
+        is_create: bool,
         value: Word,
         reversible: bool,
     ) -> Result<(), Error> {
-        if !receiver_exists && (!value.is_zero() || opcode_is_create) {
+        if !receiver_exists && (!value.is_zero() || is_create) {
             self.account_write(
                 step,
                 receiver,
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -559,13 +559,13 @@ impl<'a> CircuitInputStateRef<'a> {
     /// balance by `value`. If `fee` is existing (not None), also need to push 1
     /// non-reversible [`AccountOp`] to update `sender` balance by `fee`.
     #[allow(clippy::too_many_arguments)]
-    pub fn transfer_with_fee(
+    pub fn transfer(
         &mut self,
         step: &mut ExecStep,
         sender: Address,
         receiver: Address,
         receiver_exists: bool,
-        must_create: bool,
+        opcode_is_create: bool,
         value: Word,
         fee: Option<Word>,
     ) -> Result<(), Error> {
@@ -623,30 +623,9 @@ impl<'a> CircuitInputStateRef<'a> {
                 value_prev: sender_balance_prev,
             },
         )?;
-        self.transfer_to(step, receiver, receiver_exists, must_create, value, true)?;
+        self.transfer_to(step, receiver, receiver_exists, opcode_is_create, value, true)?;
 
         Ok(())
-    }
-
-    /// Same functionality with `transfer_with_fee` but with `fee` set zero.
-    pub fn transfer(
-        &mut self,
-        step: &mut ExecStep,
-        sender: Address,
-        receiver: Address,
-        receiver_exists: bool,
-        must_create: bool,
-        value: Word,
-    ) -> Result<(), Error> {
-        self.transfer_with_fee(
-            step,
-            sender,
-            receiver,
-            receiver_exists,
-            must_create,
-            value,
-            None,
-        )
     }
 
     /// Transfer to an address. Create an account if it is not existed before.
@@ -655,12 +634,11 @@ impl<'a> CircuitInputStateRef<'a> {
         step: &mut ExecStep,
         receiver: Address,
         receiver_exists: bool,
-        must_create: bool,
+        opcode_is_create: bool,
         value: Word,
         reversible: bool,
     ) -> Result<(), Error> {
-        // If receiver doesn't exist, create it
-        if !receiver_exists && (!value.is_zero() || must_create) {
+        if !receiver_exists && (!value.is_zero() || opcode_is_create) {
             self.account_write(
                 step,
                 receiver,
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -609,20 +609,18 @@ impl<'a> CircuitInputStateRef<'a> {
             sender_balance
         );
 
-        if value.is_zero() {
-            // Skip transfer if value == 0
-            return Ok(());
+        if !value.is_zero() {
+            self.push_op_reversible(
+                step,
+                AccountOp {
+                    address: sender,
+                    field: AccountField::Balance,
+                    value: sender_balance,
+                    value_prev: sender_balance_prev,
+                },
+            )?;
         }
 
-        self.push_op_reversible(
-            step,
-            AccountOp {
-                address: sender,
-                field: AccountField::Balance,
-                value: sender_balance,
-                value_prev: sender_balance_prev,
-            },
-        )?;
         self.transfer_to(
             step,
             receiver,
--- a/bus-mapping/src/circuit_input_builder/input_state_ref.rs
+++ b/bus-mapping/src/circuit_input_builder/input_state_ref.rs
@@ -608,18 +608,7 @@ impl<'a> CircuitInputStateRef<'a> {
             sender_balance_prev,
             sender_balance
         );
-        // If receiver doesn't exist, create it
-        if !receiver_exists && (!value.is_zero() || must_create) {
-            self.push_op_reversible(
-                step,
-                AccountOp {
-                    address: receiver,
-                    field: AccountField::CodeHash,
-                    value: CodeDB::empty_code_hash().to_word(),
-                    value_prev: Word::zero(),
-                },
-            )?;
-        }
+
         if value.is_zero() {
             // Skip transfer if value == 0
             return Ok(());
@@ -634,19 +623,7 @@ impl<'a> CircuitInputStateRef<'a> {
                 value_prev: sender_balance_prev,
             },
         )?;
-
-        let (_found, receiver_account) = self.sdb.get_account(&receiver);
-        let receiver_balance_prev = receiver_account.balance;
-        let receiver_balance = receiver_account.balance + value;
-        self.push_op_reversible(
-            step,
-            AccountOp {
-                address: receiver,
-                field: AccountField::Balance,
-                value: receiver_balance,
-                value_prev: receiver_balance_prev,
-            },
-        )?;
+        self.transfer_to(step, receiver, receiver_exists, must_create, value, true)?;
 
         Ok(())
     }
@@ -683,7 +660,7 @@ impl<'a> CircuitInputStateRef<'a> {
         reversible: bool,
     ) -> Result<(), Error> {
         // If receiver doesn't exist, create it
-        if (!receiver_exists && !value.is_zero()) || must_create {
+        if !receiver_exists && (!value.is_zero() || must_create) {
             self.account_write(
                 step,
                 receiver,
--- a/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
+++ b/zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
@@ -5,7 +5,7 @@ use crate::{
         step::ExecutionState,
         util::{
             and,
-            common_gadget::TransferWithGasFeeGadget,
+            common_gadget::TransferGadget,
             constraint_builder::{
                 ConstrainBuilderCommon, EVMConstraintBuilder, ReversionInfo, StepStateTransition,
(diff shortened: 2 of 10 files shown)


bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -635,56 +624,22 @@ impl<'a> CircuitInputStateRef<'a> {
...
         receiver: Address,
         receiver_exists: bool,
-        must_create: bool,
+        opcode_is_create: bool,
         value: Word,
         reversible: bool,
     ) -> Result<(), Error> {
-        // If receiver doesn't exist, create it
-        if (!receiver_exists && !value.is_zero()) || must_create {
+        if !receiver_exists && (!value.is_zero() || opcode_is_create) {
```
ChihChengLiang, line 642: This condition slips through the old PR that fixes all creation conditions.
https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1079/files/c0e6909d5135aa74e3ce415879d993a34050a823#r1336618956

zkevm-circuits/src/evm_circuit/execution/begin_tx.rs
```diff
@@ -469,16 +468,20 @@ impl<F: Field> ExecutionGadget<F> for BeginTxGadget<F> {
...
-        let mut caller_balance_sub_value_pair = (zero, zero);
-        let mut callee_balance_pair = (zero, zero);
-        if !tx.value.is_zero() {
-            caller_balance_sub_value_pair = rws.next().account_balance_pair();
-            callee_balance_pair = rws.next().account_balance_pair();
+        let callee_balance_pair = if !tx.value.is_zero() {
+            rws.next().account_balance_pair()
+        } else {
+            (zero, zero)
         };
```
ChihChengLiang: This is not a style fix. The order of read-write matters, and it reflects on the order of calling `rws.next()`.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
```diff
@@ -609,18 +609,7 @@ impl<'a> CircuitInputStateRef<'a> {
...
-                step,
-                AccountOp {
-                    address: receiver,
-                    field: AccountField::CodeHash,
-                    value: CodeDB::empty_code_hash().to_word(),
-                    value_prev: Word::zero(),
-                },
-            )?;
-        }
+
```
KimiWu123, line 612: It would cause rw inconsistent compared with the original implementation.
Let's say `receiver_exists` is false and `value` is zero, and
1. if `must_create` (or `opcode_is_create`) is false. It works well.
2. but if  `must_create` (or `opcode_is_create`) is true (a.k.a. this if condition holds), it pushes `AccountOp` to tw table. In your changes, you'll miss this account write operation bcs it returns at L624.

I guess that's the reason why light testing was failed.


I guess it doesn't matter but not 100% sure is that the order of rw table will be different in this change.
Before: (receiver, CodeHash) --> (sender, Balance) --> (receiver, Balance)
Now: (sender, Balance) --> (receiver, CodeHash) --> (receiver, Balance)

KimiWu123 (changes requested):
Not finished the review yet, but seems some changes needed and the light testing is falied.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: Ah yes, I fixed the begin_tx side but forgot the rest of the places. Let me do a quick fix.

ChihChengLiang (commented):
Fixing the read-write inconsistency. I'll this back to draft

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: We have to think carefully about this condition. I would love input from reviewers.
- What is receiver_exists?
  - We can check if code_hash is 0. It is an artificial rule we added before.
  - Nonce is 0. Should we add this? address_collision in create has this check.
- opcode_is_create
  - originally called must_create. I think this covers 3 cases:
  - EOA transfer to create an account. Can we call this situtation "opcode_is_create"?
  - Create receiver with CREATE opcode
  - Create receiver with CREATE2 opcode

ChihChengLiang (commented):
I collected all the reads and writes from assignment functions everywhere back to the transfer gadget. I hope this reduces the code complexity.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -380,77 +380,63 @@ impl<F: Field> TransferToGadget<F> {
...
-        receiver_address: WordLoHi<Expression<F>>,
-        receiver_exists: Expression<F>,
-        must_create: Expression<F>,
-        value_is_zero: Expression<F>,
-        reversion_info: Option<&mut ReversionInfo<F>>,
-    ) {
+        // Create account
         cb.condition(
             and::expr([
                 not::expr(receiver_exists.expr()),
```
KimiWu123, line 392: The condition here should align with [this discussion](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777#discussion_r1502323229)

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -380,77 +380,63 @@ impl<F: Field> TransferToGadget<F> {
...
         &self,
         region: &mut CachedRegion<'_, '_, F>,
         offset: usize,
-        (receiver_balance, prev_receiver_balance): (U256, U256),
-        value: U256,
+        rws: &mut StepRws,
+        (receiver_exists, value, opcode_is_create): (bool, U256, bool),
     ) -> Result<(), Error> {
+        if !receiver_exists && (!value.is_zero() || opcode_is_create) {
+            let _receiver_code_hash = rws.next().account_codehash_pair();
```
KimiWu123, line 426: I think we don't need `_receiver_code_hash` here.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -380,77 +380,63 @@ impl<F: Field> TransferToGadget<F> {
...
+        let receiver_balance_pair = if !value.is_zero() {
+            rws.next().account_balance_pair()
+        } else {
+            (0.into(), 0.into())
+        };
         self.receiver.assign(
             region,
             offset,
-            prev_receiver_balance,
+            receiver_balance_pair.1,
```
KimiWu123, line 437: Suggest to add some comment to explain what `receiver_balance_pair.0` and `receiver_balance_pair.1` are.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -309,7 +309,7 @@ impl<F: Field, const N_ADDENDS: usize, const INCREASE: bool>
     pub(crate) fn construct(
         cb: &mut EVMConstraintBuilder<F>,
         address: WordLoHi<Expression<F>>,
-        updates: Vec<Word32Cell<F>>,
+        updates: &[Word32Cell<F>],
```
KimiWu123, line 312: Just curious, what is the purpose of using `&[]` instead of `Vec`?

KimiWu123 (commented):
Not finished yet, but would like to discuss the callee existence earlier.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
KimiWu123: [From geth impl.](https://github.com/ethereum/go-ethereum/blob/577be37e0e7a69564224e0a15e49d648ed461ac5/core/vm/evm.go#L192-L208), an account which doesn't exist will be created only when transfer value is non-zero.

But, no matter this account exsists or not, it'll be created again in CREATE/2 opcode. You can see [createAccount is called in create](https://github.com/ethereum/go-ethereum/blob/577be37e0e7a69564224e0a15e49d648ed461ac5/core/vm/evm.go#L448-L449) and then will find this

// createObject creates a new state object. If there is an existing account with
// the given address, it is overwritten and returned as the second return value.
in [createObject](https://github.com/ethereum/go-ethereum/blob/577be37e0e7a69564224e0a15e49d648ed461ac5/core/state/statedb.go#L626-L628)

The conclusion is the original condition is correct. In your fix, the account won't be re-created if the receiver exists.

KimiWu123: For the naming of opcode_is_create. I would vote is_create_account or force_create_account as you mentioned "EOA transfer to create an account", it doesn't fix the name.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
ChihChengLiang: We need a codehash write there. Otherwise, the next `rws.next()` would fail.
We can comment instead of using a dummy variable. Let me know which do you prefer.
```suggestion
            // receiver's code_hash
            rws.next().account_codehash_pair();
```

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
ChihChengLiang: `Vec` takes a vector for argument.
`[]` takes a vector or slice for argument. If we already know the number of elements in an array, we can use slice `[]` instead of a vector `vec![]`. The intention here was to make lines shorter but not memory optimization.

ChihChengLiang (commented):
Replied some comments that I have quick answers

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
KimiWu123: Yeah, I know we need to call `rws.next()`. Using comment instead of a dummy variable looks good for me!

zkevm-circuits/src/evm_circuit/execution/create.rs
```diff
@@ -334,9 +334,10 @@ impl<F: Field, const IS_CREATE2: bool, const S: ExecutionState> ExecutionGadget<
                     create.caller_address(),
                     contract_addr.to_word(),
                     0.expr(),
```
KimiWu123, line 336: `false.expr()`

KimiWu123 (commented):
LGTM. It looks nice and clean now. Great work! The remaining thing is [this dicussion thread](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777#discussion_r1500541714)

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: > an account which doesn't exist will be created only when transfer value is non-zero.

I think this is correct and it matches our behavior.

> no matter this account exists or not, it'll be created again in CREATE/2 opcode. 

This is incorrect. Three lines above your quoted [createAccount](https://github.com/ethereum/go-ethereum/blob/577be37e0e7a69564224e0a15e49d648ed461ac5/core/vm/evm.go#L448-L449) locates a nonzero nonce and non-empty hash check that early exits the function and returns an `ErrContractAddressCollision` error.

The [EIP-1014: Skinny CREATE2](https://eips.ethereum.org/EIPS/eip-1014) also states "if nonce or code (of the create2 destination address) is nonzero, then the create-operation fails."

So for this PR, the conclusion is that a CREATEX opcode does not guarantee an overwrite of an account.

----

But you might wonder, why the create opcode only checks non empty code hash and non zero nonce. What about storage root and balance? Can we have an account that has code="" and nonce=0, but storage root !=0 or balance !=0? The tl;dr is we can't, for non-trivial reason. 

I fell into a rabbit hole of a historical study and how the EIP was defined. Fortunately, Gary wrote a best summary of this issue. See https://github.com/ethereum/go-ethereum/pull/28666.

ChihChengLiang (commented):
Hi @KimiWu123, I replied to the feedback to the receiver creation check.

KimiWu123 (approved):

bus-mapping/src/circuit_input_builder/input_state_ref.rs
KimiWu123: Sorry for the confusion. Maybe we could put the link https://github.com/ethereum/go-ethereum/pull/28666 in our comment.

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -370,8 +370,8 @@ impl<F: Field, const N_ADDENDS: usize, const INCREASE: bool>
 pub(crate) struct TransferToGadget<F> {
     receiver: UpdateBalanceGadget<F, 2, true>,
```
hero78119, line 371: suggest `balance: UpdateBalanceGadget<F, 2, true>` or `receiver_balance: UpdateBalanceGadget<F, 2, true>` for more self-explanation

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
```diff
@@ -380,75 +380,62 @@ impl<F: Field> TransferToGadget<F> {
...
     }
 
     pub(crate) fn assign(
         &self,
         region: &mut CachedRegion<'_, '_, F>,
         offset: usize,
-        (receiver_balance, prev_receiver_balance): (U256, U256),
-        value: U256,
+        rws: &mut StepRws,
+        (receiver_exists, value, opcode_is_create): (bool, U256, bool),
```
hero78119, line 424: A bit out-of-scope: do you think it's better we flatten `(receiver_exists, value, opcode_is_create): (bool, U256, bool)` to align with other convention? I believe it's just for avoid `#[allow(clippy::too_many_arguments)]`.

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: I added to the comment in the constraint part, which might be the place people will check more frequently [77cfbcb](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777/commits/77cfbcb9a3cdcf7c939996b339a2e7f48f9ffd30).

hero78119: > EOA transfer to create an account. Can we call this situtation "opcode_is_create"?

My opinion for `opcode_is_create` to represents `create an account` with tx.to = null is a bit confusing. How about just name it `is_create` to align with other place e.g. `call.is_create`?

hero78119 (commented):
Thanks for the code cleanup!
LGTM with few nitpicks

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
ChihChengLiang: Sounds good. Fixed in [990afe1](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777/commits/990afe19f694fa9d4fc037914758be31b7d224d1)

zkevm-circuits/src/evm_circuit/util/common_gadget.rs
ChihChengLiang: I was thinking those three arguments are related to receiver creation.
But I feel it's an overkill to group them in a tuple now. 

Fixed in [f8773e2](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777/commits/f8773e27f95259a6fe4e589e485a5fb0563746d9)

bus-mapping/src/circuit_input_builder/input_state_ref.rs
ChihChengLiang: Sorry I missed the renaming part before. I fixed the renaming here [7449289](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1777/commits/74492894b0198af74cee971eb425805b0b5b14f4)

hero78119 (approved):
👍 👍

PR #1784, opened 2024-02-27
Feat/#1752 fix witness generation - adding testing

### LINKED ISSUE #1752: [MPT - stack trie] fix witness generation

There was an implementation under `mpt-witness-generator/trie/stacktrie.go` already, but still has some bugs. This issue is mainly to fix it.
See [Miha's doc](https://hackmd.io/qqC_NERJRymwtKUEd-t-Ng?view) for more details. 

### Additional context

_No response_

KimiWu123:
### Description

closed #1752 

- Adding different numbers of txs to test different trie status (e.g. one leaf only, one ext. , one branch and one leaf ...etc)
- Fixing `GetProof` (if input is an ext node, it always assign first child of the root ext. node, `st.children[0]`)
- Fixing `getNodeFromBranchRLP` (if input is an leaf, it could throw an exception)
- Refactoring `getNodeFromBranchRLP`

### Issue Link

#1752 

### Type of change

### DIFF
--- a/geth-utils/gethutil/mpt/trie/stacktrie.go
+++ b/geth-utils/gethutil/mpt/trie/stacktrie.go
@@ -567,9 +567,9 @@ func (st *StackTrie) getNodeFromBranchRLP(branch []byte, idx int) []byte {
 	// `81` is the length of the payload and payload starts from `128`
 	start := int(branch[0]) - RLP_LONG_LIST_FLAG + 2
 
-	// If 1st node is not 128(empty node) or 160, it should be a leaf
+	// If 1st node is neither 128(empty node) nor 160, it should be a leaf
 	b := int(branch[start])
-	if b != RLP_SHORT_STR_FLAG || b != (RLP_SHORT_STR_FLAG+LEN_OF_HASH) {
+	if b != RLP_SHORT_STR_FLAG && b != (RLP_SHORT_STR_FLAG+LEN_OF_HASH) {
 		return []byte{0}
 	}
 
@@ -601,6 +601,14 @@ type StackProof struct {
 	proofC [][]byte
 }
 
+func (sp *StackProof) GetProofS() [][]byte {
+	return sp.proofS
+}
+
+func (sp *StackProof) GetProofC() [][]byte {
+	return sp.proofC
+}
+
 func (st *StackTrie) UpdateAndGetProof(db ethdb.KeyValueReader, indexBuf, value []byte) (StackProof, error) {
 	proofS, err := st.GetProof(db, indexBuf)
 	if err != nil {
@@ -627,7 +635,7 @@ func (st *StackTrie) UpdateAndGetProofs(db ethdb.KeyValueReader, list types.Deri
 	// order that `list` provides hashes in. This insertion sequence ensures that the
 	// order is correct.
 	var indexBuf []byte
-	for i := 1; i < list.Len() && i <= 0x7f; i++ {
+	for i := 0; i < list.Len() && i <= 0x7f; i++ {
 		indexBuf = rlp.AppendUint64(indexBuf[:0], uint64(i))
 		value := types.EncodeForDerive(list, i, valueBuf)
 
@@ -638,12 +646,12 @@ func (st *StackTrie) UpdateAndGetProofs(db ethdb.KeyValueReader, list types.Deri
 
 		proofs = append(proofs, proof)
 	}
-	if list.Len() > 0 {
-		indexBuf = rlp.AppendUint64(indexBuf[:0], 0)
-		value := types.EncodeForDerive(list, 0, valueBuf)
-		// TODO: get proof
-		st.Update(indexBuf, value)
-	}
+	// if list.Len() > 0 {
+	// 	indexBuf = rlp.AppendUint64(indexBuf[:0], 0)
+	// 	value := types.EncodeForDerive(list, 0, valueBuf)
+	// 	// TODO: get proof
+	// 	st.Update(indexBuf, value)
+	// }
 	for i := 0x80; i < list.Len(); i++ {
 		indexBuf = rlp.AppendUint64(indexBuf[:0], uint64(i))
 		value := types.EncodeForDerive(list, i, valueBuf)
--- a/geth-utils/gethutil/mpt/trie/stacktrie.go
+++ b/geth-utils/gethutil/mpt/trie/stacktrie.go
@@ -243,16 +243,17 @@ func (st *StackTrie) insert(key, value []byte) {
 				break
 			}
 		}
+
 		// Add new child
 		if st.children[idx] == nil {
-			st.children[idx] = stackTrieFromPool(st.db)
-			st.children[idx].keyOffset = st.keyOffset + 1
+			st.children[idx] = newLeaf(st.keyOffset+1, key, value, st.db)
+		} else {
+			st.children[idx].insert(key, value)
 		}
-		st.children[idx].insert(key, value)
+
 	case extNode: /* Ext */
 		// Compare both key chunks and see where they differ
 		diffidx := st.getDiffIndex(key)
-
 		// Check if chunks are identical. If so, recurse into
 		// the child node. Otherwise, the key has to be split
 		// into 1) an optional common prefix, 2) the fullnode
@@ -552,6 +553,7 @@ func (st *StackTrie) Commit() (common.Hash, error) {
 }
 
 const RLP_SHORT_STR_FLAG = 128
+const RLP_SHORT_LIST_FLAG = 192
 const RLP_LONG_LIST_FLAG = 248
 const LEN_OF_HASH = 32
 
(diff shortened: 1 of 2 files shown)


geth-utils/gethutil/mpt/trie/stacktrie.go
```diff
@@ -551,50 +551,49 @@ func (st *StackTrie) Commit() (common.Hash, error) {
...
-				insideInd = int(b) - 192
-			}
-			cInd += 1
-		} else {
-			if insideInd == 1 {
-				insideInd = -1
-			} else {
-				insideInd--
+			current_idx++
+		case RLP_SHORT_STR_FLAG + LEN_OF_HASH: // 160
```
miha-stopar, line 586: Note that `b = 160` could also be for any of the bytes of the branch child, it's not always the beginning of the new child, you need to somehow remember the length of the child and check whether you exhausted all the bytes of the child.

Regarding the child length - I think here you can rely it's either 32 or 1, because all the children are hashed. The child is not hashed if it's shorter than 32 bytes, but for transactions, receipts, and withdrawals I think this is never the case.

geth-utils/gethutil/mpt/witness/gen_witness_transactions_test.go
```diff
@@ -87,34 +97,59 @@ func TestNonHashedTransactionsStackTrieGetProof(t *testing.T) {
 	fmt.Println("===")
 }
 
+func TestStackTrieInsertion_1Tx(t *testing.T) {
```
miha-stopar, line 100: Perhaps would be good to add some check that the last proof element returned by `GetProof` is really the transaction for which proof we asked?

geth-utils/gethutil/mpt/trie/stacktrie.go
KimiWu123: Yep, I know `160` is not always the beginning of a new child, could be any of the bytes in a branch. My refactor is trying to parse RLP data instead of going through every single byte. And I found for transactions case, there are only `128` and `160` so I remove the original logic of handling `192`.  If I only focus on the case for transactions, wondering if this implementation make sense to you?

>Regarding the child length - I think here you can rely it's either 32 or 1, because all the children are hashed. The child is not hashed if it's shorter than 32 bytes, but for transactions, receipts, and withdrawals I think this is never the case.

I'm not sure I follow this. Do you mean for the case of transactions, it's always 1 (or 0, I'm assuming you're talking about an empty node) or 32 bytes? But above cases doesn't apply to receipts and withdraws, right? 

Could you elebarate more one this "The child is not hashed  if it's shorter than 32 bytes". Why is it shorter than 32 bytes if the child is not hashed? Won't it longer than 32 bytes?

geth-utils/gethutil/mpt/witness/gen_witness_transactions_test.go
KimiWu123: fixed in [056fa7f](https://github.com/privacy-scaling-explorations/zkevm-circuits/pull/1784/commits/056fa7f9be01929e7a28c1905d0aa90e0bf46741).

geth-utils/gethutil/mpt/trie/stacktrie.go
```diff
@@ -628,7 +635,7 @@ func (st *StackTrie) UpdateAndGetProofs(db ethdb.KeyValueReader, list types.Deri
 	// order that `list` provides hashes in. This insertion sequence ensures that the
 	// order is correct.
 	var indexBuf []byte
-	for i := 1; i < list.Len() && i <= 0x7f; i++ {
+	for i := 0; i < list.Len() && i <= 0x7f; i++ {
```
KimiWu123, line 638: Hi @miha-stopar , have a question regarding the starting index.
Why doest the for loop start from 1 instead of 0? I saw the case of index 0 was handled in L649-L654. But there is only `Update` being called, no `GetProof`. Wondering what will happen if I change to start from 0 and comment out the L649-L654? There is also a simliar case in L655, there is no `GetProof` after index is `0x80`. Wondering the reason why separate into many parts.

geth-utils/gethutil/mpt/trie/stacktrie.go
miha-stopar: But doesn't this code switch to the new child as soon as 128 or 160 is encountered (even if it's in the middle of the child and this byte actually means the child byte and not RLP encoding)?

The idea is to use this function for receipts and withdrawals as well.

When the leaf is shorter than 32 bytes, it won't get hashed when put in the branch RLP, it will be put in the branch as it is and its RLP encoding byte won't be 160 (its length won't be 32).

geth-utils/gethutil/mpt/trie/stacktrie.go
miha-stopar: Yeah, good question :). I tried to remember, but can't say for sure. I guess there was some bug in GetProof and I was debugging it. I think the index can start with 0 and we can delete these two if blocks.

geth-utils/gethutil/mpt/trie/stacktrie.go
KimiWu123: I didn't make it clear. Yes, this code switch to a new child as soon as 128 or 160. But my assumption is it won't jump into the middle of the child because it starts from index 2 (or 3, depending on starting byte is 248 or 249) and then it moves to next byte if it's a 128 or moves 32 bytes if it's a 160. If there are only 128 or 160 in the branch, then it should be fine. Is it too naive?

Agree, we should use this function for receipts ans withdrawals. However, I want to move a small step at a time and make sure everything works for transactions. That's why I only take care the case of transactions. If you think it's better to cover receipts and withdrawals, I could take time to dive into their format. That works for me as well!

I didn't realize the leaf will be shorter than 32 bytes here. I'll figure it out how to fix it.

miha-stopar: Yes, sorry, didn't realise you are increasing `i` by the length of the child. It should work fine. For non-hashed nodes I think not many changes are needed, you need to compute the length of the node as `b - 192`.

Agree on the incremental approach, we can take care of receipts and withdrawals later.

miha-stopar (approved):
LGTM!

geth-utils/gethutil/mpt/trie/stacktrie.go
KimiWu123: What if we could encode `k` in another way so we can get the value we expected? For example, for `i = 0`, we get `k = [0 0 16]`. If so, do we still need the above two blocks?

Another question is I added above two blocks back and in 129 txs case, `k` became `[8 1 8 0 16]` when `i = 129`. Does it fit our expecation? I thought the `k` is increasing like this, `[0 0 16], [0 1 16]... [1 0 16]... [1 15 16], [2 0 16]... [15 15 16], [1 0 0 16]...` dosen't it?

miha-stopar: For `i = 128`, we have `key = [8, 1, 8, 0]`, that means it's inserted in the branch at position 8 (which is a branch that has the transaction with `i = 0` at the position 0 - it was added in the block `if list.Len() > 0`). The second nibble is 1, so it's inserted at position 1 in the branch.

The state after `i = 128` is:
- top level is a branch with branches at positions 0 - 7 (all these branches are fully filled),
- at position 8 in a top level branch we have a branch with two children at positions 0 (`i = 0`) and 1 (`i = 128`).

KimiWu123: Thanks for detailed explanation again and I think I know how it works 🤞

KimiWu123:
Hey @miha-stopar , I think I fixed everything. Could you please give it a quick review again.

geth-utils/gethutil/mpt/trie/stacktrie.go
```diff
@@ -243,16 +243,17 @@ func (st *StackTrie) insert(key, value []byte) {
...
 			}
 		}
+
 		// Add new child
 		if st.children[idx] == nil {
-			st.children[idx] = stackTrieFromPool(st.db)
-			st.children[idx].keyOffset = st.keyOffset + 1
+			st.children[idx] = newLeaf(st.keyOffset+1, key, value, st.db)
+		} else {
+			st.children[idx].insert(key, value)
```
miha-stopar, line 251: Not sure what's the reason for this change?

KimiWu123: Oh, I just followed how geth implemented it (https://github.com/ethereum/go-ethereum/blob/00905f7dc406cfb67f64cd74113777044fb886d8/trie/stacktrie.go#L211-L215). Besides, it looks more self-explanation for me, one case is to new a leaf and the other is to insert a node. The original one seems too optimized and not easy to understand.

geth-utils/gethutil/mpt/trie/stacktrie.go
```diff
@@ -631,33 +662,40 @@ func (st *StackTrie) UpdateAndGetProofs(db ethdb.KeyValueReader, list types.Deri
 	for i := 1; i < list.Len() && i <= 0x7f; i++ {
```
ChihChengLiang, line 662: No change is requested here. I just comment to inform the future reviewers.

I saw the duplicated code chunks in L662's loop body, L674's condition body, and L684's loop body.
I was wondering why the duplication but the I saw your comment on L651, which says Geth is also implemented in the same style.

ChihChengLiang (approved):
LGTM. Great fix, refactors, and adding tests.

Issue #1798, opened 2024-04-03
Get bench results for average block VS keccak maxed out


ed255:
Get benchmark results (time & memory) to compare the cost of proving an average block VS a block that maxes out keccaks.

For example, given a target gas of 10k (or another sensible number) do:
- **a**. A block that consumes N gas with ERC20 transactions.  The circuit capacity is configured to only cover this case.
- **b**. A block that consumes N gas with keccaks (the highest number of keccacks you can do in 10k gas).  The circuit capacity is increased from case **a** to accommodate all the keccacks.

For case **b** there are two options to max out the keccaks (See https://github.com/privacy-scaling-explorations/zkevm-specs/issues/71#issuecomment-1125612886).
- Via EXTCODESIZE
- Via KECCAK256

It would be great to get results for both cases.

NOTE: To get meaningful results, we need to make sure the number of keccak rows for each case is close to a power of two (so that the circuit is not underused).  If we can't reach this, we could change the problem to do: given 2^k rows, do as many ops as possible to use all rows in each setting and figure out the gas spent.

The results from this will help reviewing https://github.com/ethereum/EIPs/pull/8367

ed255:
I plan to run two benchmarks.  For now I discarded the case where we abuse keccaks via EXTCODESIZE because the EIP-7667 is related to "gas cost of hash functions.

# Average block case

For this case we use a deployed ERC20 contract (based on openzeppelin implementation).  On top of ERC20 we have added a method that does many transfers in one call, so that we can do several ERC20 transfers in a single tx, to easily fill the block with operations.

We do 4 tx * 28 tranfsers (2 tx run successful ERC20 transfers, 2 txs run failed ERC20 transfers).
Gas cost is 303940

The circuit parameters required for this are:
```rust
FixedCParams {
        total_chunks: 1,
        max_rws: 70291,
        max_txs: 4,
        max_withdrawals: 0,
        max_calldata: 400,
        max_copy_rows: 25348,
        max_exp_steps: 0,
        max_bytecode: 4937,
        max_evm_rows: 200052, // max(200052, 68724)
        max_keccak_rows: 37200,
        max_vertical_circuit_rows: 0,
}
```

The biggest circuit ends up being the TxCircuit which needs 104471 rows per tx signature, so in total we need ~417884 rows.
K = 19 `log(417884, 2) = 18.67`

# Maxed keccak case

[...]

PR #1800, opened 2024-04-05
[benchmark only] Supercircuit

ed255:
SuperCircuit benchmark for k=19

This PR is only created to trigger a benchmark via github actions.

Do not merge.


ed255:
Benchmarks results in https://github.com/privacy-scaling-explorations/zkevm-circuits/issues/1798

PR #1816, opened 2024-05-09
Testool fix from Scroll

curryrasul:
### Description

This PR adds fixes from Scroll/testool - scroll-tech/zkevm-circuits#1235.

Initially I wanted to do that by `cherry-pick`, but it added much more here - so I did it manually.

### DIFF
--- a/testool/Config.toml
+++ b/testool/Config.toml
@@ -10,6 +10,13 @@ path="tests/src/GeneralStateTestsFiller/**/*"
 max_gas = 0
 max_steps = 100000
 
+[[suite]]
+id="EIP1153"
+path = "tests/src/GeneralStateTestsFiller/Cancun/stEIP1153-transientStorage/*"
+max_gas = 500000
+max_steps = 1000
+ignore_tests = []
+
 [[skip_paths]]
 desc = "unimplemented"
 paths = [
--- a/testool/Cargo.toml
+++ b/testool/Cargo.toml
@@ -34,6 +34,7 @@ rand_chacha = "0.3"
 rand = "0.8"
 halo2_proofs = { git = "https://github.com/privacy-scaling-explorations/halo2.git", tag = "v0.3.0" }
 urlencoding = "2.1.2"
+sha3 = "0.10"
 
 
 [features]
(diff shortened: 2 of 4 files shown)


testool/Config.toml
```diff
@@ -10,6 +10,13 @@ path="tests/src/GeneralStateTestsFiller/**/*"
 max_gas = 0
 max_steps = 100000
 
+[[suite]]
```
curryrasul, line 13: Not sure, do we need to add this @ChihChengLiang ?

curryrasul:
Another question - do we need to do changes in codehash.txt, as it's done [here](https://github.com/scroll-tech/zkevm-circuits/pull/1235/files#diff-af896e1754e274de6fb74de670a8c2bb1a2ffcae8ad01b716d09a9f69b5e0a15R9985-R10037)

testool/Config.toml
ChihChengLiang: We should add this one since our TLOAD TSTORE is implemented.

ChihChengLiang (commented):
> do we need to do changes in codehash.txt

We should add changes to codehash.txt. The content is the cached mapping of code hash to the corresponding bytecode. When we run the EIP1153 test, cached entries should be automatically generated. We check in those entries to reduce the testing overhead.

curryrasul:
> We should add changes to codehash.txt. The content is the cached mapping of code hash to the corresponding bytecode. When we run the EIP1153 test, cached entries should be automatically generated. We check in those entries to reduce the testing overhead.

@ChihChengLiang - We have an old version of ethereum/tests submodule, that doesn't contain Cancun tests. 
I updated it for the latest version and got the error when was running both default suite and EIP1153; and I also tried to switch to the commit that's scroll locked on and it also gives me error. 
The error is:
> Error found not a str

curryrasul:
@ChihChengLiang I updated the tests submodule

To reproduce the error:
1. `cd testool`
2. `cargo run --release --bin testool -- --suite EIP1153` 
or
`cargo run --release --bin testool` 

It returns:
*[2024-05-15T17:44:24Z INFO  testool] Using suite 'EIP1153'
[2024-05-15T17:44:24Z INFO  testool] Parsing and compiling tests...
Error found not a str*

